PR #4, opened 2019-04-02
protobuf parser plugin

samjsong:
Fluentd parser plugin, input is from Prometheus, and output is timeseries event of the format expected by the datapoint output plugin.

I manually tested in my Kubernetes cluster running Prometheus and Fluentd, with this parser plugin and the datapoint output plugin. The messages are of the expected form:

```json
{"@metric":"container_cpu_schedstat_runqueue_seconds_total","endpoint":"http-metrics","id":"/system.slice/systemd-random-seed.service","instance":"172.20.91.206:10255","job":"kubelet","node":"ip-172-20-91-206.us-west-1.compute.internal","prometheus":"monitoring/prometheus-operator-prometheus","prometheus_replica":"prometheus-prometheus-operator-prometheus-0","service":"prometheus-operator-kubelet","@timestamp":1554246017731,"@value":0.0}
```

FYI @lei-sumo @maimaisie @yuting-liu

### DIFF
--- a/fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
+++ b/fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
@@ -13,16 +13,14 @@ module Fluent
       def parse(text)
         begin
           inflated = Snappy.inflate(text)
-          log.info("HIT PARSER: #{Base64.encode64(inflated)}")
-  
           decoded = Prometheus::WriteRequest.decode(inflated)
           decoded.timeseries.map { |ts|
             log.debug(ts)
-            yield nil, ts
+            ts
           }
         rescue => e
           log.error("ERROR during decoding: #{e.message}")
-          yield nil, nil
+          nil
         end
       end
     end
--- a/fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
+++ b/fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
@@ -1,4 +1,4 @@
-require "fluent/plugin/parser"
+require 'fluent/plugin/parser'
 require 'base64'
 require 'google/protobuf'
 require 'snappy'
@@ -7,8 +7,9 @@ require_relative '../../remote_pb'
 
 module Fluent
   module Plugin
+    # fluentd parser plugin to parse Prometheus metrics into timeseries events.
     class ProtobufParse < Fluent::Plugin::Parser
-      Fluent::Plugin.register_parser("protobuf", self)
+      Fluent::Plugin.register_parser('protobuf', self)
 
       def parse(text)
         begin
@@ -18,9 +19,8 @@ module Fluent
             log.debug(ts)
             ts
           }
-        rescue => e
-          log.error("ERROR during decoding: #{e.message}")
-          nil
+        rescue StandardError => exception
+          log.error('ERROR during decoding', error: exception)
         end
       end
     end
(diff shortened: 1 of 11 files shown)


fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
```diff
@@ -13,16 +13,14 @@ class ProtobufParse < Fluent::Plugin::Parser
       def parse(text)
         begin
           inflated = Snappy.inflate(text)
-          log.info("HIT PARSER: #{Base64.encode64(inflated)}")
-  
           decoded = Prometheus::WriteRequest.decode(inflated)
```
lei-sumo: One question, do we use official Prometheus proto schemas compiled `WriteRequest` class to decode ?

samjsong: Unfortunately no. Since the `remote.proto` and `types.proto` files are written in `proto3` syntax, but `gogoproto` is written in `proto2`, we cannot compile the files into Ruby currently. According to the [protobuf docs](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated), `For the moment only proto3 is supported. Support for proto2 is planned, but not available yet.`

I tried a hacky thing by changing the `gogo.proto` file to `syntax = "proto3"` (and took out all of the `optional` keywords in the file since that is `proto2` syntax). Then I could compile `remote.proto` and `types.proto` without the compiler complaining that `gogo.proto` was `proto2`. However, the only difference in the generated Ruby file compared to when taking out the `gogoproto` dependency from the `remote.proto` and `types.proto` files was one line:
```
require 'gogoproto/gogo_pb'
```
so I believe the schema we are using here is valid and consistent with the official Prometheus proto schema.

But unfortunately I don't think there is anything we can do about this until the Ruby compiler supports `proto2`, or `gogoproto` is updated to `proto3`.

fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
```diff
@@ -13,16 +13,14 @@ class ProtobufParse < Fluent::Plugin::Parser
       def parse(text)
         begin
           inflated = Snappy.inflate(text)
```
bin3377: open question - all Prometheus write API streams are in same compressing and encoding settings?

fluent-plugin-protobuf/test/plugin/test_parse_protobuf.rb
```diff
@@ -6,13 +6,58 @@ class ProtobufParseTest < Test::Unit::TestCase
     Fluent::Test.setup
   end
 
-  test "failure" do
-    # flunk
+  sub_test_case "single batch single sample" do
```
bin3377: using single quote mark (`'`) if no string interpret

fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
```diff
@@ -13,16 +13,14 @@ class ProtobufParse < Fluent::Plugin::Parser
...
           decoded = Prometheus::WriteRequest.decode(inflated)
           decoded.timeseries.map { |ts|
             log.debug(ts)
-            yield nil, ts
+            ts
           }
         rescue => e
           log.error("ERROR during decoding: #{e.message}")
-          yield nil, nil
+          nil
```
bin3377: I think you can remove the `nil` return

fluent-plugin-protobuf/lib/fluent/plugin/parse_protobuf.rb
samjsong: I believe so. It seems that other Prometheus remote write integrations assume the same for compressing and encoding

PR #5, opened 2019-04-03
Byi carbonv2 plugin

bin3377:
plugin for transform data point to carbon v2 format
it supports:
- flatten nested json with `x.y.z`
- relabel the field name
- white/black list the records with regular expression rule applying on the value of fields

FYI @frankreno @rvmiller89 @maimaisie @yuting-liu


fluent-plugin-carbon-v2/README.md
```diff
@@ -22,22 +55,49 @@ gem "fluent-plugin-carbon-v2"
...
+For __any__ (`key`, `value`) pair in the hash, following records will be excluded in the output:
+
+- the value of field named with `key` matches the `value` (as regular expression).
+
+Default value: `{}`.
+
+### strict_exclusions (bool) (optional)
 
-You can copy and paste generated documents here.
+If `true`, records missing any field in keys of `exclusions` will be dropped.
```
samjsong: Is this the behaviour we want for `strict_exclusions`? In my mind, `strict_exclusions` would drop any records that include any field in keys of `exclusions`, rather than dropping the records that are missing those fields.

fluent-plugin-carbon-v2/lib/fluent/plugin/filter_carbon_v2.rb
```diff
@@ -1,13 +1,161 @@
...
+        wrap(to_carbon_line(relabeled)) if valid?(relabeled)
+      end
+
+      private
+
+      KEY_METRIC = '@metric'.freeze
+      KEY_TIMESTAMP = '@timestamp'.freeze
+      KEY_VALUE = '@value'.freeze
+      KEY_MESSAGE = 'message'.freeze
+      SPLITOR = '.'.freeze
```
samjsong: sp: SPLITTER

fluent-plugin-carbon-v2/README.md
bin3377: I may not very follow your statement here - drop any records that include any field in the keys of exclusions. 

Let's take a look case `exclusions = {"a" => "va", "b" => "vb"}` and `strict_exclusions = true`, so you think any record contains "a" or "b" will be dropped? so the value of `exclusions` mapping is useless here?

In my thought, on above case it should means record must have both "a" and "b" fields and `record["a"] != "va" && record["b"] != "vb"`

fluent-plugin-carbon-v2/lib/fluent/plugin/filter_carbon_v2.rb
bin3377: Not sure. In my mind these two words are different:

https://github.com/search?q=splitor&type=Code

https://github.com/search?q=splitter&type=Code

The earlier one may be more suitable in this case?

samjsong: Ahh, my bad. I did a quick google search but didn't find anything so I assumed you meant splitter :(

fluent-plugin-carbon-v2/README.md
samjsong: I see what you mean, thanks for clarifying. I agree that this is probably the more useful use case.

PR #7, opened 2019-04-08
space_as parameter

bin3377:
Since the space is not allowed in carbon V2 tags (name or value), add a parameter `space_as` to handle this case with replace all spaces in tags to a given string (by default use `_`)

### DIFF
--- a/fluent-plugin-carbon-v2/README.md
+++ b/fluent-plugin-carbon-v2/README.md
@@ -33,7 +33,7 @@
 
 ```json
 {
-    "message": "metric=http_request_size_bytes_sum _sourceType=kubernetes endpoint=http-metrics handler=prometheus instance=172.20.36.191:10251 job=kube-scheduler kubernetes.pod.name=kube-scheduler-ip-172-20-36-191.us-west-1.compute.internal kubernetes.service.name=kube-scheduler namespace=kube-system prometheus=monitoring/prometheus-operator-prometheus prometheus_replica=prometheus-prometheus-operator-prometheus-0 service=prometheus-operator-kube-scheduler   1619905.0 1550862304339"
+    "message": "metric=http_request_size_bytes_sum _origin=kubernetes endpoint=http-metrics handler=prometheus instance=172.20.36.191:10251 job=kube-scheduler kubernetes.pod.name=kube-scheduler-ip-172-20-36-191.us-west-1.compute.internal kubernetes.service.name=kube-scheduler namespace=kube-system prometheus=monitoring/prometheus-operator-prometheus prometheus_replica=prometheus-prometheus-operator-prometheus-0 service=prometheus-operator-kube-scheduler   1619905.0 1550862304339"
 }
 ```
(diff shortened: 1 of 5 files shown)


fluent-plugin-carbon-v2/README.md
```diff
@@ -100,4 +100,10 @@ If `true`, records missing any field in keys of `exclusions` will be dropped.
 
 Default value: `false`
 
-__NOTE__ inclusions/exclusions rules are applied after relabeling and flatten
\ No newline at end of file
+__NOTE__ inclusions/exclusions rules are applied after relabeling and flatten.
+
+### space_as (string) (optional)
+
+The space (`' '`) in the name or value of fields will be replaced with the given string in output.
```
lei-sumo: We should sync with @frankreno to see if this is acceptable. I feel name is fine, but changing value is bit weird.

bin3377: Okay... we can ask @frankreno for input here, but we don't have much choice, since AFAIK we don't support query any dimensions name/value includes space

bin3377: By looking into the `Carbon2Parser` code in receiver, looks like it just simply not allow any whitespace in key or value and will throw exception. There is no way to escape whitespace also.
The [metrics 2.0 spec](http://metrics20.org/implementations/) doesn't give any escaping approach as well. It just said "tags are space separated. Note double space to separate tags from meta_tags" so I guess not very helpful either

lei-sumo: I guess for carbon2, it makes sense that no space allowed as it's ambiguous (you can't differentiate if it's a delimiter or part of value). But does Prometheus have the same constraint? Wonder if it's a common case, we will get spaces in key or value. Also we should be consistent with the way our backend parses Prometheus format, but not carbon2 format?

bin3377: Prometheus doesn't have such constraint so the Prometheus format parser don't need to handle whitespace case either (it using `,` as separator between tags and `""` enclosing value). 
I guess this is an edge case but consider Prometheus/Kubernetes support customized tags it's a necessary to have a safe net anyway. 
On the other hand, is that possible to query some dimension with name/value include whitespace? If we cannot do it anyway on UI, keep whitespace in our ingested metadata will only cause confusing. right?

lei-sumo: Looks like we retain white space in value in `PrometheusParser` if the metrics is sent in Prometheus format. Could you double check that? That means, if we want to retain the format, we will have to convert time series to Prometheus format. I'm fine with merging now to unblock QE. But before deciding if we need send Prometheus format eventually, we'd like to understand:
1. Confirm if our backend retains spaces for prometheus metrics.
2. Understand where do those metrics with spaces key/value come from? If we see them often or just a few specific metrics having it?
3. Understand if our UI support searching kay/value with spaces.

cc @frankreno

PR #10, opened 2019-04-11
move _origin to meta tag

bin3377:
Move `_orgin=kubernetes` as a meta tag so we can workaround the 2 or 3 spaces problem for now.
We may move to prometheus format later for resolving the whitespace in tag
@samjsong @rvmiller89 @lei-sumo


fluent-plugin-carbon-v2/lib/fluent/plugin/filter_carbon_v2.rb
```diff
@@ -72,7 +72,7 @@ def to_carbon_line(record)
         metric = @metric_accessor.call(record).gsub(/\s/, @space_as)
         timestamp = @timestamp_accessor.call(record)
         value = @value_accessor.call(record)
-        "metric=#{metric} #{to_tags(record)}   #{value} #{timestamp}"
+        "metric=#{metric} #{to_tags(record)}  #{ORIGIN_KEY}=#{ORIGIN_VALUE} #{value} #{timestamp}"
```
rvmiller89: LGTM, although I think we determined we could also just change this to
```
metric=#{metric} #{to_tags(record)}  #{value} #{timestamp}
```
(2 spaces)

bin3377: yes. it should also works. 
We may change it to Prometheus format in near future, so just keep it for now to unblock E2E

PR #11, opened 2019-04-12
Byi helm yaml

bin3377:
- move all stuff about fluentd into single .yaml file for easier setup
- change `service` to internal
- add files/steps for setting up Prometheus
@lei-sumo @samjsong


deploy/README.md
```diff
@@ -0,0 +1,85 @@
+# Deployment Guide (Draft)
+
+## Rre-requests
```
rvmiller89: Prerequisites ?

deploy/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+* Make sure a Kubernetes cluster is created and you can use `kubectl` to access it.
+* Make sure Kubernetes cluster enable the DNS service ([steps](https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#dns))
```
rvmiller89: s/enable/enables

deploy/README.md
```diff
@@ -0,0 +1,85 @@
...
+* Create HTTP source(s) in your Sumo Logic account and get HTTP Source URL(s).
+* Run `echo -n '<YOUR_HTTP_URL>' | base64` to encode the URL with base-64.
```
rvmiller89: what's the reason we encode the URL with base64? Our other open source projects aren't asking the user to do this.

samjsong: Secrets in Kubernetes require base64 encoded data in the yaml file. I believe alternatively we could do
```
stringData:
  endpoint-metrics: <URL>
```
rather than
```
data:
  endpoint-metrics: <encoded URL>
```
but then the customer is storing their endpoint in plain text, not sure if that's a concern.

rvmiller89: base64 is plaintext :D

samjsong: oh 😂nice

bin3377: hmm... I don't know it can be a plain text. That will be easier

bin3377: looks like the plain text doesn't work:
my yaml:
```
apiVersion: v1
kind: Secret
metadata:
  name: sumologic
  namespace: sumologic
  labels:
    k8s-app: fluentd-sumologic
type: Opaque
data:
  stringData: https://stag-events.sumologic.net/receiver/v1/http/XXXX==
```

```
Error from server (BadRequest): error when creating "./fluentd-sumologic.yaml": Secret in version "v1" cannot be handled as a Secret: v1.Secret.ObjectMeta: v1.ObjectMeta.TypeMeta: Kind: Data: decode base64: illegal base64 data at input byte 5, error found in #10 byte of ...|ONAeDNA=="},"kind":"|..., bigger context ...|8-KUxvOfgLKqJMmoiq5-2divPm1_55jwiFAaqoq2ONAeDNA=="},"kind":"Secret","metadata":{"annotations":{"kube|...

```

bin3377: NVM, looks like I place it to wrong place...

deploy/README.md
```diff
@@ -0,0 +1,84 @@
+# Deployment Guide (Draft)
+
+## Prerequests
```
rvmiller89: I think you mean "Prerequisites".  "Prerequests" isn't a word

bin3377: 😂I need a spell checking plugin

bin3377:
@ggarg2906sumo you can try with the new .yaml files

ggarg2906sumo:
Thanks @bin3377 , I will try this :)

PR #12, opened 2019-04-16
select metrics

bin3377:
filter the metrics from write api config following https://docs.google.com/document/d/1z0KFYygDMahOXsXrAYPOwmSkTmSDTOhUdVwOwxXusv0

@frankreno @samjsong @lei-sumo 

### DIFF
--- a/deploy/helm/overrides.yaml
+++ b/deploy/helm/overrides.yaml
@@ -1,4 +1,58 @@
 prometheus:
   prometheusSpec:
     remoteWrite:
-    - url: http://fluentd:9888/prometheus.metrics
+    - url: http://fluentd:9888/prometheus.metrics.state.daemonset
+      writeRelabelConfigs:
+      - action: keep
+        regex: kube-state-metrics;(kube_daemonset_status_current_number_scheduled|kube_daemonset_status_desired_number_scheduled|kube_daemonset_status_number_misscheduled|kube_daemonset_status_number_unavailable|kube_daemonset_metadata_generation)
+        sourceLabels: [job, __name__]
+    - url: http://fluentd:9888/prometheus.metrics.state.deployment
+      writeRelabelConfigs:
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -72,13 +72,16 @@ data:
...
+    <match prometheus.metrics.**>
       @type datapoint
-      tag "prometheus.datapoint"
+      tag prometheus.datapoint
     </match>
     <filter prometheus.datapoint>
       @type carbon_v2
-    </filter>    
+    </filter>
+    # <filter prometheus.datapoint>
```
samjsong: I think we should remove the comment

bin3377: It's for easier debugging. I'm ok to remove it

deploy/helm/overrides.yaml
```diff
@@ -1,4 +1,58 @@
 prometheus:
   prometheusSpec:
     remoteWrite:
-    - url: http://fluentd:9888/prometheus.metrics
+    - url: http://fluentd:9888/prometheus.metrics.state.daemonset
```
lei-sumo: Are these new url matching to the source endpoints created by Sam's PR?

samjsong: I don't think these urls need to match with the source endpoints created by my PR. Since fluentd accepts all of these metrics from prometheus and more or less treats them all the same (since they are all retagged with `prometheus.datapoint`). Then we separate the metrics again by `job` to send to the source endpoints on Sumo's end.

bin3377: yes. they are transformed to same tag in datapoint filter stage so no impact. spliting them into different urls maybe benefit for the statistics later.

PR #13, opened 2019-04-16
Split metrics to multiple http sources

samjsong:
We want to split metrics to multiple HTTP sources to prevent blacklisting. Synced with Frank about the more common jobs we'd want to separate out, which included:
```
apiserver
cadvisor
kubelet
kube-controller-manager
kube-scheduler
kube-state*
node-exporter
prometheus-operator-*
```
We still include a catchall http source for those metrics not belonging to any of the above jobs.

@frankreno @bin3377 @lei-sumo 

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml
+++ b/deploy/kubernetes/fluentd-sumologic.yaml
@@ -84,17 +84,17 @@ data:
       @type datapoint
       tag "prometheus.datapoint"
     </match>
-    <filter prometheus.datapoint>
-      @type carbon_v2
-    </filter>
     <match prometheus.datapoint>
       @type rewrite_tag_filter
       <rule>
-        key message
-        pattern job=(.*?)\s
+        key job
+        pattern ^(.*)$
         tag ${tag}.$1
       </rule>
     </match>
+    <filter prometheus.datapoint.**>
+      @type carbon_v2
+    </filter>
 
     <match prometheus.datapoint.apiserver**>
       @type sumologic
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -78,8 +86,65 @@ data:
     </match>
     <filter prometheus.datapoint>
       @type carbon_v2
-    </filter>    
+    </filter>
     <match prometheus.datapoint>
+      @type rewrite_tag_filter
```
bin3377: Is that possible to move this filter before transform to `carbon_v2`?
rule may be something like `key=job`, `pattern=^(.*)$`.
The reason is we may move to Prometheus format later so we don't want it impacted

samjsong: That makes sense. Let me try this and get back to you.

samjsong: It seems to have worked! resolving

deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -76,10 +84,67 @@ data:
       @type datapoint
       tag "prometheus.datapoint"
     </match>
-    <filter prometheus.datapoint>
-      @type carbon_v2
-    </filter>    
     <match prometheus.datapoint>
+      @type rewrite_tag_filter
```
lei-sumo: What does this part do?

samjsong: The `rewrite_tag_filter` output plugin allows you to rewrite the tag for rerouting. In this case the plugin looks for the job and appends it to the already existing tag, which becomes `prometheus.datapoint.<JOB>`

samjsong:
cc @ggarg2906sumo this PR has been merged, so now you will need multiple (nine) http sources on the sumo side

PR #14, opened 2019-04-23
Byi fix timeseries

bin3377:
Fix the problem: only one point goes into pipeline in one HTTP request.
The problems are:
1. parse plug-ins need to implement `parse(text)` with `yield` return with time and record (e.g. hash)
refer to:
https://docs.fluentd.org/v0.12/articles/plugin-development#parser-plugins
https://github.com/fluent/fluentd/tree/master/lib/fluent/plugin
2. since single HTTP request may contains a list of TimeSeries not one, add a wrapper to make it a hash and also change datapoint plugin to handle it

update unit test make sure the time stamp and records are correctly bring in.

verified with private docker image on stag. the timeseries now giving points in fixed interval

### DIFF
--- a/fluent-plugin-protobuf/lib/fluent/plugin/parser_protobuf.rb
+++ b/fluent-plugin-protobuf/lib/fluent/plugin/parser_protobuf.rb
@@ -1,5 +1,4 @@
 require 'fluent/plugin/parser'
-require 'base64'
 require 'google/protobuf'
 require 'snappy'
 require_relative '../../types_pb'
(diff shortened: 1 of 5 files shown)


fluent-plugin-protobuf/lib/fluent/plugin/parser_protobuf.rb
```diff
@@ -11,17 +10,15 @@ module Plugin
...
-          }
-        rescue StandardError => exception
-          log.error('ERROR during decoding', error: exception)
-        end
+        inflated = Snappy.inflate(text)
+        decoded = Prometheus::WriteRequest.decode(inflated)
+        log.trace "protobuf::parse - in: (#{text.bytesize}/#{inflated.bytesize}), out: #{decoded.timeseries.length}"
+        record = {}
+        record[KEY_TIMESERIES] = decoded.timeseries
+        yield Fluent::EventTime.now, record
```
lei-sumo: Do we override the time eventually using the value in record itself?

lei-sumo:
cc @ggarg2906sumo This supposed to fix the missing metrics issue.

fluent-plugin-protobuf/lib/fluent/plugin/parser_protobuf.rb
bin3377: yes. this time is only used in fluentd pipeline

bin3377:
@ggarg2906sumo new image pushed to dockerhub now

PR #15, opened 2019-04-24
Prometheus format filter plugin

samjsong:
This is the filter plugin for converting data points into the Prometheus format. Mostly the same as the Carbon V2 plugin, the major difference is we get rid of the `space_as` config option since space is allowed in the Prometheus format.


fluent-plugin-prometheus-format/lib/fluent/plugin/filter_prometheus_format.rb
```diff
@@ -0,0 +1,157 @@
...
+            record[new_key] = value unless new_key.nil? || new_key.empty?
+          end
+        end
+        record
+      end
+
+      def to_tags(hash)
+        array = @sort_labels ? hash.sort : hash.to_a
+        array.map do |key, value|
+          "#{key}=\"#{value}\""\
```
bin3377: Need to escape `\` `"` and `\n` in value
Please also add unit test for these escaping

https://github.com/Sanyaku/sumologic/blob/49f0f94b92a1bcada303f55ab71e1151779246d0/collector-interchange/src/main/scala/com/sumologic/scala/interchange/collector/PrometheusParser.scala#L147

PR #16, opened 2019-04-25
Byi override yaml update

bin3377:
update yaml files for:
- fix the changed name/job in metrics
- add cluster label
- remove cadvisor source since it's combined into kubelet

### DIFF
--- a/deploy/helm/overrides.yaml
+++ b/deploy/helm/overrides.yaml
@@ -31,6 +31,7 @@ prometheus:
       - action: keep
         regex: scheduler_e2e_scheduling_latency_microseconds.*|scheduler_binding_latency_microseconds.*|scheduler_scheduling_algorithm_latency_microseconds.*
         sourceLabels: [__name__]
+    # api server metrics
     - url: http://fluentd:9888/prometheus.metrics.apiserver
       writeRelabelConfigs:
       - action: keep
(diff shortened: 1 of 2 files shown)


deploy/helm/overrides.yaml
```diff
@@ -1,56 +1,137 @@
 prometheus:
   prometheusSpec:
+    externalLabels:
+      # Set this to a value to distinguish between different k8s clusters
+      cluser: kubernetes
```
frankreno: missing the t in cluster

bin3377:
@ggarg2906sumo this change should fix all missing time series. These things need to be noticed:

1. There is an incompatible between Prometheus-Operator chart and latest version of Kubernetes. So we need a workaround for install helm chart
 a) clone https://github.com/helm/charts/
 b) modify `values.yaml`: find `component: kube-controller-manager` and replace to `k8s-app: kube-controller-manager`; find `component: kube-scheduler` and replace to `k8s-app: kube-scheduler`
 c) install chart with modified values.yaml with running following command in <path-to-charts-repo>/stable:
`helm install --namespace sumologic --name prometheus-operator ./prometheus-operator -f prometheus-operator/values.yaml -f <path-to-sumologic-kubernetes-collection>/deploy/helm/overrides.yaml --no-crd-hook`

2. time series in cadvisor is combined into kubelet 

3. some metrics name changed in [node-exporter part](https://docs.google.com/document/d/1z0KFYygDMahOXsXrAYPOwmSkTmSDTOhUdVwOwxXusv0/edit#heading=h.yz65u4z4c7x1)

frankreno:
why clone? why not set in helm overrides as it can override those value?

frankreno:
I also think that due to the version change and use of many k8s versions in the wild, we should have an FAQ doc or something and just call this out specifically.

bin3377:
I tried with make these changes into override.yaml, but it will not overwrite the original selector but add them as:
```
Name:              prometheus-operator-kube-controller-manager
...
Selector:          component=kube-controller-manager,k8s-app=kube-controller-manager
```

frankreno:
Crap. I forgot about that bug in helm. Ok is what it is for the moment.

bin3377:
@lei-sumo @samjsong

PR #19, opened 2019-05-02
rename from pod_name/container_name to pod/container

bin3377:
relabel pod to pod_name, container to container_name
https://jira.kumoroku.com/jira/browse/SUMO-109075
FYI - @frankreno @lei-sumo


frankreno:
@bin3377 we should go the other direction, rename pod_name -> pod and container_name->container as this is the direction the sig insturmentation is moving to to stanardize - https://github.com/kubernetes/kubernetes/pull/69099

PR #23, opened 2019-05-06
Byi more replicas

bin3377:
Tuning fluentd resource to fix [Context canceled issue](https://jira.kumoroku.com/jira/browse/SUMO-109456)

- Use `3` replica rather than `1`
- Reduce Memory from `2Gi` to `256Mi` (each container)
- Reduce CPU from `200m` to `100m` (each container)

Monitoring for couple of hours and looks like no more `context canceled` error in Prometheus log now

FYI - [definition of Kubernetes resource limitation](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/)
@frankreno @ggarg2906sumo

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml
+++ b/deploy/kubernetes/fluentd-sumologic.yaml
@@ -155,7 +155,7 @@ spec:
   selector:
     matchLabels:
       k8s-app: fluentd-sumologic
-  replicas: 1
+  replicas: 3
   template:
     metadata:
       labels:
--- a/deploy/kubernetes/fluentd-sumologic.yaml
+++ b/deploy/kubernetes/fluentd-sumologic.yaml
@@ -176,11 +176,11 @@ spec:
         imagePullPolicy: Always
         resources:
           limits:
-            memory: 2Gi
-            cpu: "200m"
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -155,7 +155,7 @@ spec:
   selector:
     matchLabels:
       k8s-app: fluentd-sumologic
-  replicas: 1
+  replicas: 3
```
samjsong: Is there any data duplication with 3 replicas of Fluentd running? Or when Prometheus writes to the remote write endpoint do the metrics only get sent to one of the 3 replicas?

bin3377: No data duplication. the idea is much similar as our receiver nodes. from outside (e.g. Prometheus) there is only single endpoint (which we defined as a kubernetes service). The traffic will be distributed to 3 nodes. We don't have any session or partition concept here so basically it's just randomly pick anyone

samjsong: Nice, thanks for confirming

PR #33, opened 2019-05-28
implement enhance-k8s-metadata plugin with label reading

bin3377:
- reading labels of pod
- caching
- unit test

@lei-sumo @samjsong 
FYI - @maimaisie @yuting-liu @rvmiller89


fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -6,7 +6,91 @@ module Plugin
...
+        super
+        normalize_param
+        connect_kubernetes
+        init_cache
+        @in_namespace_ac = record_accessor_create(@in_namespace_path)
+        @in_pod_ac = record_accessor_create(@in_pod_path)
+      end
+
       def filter(tag, time, record)
+        decorate_record(record)
```
lei-sumo: time not used here? Or it is inside record?

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -6,7 +6,91 @@ module Plugin
...
+      private
+
+      def decorate_record(record)
+        namespace_name = @in_namespace_ac.call(record)
+        pod_name = @in_pod_ac.call(record)
+        if namespace_name.nil?
+          log.debug "Record doesn't have [#{@in_namespace_path}] field"
+        elsif pod_name.nil?
+          log.debug "Record doesn't have [#{@in_pod_path}] field"
+        else
```
lei-sumo: So we will only get label for pod, but not for other objects?

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -6,7 +6,91 @@ module Plugin
...
+
+      helpers :record_accessor
+      include SumoLogic::Kubernetes::Connector
+      include SumoLogic::Kubernetes::Reader
+      include SumoLogic::Kubernetes::CacheStrategy
+
+      # parameters for read/write record
+      config_param :in_namespace_path, :string, default: '$.namespace'
+      config_param :in_pod_path, :string, default: '$.pod'
+      config_param :out_root, :string, default: 'kubernetes'
```
lei-sumo: Is this part even configurable?

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -6,7 +6,91 @@ module Plugin
...
+      def normalize_param
+        # Use Kubernetes default service account if running in a pod.
+        if @kubernetes_url.nil?
+          log.debug 'Kubernetes URL is not set - inspecting environment'
+          env_host = ENV['KUBERNETES_SERVICE_HOST']
+          env_port = ENV['KUBERNETES_SERVICE_PORT']
+          @kubernetes_url = "https://#{env_host}:#{env_port}/api" unless env_host.nil? || env_port.nil?
+        end
+        log.info "Kubernetes URL: '#{@kubernetes_url}'"
+
```
lei-sumo: Could you explain a bit how authentication works? Where does customer get certificates or bearer token?

fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/cache_strategy.rb
```diff
@@ -0,0 +1,35 @@
...
+      require 'lru_redux'
+      require_relative 'reader.rb'
+
+      CACHE_TYPE_POD_LABELS = 'pod_labels'.freeze
+      CACHE_TYPE_OWNER_REFS = 'owner_refs'.freeze
+
+      def init_cache
+        @all_caches = {
+          CACHE_TYPE_POD_LABELS => LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl),
+          CACHE_TYPE_OWNER_REFS => LruRedux::TTL::ThreadSafeCache.new(@cache_size, @cache_ttl)
```
lei-sumo: What does owner refs do? Don't see any usage here?

fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/cache_strategy.rb
```diff
@@ -0,0 +1,30 @@
...
+        }
+      end
+
+      def get_pod_labels(namespace_name, pod_name)
+        key = "#{namespace_name}::#{pod_name}"
+        cache = @all_caches[CACHE_TYPE_POD_LABELS]
+        labels = cache[key]
+        if labels.nil?
+          labels = fetch_pod_labels(namespace_name, pod_name)
```
lei-sumo: A general question: it's a bit hard to follow this kind of function call in another module. Is it by convention people ignore the class reference? Would be much easier to follow if we know where doesn't that function come from. Also what if required modules overload a method, which one would be used?

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
bin3377: it's inside the record. time here is only for showing relative message in fluentd log

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
bin3377: yes

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
bin3377: we probably need it's configurable for reuse it on metrics/log. I don't think it's changeable if we want application work

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
bin3377: These files/tokens/env are injected by k8s when containers are launched so the processes inside of containers can talk to k8s-apiserver, k8s-dns, services, etc.
In a typical cluster, customer don't need to setup any of them, they are automatically generated.

To look into it, a simplest way is running into the containers:
 `kubectl exec -it <pod-name> /bin/sh`

fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/cache_strategy.rb
bin3377: Planning to use it later for deployment lookup. I can remove it for now

fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/cache_strategy.rb
bin3377: You can consider the module is another approach to do something like writing util functions in `trait` in Scala but more flexible. (e.g. [duck typing](https://stackoverflow.com/a/4205163/5472728)) It's more natural way to reuse code in Ruby rather than using inheritance.

For connivence, if you install Ruby plugin in VS code (and some dependency gems following the instruction), it can support Cmd+Click to jump between functions from different modules

lei-sumo: Gtk. Thanks!

fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/connector.rb
```diff
@@ -0,0 +1,60 @@
+module SumoLogic
+  module Kubernetes
+    # module for connecting to Kubernetes cluster
+    module Connector
```
maimaisie: is it possible to re-use it in the events plugin?

bin3377: It's possible, but need to split the code into a mixin so different plugins can refer to. 
Or you can just copy the code into your plugin first with same module names and later we can merge and change the dependencies

PR #34, opened 2019-05-29
Collect logs with FluentBit and Fluentd

samjsong:
The goal of this PR is to create a FluentBit daemonset that collects the same logs as the existing Fluentd setup [here](https://github.com/SumoLogic/fluentd-kubernetes-sumologic).

The FluentBit daemonset will send the logs (without any preprocessing to keep it lightweight) to the Fluentd deployment, which acts as an aggregator, attaches the relevant metadata to the logs, then forwards to Sumo.

We make no code changes in this PR - we utilize the FluentBit helm chart to make only config changes. The filter plugin from `fluentd-kubernetes-sumologic` is copied over as is

Steps for installing (will update the README's in a future PR):
```
helm install --namespace sumologic --name fluent-bit -f ./deploy/fluent-bit/overrides.yaml stable/fluent-bit
kubectl apply -f ./deploy/kubernetes/fluentd-sumologic.yaml
```

FYI - @maimaisie @yuting-liu @rvmiller89


bin3377:
LGTM. We can add `SUMO_ENDPOINT_LOGS` source creating in the deployment script. but it can be a separated PR. Thanks!

PR #35, opened 2019-05-31
Events plugin

maimaisie:
This PR adds the event plugin to collect k8s events from the API server. To start the events deployment, run 
```
kubectl -n sumologic create secret generic sumologic --from-literal=endpoint-events=<events-http-source-url>
kubectl apply -f deploy/kubernetes/fluentd-sumologic-events.yaml
```

We will need to move everything in this yaml file to the existing file `fluentd-sumologic.yaml` once everything has finalized, in a separate PR.

This diagram shows the flow of the plugin:
Design Document - Kubernetes Events Ingestion

Currently I am duplicating connector.rb (connects to kubeclient) that @bin3377 wrote in the enhance-k8s-metadata plugin. We will refactor it later.

FYI @lei-sumo @samjsong @rvmiller89


fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -0,0 +1,144 @@
...
+
+      def start_configmap_flush_thread
+        thread_create(:"update_configmap") do
+          resource = ::Kubeclient::Resource.new
+          resource.metadata = {
+            name: "fluentd-config-resource-version",
+            namespace: "sumologic"
+          }
+          while true do
+            sleep(10)
```
maimaisie: maybe we can make the flush interval configurable

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -0,0 +1,144 @@
...
+          thread_create(:"watch_events") do
+            watcher.each do |entity|
+              log.debug { "Received new object from watching events" }
+              entity = JSON.parse(entity)
+              router.emit tag, Fluent::Engine.now, entity
+              @resource_version = entity['object']['metadata']['resourceVersion']
+  
+              if (!@resource_version)
+                @resource_version = 0
+                sleep(5)
```
yuting-liu: Maybe this one can also be configurable?

maimaisie: yeah I thought about this, but since this is just a safeguard to avoid the plugin going into a hot loop when in case every event it gets has empty resource version, I thought it probably doesn't really provide much value to make it configurable for the user

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -0,0 +1,144 @@
...
+                start_watcher_thread
```
bin3377: If my understanding correctly, you want to restart a watcher if something get wrong, but that should also handle the exceptions like invalid JSON or parsing failure?
And on the other side, the watcher restart in another thread, may cause thread leaking? (like different threads start multiple watcher?

maimaisie: > that should also handle the exceptions like invalid JSON or parsing failure?

This is a good point. I will add some exception handling for invalid JSON and parsing failure after we receive each event. What should we do in case of that? Should we just log the exception/error and skip the event we can't parse and move on?

> the watcher restart in another thread, may cause thread leaking? (like different threads start multiple watcher?

After the current thread calls `start_watcher_thread` here, it breaks out of the watch loop and finishes executing its block of code. I think at this point this thread is considered completed and not running.

PR #37, opened 2019-06-03
Byi oweners ref

bin3377:
Append names of resources in owner chain to metadata:
- Fetch owner information in BFS
- Cache owners' informatin
- Using multiple kubeclient objects to get resources in different API groups/versions
- Change `in_namespace_path` and `in_pod_path` to array since we need to adaptor both `pod_name` and `pod`

FYI - @maimaisie @yuting-liu


fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/reader.rb
```diff
@@ -4,31 +4,117 @@ module Kubernetes
...
+        while !queue.empty? && depth <= MAX_BFS_DEPTH
+          size = queue.size
+          [1..size].each do
+            current = queue.shift
+            result << current unless current == resource
+
+            owner_references(current).each do |owner_reference|
+              begin
+                if visited.key?(owner_reference['uid'])
+                  log.debug "#{owner_reference['name']} visted."
```
samjsong: nit: `visited`

PR #38, opened 2019-06-04
Update setup script for logs and events

samjsong:
1. We update the setup script to include a source for logs and events, and attach `-metrics` suffix to metrics sources
2. We rename the secret to `sumologic` and include all endpoints in this single secret
3. We include an optional `cluster` field to the collector upon creation to keep it consistent with metrics

FYI - @ggarg2906sumo @yuting-liu @lei-sumo

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -65,17 +65,18 @@ This approach requires access to the Sumo Logic Collector API. It will create a 
 
 ```sh
 curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/kubernetes/setup.sh \
-  | bash -s <api-endpoint> <access-id> <access-key> [collector-name]
+  | bash -s [-c collector-name] [-k cluster-name] <api-endpoint> <access-id> <access-key>
 ```
 
 __NOTE__ This script will be executed in bash and requires [jq command-line JSON parser](https://stedolan.github.io/jq/download/) to be installed.
 
 #### Parameters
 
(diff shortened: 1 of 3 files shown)


samjsong:
Thanks for the catch, let me make the relevant README changes in this PR

deploy/README.md
```diff
@@ -65,41 +65,44 @@ This approach requires access to the Sumo Logic Collector API. It will create a
...
-  | bash -s <api-endpoint> <access-id> <access-key> [collector-name]
+  | bash -s [-c collector-name] [-k cluster-name] <api-endpoint> <access-id> <access-key>
 ```
 
 __NOTE__ This script will be executed in bash and requires [jq command-line JSON parser](https://stedolan.github.io/jq/download/) to be installed.
 
 #### Parameters
 
+* __-c collector-name__ - optional. Name of Sumo collector that will be created. If not specified, it will be named as `kubernetes-<timestamp>`
+* __-k cluster-name__ - optional. Name of the Kubernetes cluster that will be attached to logs as metadata. If not specified, it will be named as `kubernetes-<timestamp>`
```
bin3377: so this change only apply to log? (how about metrics/event?)

samjsong: `cluster name` attaches the `fields:{cluster:CLUSTER_NAME}` to the metadata of the collector, which gets attached to all the logs sent to that collector. Actually since we receive events as logs in Sumo, this should also apply for events. For metrics, you have to set the same `CLUSTER_NAME` in the prometheus operator `overrides.yaml`
```
prometheus:
  prometheusSpec:
    externalLabels:
      # Set this to a value to distinguish between different k8s clusters
      cluster: kubernetes
```

Eventually this value can be enforced the same in the helm chart when we release that.

PR #40, opened 2019-06-06
Update README with events collection

maimaisie:
cc @ggarg2906sumo


deploy/README.md
```diff
@@ -254,6 +263,34 @@ This filter will:
 * Rename the label/metadata `container_name` to `container`, and `pod_name` to `pod`.
 * Only apply to metrics with the `kube-system` namespace
 
+### Events
+
+#### Filter events
+
+By default, the events plugin collects all events from all namespaces. You can filter the events collected by specifying  `namespace`, `label_selector` or `field_selector`.
```
maimaisie:
```suggestion

By default, the events plugin collects all events from all namespaces. You can filter the events collected by specifying  `namespace`, `label_selector` or `field_selector`. You can also filter on the event type using `type_selector`. By default, we only collect `ADDED` and `MODIFIED` events. (`DELETED` events are ignored.) 
```

PR #42, opened 2019-06-09
Forwarding all Remote Write Metrics

abhi-sumo:
* Sending all remote write related Prometheus metrics to Sumo
* This will allow us to monitor and manage Prometheus better on customer accounts by showing us the delays, replays, and instability in using Prometheus as a collector.

### DIFF
--- a/deploy/helm/overrides.yaml
+++ b/deploy/helm/overrides.yaml
@@ -238,7 +238,7 @@ prometheus:
       - action: keep
         regex: up
         sourceLabels: [__name__]
-    # prometheus metrics
+    # prometheus remote write metrics
     - url: http://fluentd:9888/prometheus.metrics
       writeRelabelConfigs:
       - action: keep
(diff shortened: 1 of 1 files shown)


deploy/helm/overrides.yaml
```diff
@@ -238,3 +238,9 @@ prometheus:
       - action: keep
         regex: up
         sourceLabels: [__name__]
+    # prometheus metrics
```
lei-sumo: should be prometheus remote write metrics?

deploy/helm/overrides.yaml
```diff
@@ -238,3 +238,9 @@ prometheus:
...
+    - url: http://fluentd:9888/prometheus.metrics
+      writeRelabelConfigs:
+      - action: keep
+        regex: 'prometheus_remote_storage_*'
```
lei-sumo: @abhi-sumo Could you link us the prometheus doc on this?

samjsong: synced with Abhi offline, he'll loop back with us with the docs. In the meantime he can show us the metrics in prometheus. going to merge the PR though so the metrics are available

PR #43, opened 2019-06-10
Increase the resource limits on CPU and memory for fluentd deployment

samjsong:
Based on the metrics I collected on CPU and Memory usage, I feel these are reasonable numbers for request and limit. This is sufficient for collecting up to 5MB/sec = 432GB/day rates. For greater loads, we can either increase the number of Fluentd replicas, or increase the CPU limits. Memory usage seems to be relatively stable even with higher loads.

FYI @ggarg2906sumo @lei-sumo @maimaisie @yuting-liu @frankreno

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml
+++ b/deploy/kubernetes/fluentd-sumologic.yaml
@@ -291,11 +291,11 @@ spec:
         imagePullPolicy: Always
         resources:
           limits:
-            memory: 256Mi
-            cpu: "100m"
+            memory: 1000Mi
+            cpu: 1
           requests:
-            memory: 256Mi
-            cpu: "100m"
+            memory: 700Mi
+            cpu: 0.5
         ports:
         - name: prom-write
           containerPort: 9888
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -291,11 +291,11 @@ spec:
         imagePullPolicy: Always
         resources:
           limits:
-            memory: 256Mi
-            cpu: "100m"
+            memory: 1000Mi
```
rvmiller89: nit: make this `1Gi`.  `1000Mi` is like the worst of both worlds... using base-2 nomenclature but giving it a base-10 value.

deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -291,11 +291,11 @@ spec:
...
+            cpu: 1
           requests:
-            memory: 256Mi
-            cpu: "100m"
+            memory: 700Mi
```
rvmiller89: nit: stick to a base-2 value, `768Mi`

deploy/kubernetes/fluentd-sumologic.yaml
samjsong: haha, sounds good. thanks for the feedback Ryan

PR #45, opened 2019-06-13
CPU % extra metrics

abhi-sumo:
Adding container_spec_cpu_quota and container_spec_cpu_period so that we can actually calculate CPU %.

https://github.com/google/cadvisor/issues/2026#issuecomment-486134079

### DIFF
--- a/deploy/helm/overrides.yaml
+++ b/deploy/helm/overrides.yaml
@@ -123,6 +123,9 @@ prometheus:
       - action: keep
         regex: kubelet;container_spec_memory_(?:|swap|reservation)_limit_bytes
         sourceLabels: [job, __name__]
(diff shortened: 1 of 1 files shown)


deploy/helm/overrides.yaml
```diff
@@ -123,6 +123,9 @@ prometheus:
       - action: keep
         regex: kubelet;container_spec_memory_(?:|swap|reservation)_limit_bytes
         sourceLabels: [job, __name__]
+      - action: keep
+        regex: kubelet;container_spec_cpu_(?:|quota|period)
+        sourceLabels:[job, __name__]
```
samjsong: nit: space between : and [

PR #48, opened 2019-06-19
Fix events plugin

maimaisie:
After discovering the issue with k8s watch and list API regarding the resource version parameter, we redesigned the plugin flow. Now, the main thread acts as a monitor that loops forever (until the plugin is told to shut down) to update configmap and preemptively recreate the watch thread periodically. In more detail, it mainly does these things:
1. Pull latest resource version from the `List` API and store in configmap every `configmap_update_interval_seconds` (10s by default).
2. Check if the watch thread is running. If it's not, create the watcher thread using the resource version from the the configmap to resume from an earlier point in time if 
    1. this is the first iteration (when the plugin just started and the watch thread does not exist)
    2. the watcher thread somehow died (thread is not alive)
3. Check if it has been `watch_interval_seconds` since the last time we recreated the watch thread. If so, first pull latest resource version, close the watch thread by calling kubeclient method to close the http connection, then create a new watch stream using the RV we just pulled

We listen for SIGTERM so on graceful shutdown we can break out of the while loop in `start_monitor` (apparently if the while loop keeps running, the shutdown sequence won't get triggered)

FYI @rvmiller89 @lei-sumo @frankreno

### DIFF
--- a/fluent-plugin-events/lib/fluent/plugin/in_events.rb
+++ b/fluent-plugin-events/lib/fluent/plugin/in_events.rb
@@ -131,19 +131,19 @@ module Fluent
         end
       end
 
-      def create_config_maps
+      def create_config_map
         @resource_version = 0
         @configmap.data = { "resource-version": "#{@resource_version}" }
-        @client.public_send("create_config_map", @configmap).tap do |maps|
-          log.debug "Created config maps: #{maps}"
+        @client.public_send("create_config_map", @configmap).tap do |map|
+          log.debug "Created config map: #{map}"
         end
       end
 
       def update_config_map
         pull_resource_version
         @configmap.data = { "resource-version": "#{@resource_version}"}
-        @client.public_send("update_config_map", @configmap).tap do |maps|
-          log.debug "Updated config maps: #{maps}"
+        @client.public_send("update_config_map", @configmap).tap do |map|
+          log.debug "Updated config map: #{map}"
         end
       end
 
@@ -157,12 +157,12 @@ module Fluent
         # get or create the config map
         begin
           @client.public_send("get_config_map", "fluentd-config-resource-version", "sumologic").tap do |resource|
-            log.debug "Got config maps: #{resource}"
+            log.debug "Got config map: #{resource}"
             version = resource.data['resource-version']
             @resource_version = version.to_i if version
           end
         rescue Kubeclient::ResourceNotFoundError
-          create_config_maps
+          create_config_map
         end
       end
 
--- a/fluent-plugin-events/lib/fluent/plugin/in_events.rb
+++ b/fluent-plugin-events/lib/fluent/plugin/in_events.rb
@@ -33,6 +33,7 @@ module Fluent
       config_param :label_selector, :string, default: nil
       config_param :field_selector, :string, default: nil
       config_param :configmap_update_interval_seconds, :integer, default: 10
+      config_param :watch_interval_seconds, :integer, default: 300
       
       def configure(conf)
         super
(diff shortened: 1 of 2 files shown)


fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -43,90 +44,139 @@ def configure(conf)
...
         initialize_resource_version
-        start_watcher_thread
-        start_configmap_flush_thread
+        start_monitor
       end
   
-      def close
-        @watcher.each &:finish
+      def stop
+        log.debug "Clean up before stopping completely"
```
rvmiller89: can we be sure to document in the README how to enable debug logs for a fluentd plugin (is there a way to only enable debug logs for a specific plugin?) so that the customer/support team can enable these debug logs as needed?

maimaisie: Yes, in the README Sam put in troubleshooting steps that includes turning on debug logs for metrics. We'll make that generic to all three data types afterwards

rvmiller89: 👍  Thanks

samjsong: Thanks, I've just updated the plugin with steps to enable debug logs for a specific plugin

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -43,90 +44,139 @@ def configure(conf)
...
+        end
+
+        while !interrupted do
+          # Periodically restart watcher connection by checking if enough time has passed since 
+          # last time watcher thread was recreated or if the watcher thread has been stopped.
+          now = Time.now.to_i
+          watcher_exists = Thread.list.select {|thread| thread.object_id == @watcher_id && thread.alive?}.count > 0
+          if now - last_recreated >= @watch_interval_seconds || !watcher_exists
+            
+            log.debug "Recreating watcher thread. Use resource version from latest snapshot if watcher is running"
```
yuting-liu: Do we need to put the value of `@resource_version` in the log line, maybe after `pull_resource_version`?

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -43,90 +44,139 @@ def configure(conf)
...
   
       def start_watcher_thread
+        log.debug "Starting watcher thread"
         params = Hash.new
         params[:as] = :raw
         params[:resource_version] = @resource_version
         params[:field_selector] = @field_selector
         params[:label_selector] = @label_selector
         params[:namespace] = @namespace
+        params[:timeout_seconds] = @watch_interval_seconds + 60
```
yuting-liu: Are we going to tune this value later, so maybe need a variable for this 60 seconds?

maimaisie:
@samjsong Good question. We found that if we don't specify the timeout parameter on the watch stream, it will be closed by the API server anyway after around 50 minutes even if events are continuously being generated. So now we specify a timeout on the watch stream and preemptively close it before the timeout so we can pull the latest RV before we close it. If we only rely on the timeout and let the API server close it, we wound't have a way to do anything before it gets closed. (we can keep a timer ourselves but we don't know if it will be in sync with the k8s api server)

fluent-plugin-events/lib/fluent/plugin/in_events.rb
maimaisie: We log the resource version being used to start the watch stream in `start_watcher_thread`
```
log.debug "New thread to watch events #{@watcher_id} from resource version #{params[:resource_version]}"
```

fluent-plugin-events/lib/fluent/plugin/in_events.rb
maimaisie: Currently not seeing the need to make this user-configurable, so probably no need to make it a variable either

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -43,90 +44,139 @@ def configure(conf)
...
-          }
-          while true do
-            sleep(@configmap_update_interval_seconds)
-            resource.data = { "resource-version": "#{@resource_version}"}
-
-            # update the config map
-            @client.public_send("update_config_map", resource).tap do |maps|
-              log.debug "Updated config maps: #{maps}"
-            end
+      def create_config_maps
```
samjsong: oops, sorry forgot to add the nit comment: since we have only one config map, can we change all the `maps` to `map`?

fluent-plugin-events/lib/fluent/plugin/in_events.rb
yuting-liu: That works.

PR #51, opened 2019-06-19
Add configurable event type filter

maimaisie:
This PR adds a fluentd config param for specifying a list of event types to ingest. An error will be thrown if (1) the list is empty; (2) the list size exceeds three; (3) each element in the list is not one of ADDED, MODIFIED or DELETED (uses **case-insensitive** string compare, Ruby's `casecmp` is case insensitive and returns 0 if two strings are the same ignoring case, ie, `"ADDED".casecmp("addED") == 0`)

User can then have something like this:
```
    <source>
      @type events
      type_selector ["ADDED", "deleted"]
    </source>
```

### DIFF
--- a/fluent-plugin-events/lib/fluent/plugin/in_events.rb
+++ b/fluent-plugin-events/lib/fluent/plugin/in_events.rb
@@ -32,11 +32,17 @@ module Fluent
       config_param :namespace, :string, default: nil
       config_param :label_selector, :string, default: nil
       config_param :field_selector, :string, default: nil
+      config_param :type_selector, :array, default: ["ADDED", "MODIFIED"], value_type: :string
       config_param :configmap_update_interval_seconds, :integer, default: 10
       config_param :watch_interval_seconds, :integer, default: 300
       
       def configure(conf)
         super
+
+        @valid_types = ["ADDED", "MODIFIED", "DELETED"]
+        raise Fluent::ConfigError, 'type_selector needs to be an array with maximum 3 elements: ADDED, MODIFIED and DELETED.' \
+          if @type_selector.length > 3 || !@type_selector.any? || !@type_selector.all? {|type| @valid_types.any? {|valid| valid.casecmp(type) == 0}}
+
         normalize_param
         connect_kubernetes
       end
@@ -114,7 +120,7 @@ module Fluent
             watcher.each do |entity|
               begin
                 entity = JSON.parse(entity)
-                router.emit tag, Fluent::Engine.now, entity
+                router.emit tag, Fluent::Engine.now, entity if @type_selector.any? {|type| type.casecmp(entity['type']) == 0}
                 rv = entity['object']['metadata']['resourceVersion']
               rescue => e
                 log.error "Got exception #{e} parsing entity #{entity}. Skipping."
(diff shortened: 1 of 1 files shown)


fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -32,11 +32,17 @@ class EventsInput < Fluent::Plugin::Input
...
       config_param :label_selector, :string, default: nil
       config_param :field_selector, :string, default: nil
+      config_param :type_selector, :array, default: ["ADDED", "MODIFIED"], value_type: :string
       config_param :configmap_update_interval_seconds, :integer, default: 10
       config_param :watch_interval_seconds, :integer, default: 300
       
       def configure(conf)
         super
+
+        @valid_types = ["ADDED", "MODIFIED", "DELETED"]
```
samjsong: Is this a complete list of the event types? I couldn't find any docs on this, would appreciate if you could point me in the right direction

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -32,11 +32,17 @@ class EventsInput < Fluent::Plugin::Input
...
+        raise Fluent::ConfigError, 'type_selector needs to be an array with maximum 3 elements: ADDED, MODIFIED and DELETED.' \
+          if @type_selector.length > 3 || !@type_selector.any? || !@type_selector.all? {|type| @valid_types.any? {|valid| valid.casecmp(type) == 0}}
```
samjsong: I think since we only care about exact equality (0 case) then we can use `casecmp?(other_str)` which returns true for equality and false for other cases

fluent-plugin-events/lib/fluent/plugin/in_events.rb
maimaisie: that reminds me there is also an `ERROR` type. I think it's errors related to using the watch API. in that case there won't be a resource version returned, and when that happens we skip the event and start a new watch stream. I will try to find docs on this, but I haven't seen any TBH

fluent-plugin-events/lib/fluent/plugin/in_events.rb
maimaisie: oh, did not see that one

fluent-plugin-events/lib/fluent/plugin/in_events.rb
maimaisie: [This](https://github.com/kubernetes/kubernetes/commit/2a76b82394a8853298f64bfe83df1015a05259e6#diff-54aa255ae082627d581223e1b6257a1aR44) code snippet shows the types. `BOOKMARK` is very interesting that I just discovered. It's an event that only contains a resource version for a resource. It acts like a heartbeat for watch that tells you even though you didn't get a watch event for a resource, there hasn't been any changes to that resource up to this resource version. more details [here](https://github.com/kubernetes/kubernetes/issues/73585). But it is still in alpha so I don't think we want to add it here yet.

samjsong: Thanks for looking into it Maisie, that sounds good to me.

samjsong:
good catch 😅

PR #52, opened 2019-06-24
Nik automation friendly

fernandesnikhil:
I found I can't customize the namespace all the Sumo collection components deploy into - so I changed the scripts a bit to allow that.

I also find it easier to work with environment variables than script args so I added support for that.


deploy/kubernetes/setup.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e
+set -x
```
fernandesnikhil: whoops will remove this

deploy/kubernetes/setup.sh
```diff
@@ -58,18 +58,38 @@ do
...
+    CLUSTER_NAME="kubernetes-$TIME";
+  else
+    CLUSTER_NAME=$KUBERNETES_CLUSTER_NAME
+  fi
+fi
+
+if [ -z $NAMESPACE]; then
+  if [ -z $SUMO_NAMESPACE ]
+  then
+    NAMESPACE="sumo"
```
samjsong: Can we make the default `sumologic` to stay consistent with the original default value?

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -437,7 +432,6 @@ apiVersion: v1
 kind: Service
 metadata:
   name: fluentd
-  namespace: sumologic
```
samjsong: Will the script fill these values in automatically? seems like we only added the `namespace: $NAMESPACE` for the service account

deploy/kubernetes/setup.sh
fernandesnikhil: Good call - Didn't intend to change the default.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
fernandesnikhil: If you look at line 189 of `setup.sh` we use `kubectl -n` so for all objects that don't have a namespace set in the YAML it will use the namespace passed in via `-n`.
The ClusterRoleBinding is different in that we need to define in which namespace the `subject` lives. So for that we keep the `subjects[].ServiceAccount.namespace` field.

samjsong: Gotcha, thanks for clarifying.

samjsong: I guess the issue is that people who don't use the setup script will have to manually replace the `$NAMESPACE` or remember to use sed before applying. Since setup is generally a one-time process I'm okay with this, and when we integrate helm this should be a non-issue.

fernandesnikhil: 👍

samjsong: We should update the deployment README to specify this change though, cc @frankreno

samjsong:
cc @maimaisie @yuting-liu @ggarg2906sumo

fernandesnikhil:
And a final test to check it works is also a good idea. Hoping @abhi-sumo has cycles to help with that.

maimaisie:
I noticed the file name change from `fluentd-sumologic.yaml` to `fluentd-sumologic.yaml.tmpl`, is this intentional?

fernandesnikhil:
@maimaisie Yes the file rename is intentional and serves to indicate you can't just apply that file as-is with `kubectl` for those users to prefer to set things up manually.

deploy/kubernetes/setup.sh
```diff
@@ -58,18 +58,38 @@ do
  in
  c) COLLECTOR_NAME=${OPTARG};;
  k) CLUSTER_NAME=${OPTARG};;
+ n) NAMESPACE=${OPTARG};;
```
abhi-sumo: You need to update line 55 to also have the getopts recognize the  `-n`

samjsong:
+1 on Maisie's comment, I was worried about having a separate .yaml file initially since we don't want to have the files diverge accidentally, but if the script generates the file then there should be no divergence.

fernandesnikhil:
I don't quite follow the diverging files statement. We have `setup.sh` that uses `fluentd-sumologic.yaml.tmpl`. Which is the same as what we do before except that before the YAML file as a complete unit. And with these changes it needs to have the namespace passed in. If I renamed it to `fluentd-sumologic.tmpl` would that be better? Then it becomes clear this isn't a standalone YAML file.

fernandesnikhil:
Ah OK read @maimaisie full comment at will think about that.

fernandesnikhil:
@maimaisie am I current in understanding you want `setup.sh` to render the YAML to local disk so that customers can re-apply it ad-lib when needed?

maimaisie:
Yep exactly

fernandesnikhil:
@maimaisie @samjsong  could you'll take another look? Made the requested changes.

deploy/README.md
```diff
@@ -74,10 +74,19 @@ __NOTE__ This script will be executed in bash and requires [jq command-line JSON
 
 * __-c collector-name__ - optional. Name of Sumo collector that will be created. If not specified, it will be named as `kubernetes-<timestamp>`
 * __-k cluster-name__ - optional. Name of the Kubernetes cluster that will be attached to logs and events as metadata. If not specified, it will be named as `kubernetes-<timestamp>`. For metrics, specify the cluster name in the `overrides.yaml` provided for the prometheus operator; further details in [step 2](#step-2-configure-prometheus).
+* __-n namespace__ - optional. Name of the Kubernetes namespace in which to deploy resources. If not specified, the namesapce will default to `sumologic`
```
samjsong: nit: `namesapce` (line 86 as well)

deploy/kubernetes/setup.sh
```diff
@@ -162,6 +182,9 @@ kubectl -n sumologic create secret generic sumologic \
   --from-literal=endpoint-events=$ENDPOINT_EVENTS
 
 echo "Applying deployment 'fluentd'..."
-kubectl apply -f https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/kubernetes/fluentd-sumologic.yaml
+curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/kubernetes/fluentd-sumologic.yaml.tmpl | \
+sed 's/\$NAMESPACE'"/$NAMESPACE/g" | \
+tee sumologic-collection.yaml | \
+kubectl -n $NAMESPACE apply -f sumologic-collection.yaml
```
samjsong: I like the name `sumologic-collection`, should we consider renaming `fluentd-sumologic.yaml.tmpl` to `sumologic-collection.yaml.tmpl` as well? cc @frankreno @abhi-sumo

fernandesnikhil: Or the other way around? I should leave the generated file as `fluentd-sumologic.yaml`?

samjsong: Actually, on second thought any customer that deploys manually will need to deploy this yaml for fluentd, then deploy fluentbit and prometheus separately. So I think it makes more sense to leave the name as `fluentd-sumologic`, sorry @fernandesnikhil do you mind changing it back?

fernandesnikhil: Done.

PR #53, opened 2019-06-26
add fluent metrics

frankreno:
Not Complete, don't merge just yet.  This PR does the following:

1) Enables Fluent-Bit metrics in Fluent-Bit helm overrides
2) Adds the writeLabelConfig to collect fluentD and Fluent-Bit metrics
3) Adds the configuration needed to expose fluentD metrics in Event pod

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml
+++ b/deploy/kubernetes/fluentd-sumologic.yaml
@@ -378,11 +378,51 @@ data:
     <source>
       @type events
     </source>
+    <source>
+      @type prometheus
+      bind 0.0.0.0
+      port 24322
+      metrics_path /metrics
+    </source>
+    <source>
+      @type prometheus_output_monitor
+      interval 10
+      <labels>
+        hostname ${hostname}
+      </labels>
+    </source>
+    <filter kubernetes.**>
+      @type prometheus
+      <metric>
+        name fluentd_input_status_num_records_total
+        type counter
+        desc The total number of incoming records
+        <labels>
+          tag ${tag}
+          hostname ${hostname}
+        </labels>
+      </metric>
+    </filter>
     <match kubernetes.**>
-      @type sumologic
-      endpoint "#{ENV['SUMO_ENDPOINT_EVENTS']}"
-      data_type logs
-      disable_cookies true
+      @type copy
+      <store>
+        @type sumologic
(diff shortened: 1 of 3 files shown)


Gourav2906:
@frankreno when are we planning to merge this?

frankreno:
I need more time to figure out why it does not work with our FluentD deployment as its an incomplete picture.  I likely won't have time to look at until next week.

deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -378,11 +378,51 @@ data:
     <source>
       @type events
     </source>
+    <source>
+      @type prometheus
```
rvmiller89: do we need to add gem install like https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/83/files#diff-aa72c1ddeec59a31f30834d956165473R29

deploy/kubernetes/fluentd-sumologic.yaml
```diff
@@ -378,11 +378,51 @@ data:
     <source>
       @type events
     </source>
+    <source>
```
rvmiller89: this will only add metrics to fluentd-events, but we probably want to instrument fluentd deployment for logs/metrics?  Not sure how valuable it is to get fluentbit metrics.

deploy/kubernetes/fluentd-sumologic.yaml
rvmiller89: Ah, we already are including it ... we may want to bump the version to 1.4, but never mind this comment.

Gourav2906:
for fluent-bit we can check 
https://github.com/helm/charts/pull/14416/files

deploy/kubernetes/fluentd-sumologic.yaml
frankreno: Yes.  Totaly going to, I need to play with it more.  I was hitting an issue getting it to work with our plugin.  The Prometheus fluentD plugin that is used to instrument was not happy with the fluentD deployment.  Fluent-Bit metrics could be used to identify issues with dropped data as it will tell us what it is reading by tag, which is the full file path.

frankreno:
@rvmiller89 is working on a new PR for this, closing and I will submit fluent bit.

PR #56, opened 2019-06-27
Fix missing metrics

samjsong:
During testing we found the following metrics were missing:
- `cloudprovider_.*_api_request_duration_seconds.*`
  - this was due to the job changing from `kube-controller-manager` to `kubelet`
- `container_memory_swap`
  - this was due to a bug in the regex
- `container_spec_.*`
  - this was due to trying to add multiple regex matchers in the same writeRelabelConfigs section, which I believe is piped and therefore the metrics were being dropped
- `node_memory_Free_bytes`
  - this metric was renamed to `node_memory_MemFree_bytes` and `node_memory_SwapFree_bytes`

cc @ggarg2906sumo

### DIFF
--- a/deploy/helm/overrides.yaml
+++ b/deploy/helm/overrides.yaml
@@ -121,7 +121,10 @@ prometheus:
         regex: POD
         sourceLabels: [container_name]
       - action: keep
-        regex: kubelet;container_spec_memory_(?:|swap|reservation)_limit_bytes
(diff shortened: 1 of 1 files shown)


Gourav2906:
@samjsong Can you please update metrics list document also with updated names

samjsong:
@ggarg2906sumo done, thanks for the reminder!

PR #62, opened 2019-07-03
Parameterize resource name to support watching other resources in events plugin

maimaisie:
Introduce new config param `resource_name` to take in a resource name to watch, defaulted to `events`. Use the same config map to store per-resource RV using key in the `resource-version-<resource-name>` format.

Minor change: rename `apiVersion` to `api_version` to follow the convention of other config param naming.

FYI @rvmiller89 @lei-sumo

### DIFF
--- a/fluent-plugin-events/lib/fluent/plugin/in_events.rb
+++ b/fluent-plugin-events/lib/fluent/plugin/in_events.rb
@@ -14,7 +14,7 @@ module Fluent
       
       # parameters for connecting to k8s api server
       config_param :kubernetes_url, :string, default: nil
-      config_param :apiVersion, :string, default: 'v1'
+      config_param :api_version, :string, default: 'v1'
       config_param :client_cert, :string, default: nil
       config_param :client_key, :string, default: nil
       config_param :ca_file, :string, default: nil
@@ -28,6 +28,7 @@ module Fluent
       config_param :ssl_partial_chain, :bool, default: false
       
       # parameters for events collection
+      config_param :resource_name, :string, default: "events"
       config_param :tag, :string, default: 'kubernetes.*'
       config_param :namespace, :string, default: nil
       config_param :label_selector, :string, default: nil
@@ -55,7 +56,7 @@ module Fluent
   
       def stop
         log.debug "Clean up before stopping completely"
-        update_config_map
+        patch_config_map
         @watch_stream.finish if @watch_stream
         super
       end
@@ -96,7 +97,7 @@ module Fluent
             log.debug "last_recreated updated to #{last_recreated}"
           end
 
-          update_config_map
+          patch_config_map
           sleep(@configmap_update_interval_seconds)
         end
       end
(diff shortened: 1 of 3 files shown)


fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -138,22 +139,21 @@ def start_watcher_thread
...
         end
       end
 
-      def update_config_map
+      def patch_config_map
         pull_resource_version
-        @configmap.data = { "resource-version": "#{@resource_version}"}
-        @client.public_send("update_config_map", @configmap).tap do |map|
-          log.debug "Updated config map: #{map}"
+        @client.public_send("patch_config_map", "fluentd-config-resource-version", {data: { "resource-version-#{resource_name}": "#{@resource_version}"}}, "sumologic").tap do |map|
```
samjsong: I just realized we're creating this config map in the `sumologic` namespace, can we make this a configurable parameter? Since some customers might want to deploy their collection setup in other namespaces

maimaisie: good call, will change that

maimaisie:
@rvmiller89  Yes we probably should. @yuting-liu If you are working on tests for watching events, let's add one for another resource type for example pods

yuting-liu:
Sure. I'd go ahead adding one test case for this.

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -138,33 +139,32 @@ def start_watcher_thread
...
         end
       end
 
       def initialize_resource_version
+        @resource_version = 0
         @configmap = ::Kubeclient::Resource.new
         @configmap.metadata = {
           name: "fluentd-config-resource-version",
-          namespace: "sumologic"
+          namespace: @namespace
```
samjsong: Correct me if I'm wrong, but I thought `@namespace` was for which namespace we want to watch resources from? I think we might need a separate parameter for which namespace we want to deploy in

maimaisie: ah I think you are right.. I saw the namespace param and immediately thought it's the one we can use 😂

fluent-plugin-events/lib/fluent/plugin/in_events.rb
```diff
@@ -138,33 +140,32 @@ def start_watcher_thread
...
         @configmap.metadata = {
           name: "fluentd-config-resource-version",
-          namespace: "sumologic"
+          namespace: @deploy_namespace
         }
 
         # get or create the config map
         begin
-          @client.public_send("get_config_map", "fluentd-config-resource-version", "sumologic").tap do |resource|
+          @client.public_send("get_config_map", "fluentd-config-resource-version", @deploy_namespace).tap do |resource|
```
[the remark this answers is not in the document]
maimaisie: the config map is created in the `deploy_namespace` namespace, and because config map can have multiple KVPs, each resource will have a KVP for its own resource version.

```
$ kubectl describe configmap fluentd-config-resource-version
Name:         fluentd-config-resource-version
Namespace:    sumologic
Labels:       <none>
Annotations:  <none>

Data
====
resource-version-config_maps:
----
2024903
resource-version-pods:
----
2025332
resource-version-services:
----
2024977
Events:  <none>
```
You can see here in the `Data` field there are 3 KVPs

maimaisie:
I didn't update `Starting events collection` since we are still collecting events even if the resource name is pods or services, it's just pod events and service events in that case. There is another log line that prints the resource name: `New thread to watch #{resource_name} blah..`

maimaisie:
@ggarg2906sumo FYI, after this PR is merged you'll be able to watch events for a specific resource type in K8s. We want to make sure to cover that in testing

PR #63, opened 2019-07-09
SUMO-111064 Add unit test for events plugin

yuting-liu:
Add unit tests for events plugin:
* pull_initial_resource_version
* initialize_resource_version
* watch events: by default only with type `ADDED` and `MODIFIED`. Customers can also decide `type_selector` by configuration to decide whether `DELETED` is also selected
* watch other resources


fluent-plugin-events/test/resources/api_list_events_v1.json
```diff
@@ -0,0 +1,171 @@
+{
+    "kind": "EventList",
```
samjsong: Is there no event type like `ADDED` or `MODIFIED` specified in the `EventList` resource type?

fluent-plugin-events/test/plugin/test_in_events.rb
```diff
@@ -1,17 +1,99 @@
...
+    config = %([])
+    driver = create_driver(config).instance
+    driver.instance_variable_set(:@client, @client)
+    mock_watch_events
+
+    selected_events_count = get_watch_resources_count_by_type_selector(["ADDED", "MODIFIED"],
+      'api_watch_events_v1.txt')
+    driver.router.expects(:emit).times(selected_events_count).with(anything, anything, anything)
+    events = driver.start_watcher_thread
+    sleep 5
```
samjsong: Is there some way to do `eventually` in these tests rather than `sleep 5`? I assume it's probably fine to leave it as is, but unless it takes a long time to change I think `eventually` might be better

fluent-plugin-events/test/plugin/test_in_events.rb
```diff
@@ -1,17 +1,99 @@
...
+    assert_equal events.length, 9
```
samjsong: Just wanted to clarify, so `events.length` is the number of events that the watcher thread saw, but line 58 `driver.router.expects(:emit).times(selected_events_count)` is the thing that tests that we only emitted the `ADDED` and `MODIFIED` events, right?

fluent-plugin-events/test/resources/api_list_events_v1.json
maimaisie: `Event` is a resource type, and `ADDED` or `MODIFIED` types are only for change of resources, eg. change of events, change of endpoints etc. This is also why you don't see `ADDED` or `MODIFIED` in `EndpointList`, but only when you watch for endpoint events (changes)

samjsong: Thanks for clarifying Maisie! I appreciate the help

fluent-plugin-events/test/helper.rb
```diff
@@ -3,6 +3,54 @@
...
+
+def get_watch_resources_count_by_type_selector(type_selector, file_name)
+  text = File.read(test_resource(file_name))
+  objects = text.split(/\n+/).map {|line| JSON.parse(line)}
+  objects.select {|object| type_selector.any? {|type| type.casecmp?(object['type'])}}.count
+end
+
+def init_globals
+  @kubernetes_url = 'http://localhost:8080'
+  @apiVersion = 'v1'
```
maimaisie: `apiVersion` was renamed to `api_version` though?

fluent-plugin-events/test/plugin/test_in_events.rb
```diff
@@ -1,17 +1,99 @@
...
 class EventsInputTest < Test::Unit::TestCase
-  setup do
-    Fluent::Test.setup
-  end
+  include SumoLogic::Kubernetes::Connector
 
-  test "failure" do
+  def setup
+    # runs before each test
+    @api_version = 'v1'
```
maimaisie: `init_globals` also initializes the api version

fluent-plugin-events/test/helper.rb
samjsong: Looks like we'll have to update this in the `enhance_k8s_metadata` test helper file as well

fluent-plugin-events/test/plugin/test_in_events.rb
yuting-liu: That's right.

fluent-plugin-events/test/plugin/test_in_events.rb
yuting-liu: `eventually` is really a good idea, however I didn't find something useful that can work in that way. Or maybe I missed something.
Typically, `expects(:method).times(...)` should work. However, it is way too flaky for this test case. And I chose to `sleep 5` to make sure it does verify `emit` is called. Not sure if things can get better/less flaky if I reduced the volume of events.

fluent-plugin-events/test/helper.rb
yuting-liu: Good catch. Going to fix this, as well as `enhance_k8s_metadata` test helper.

yuting-liu: BTW, it seems that for `enhance_k8s_metadata` test helper, `@apiVersion` or `@api_version` is no longer used there. so we can just removed it in `enhance_k8s_metadata` test helper?

samjsong: Ah, yeah looks like we can remove it 👍

yuting-liu:
@samjsong you are right. I think in some of our other test files (`test_cache_strategy.rb` and `test_reader.rb`), `assert` doesn't in the right order of `assert expected, actual` way. Maybe we should fix them all?

samjsong:
Good catch, I think we can fix those in another PR :)

PR #67, opened 2019-07-11
Fix order of assert_equal in test files

yuting-liu:
The correct order for `assert_equal` should be `assert_equal expected, actual`.
The order is incorrect for some test files.


yuting-liu:
@samjsong  Thanks. Just tried to ask your approval again 🤦‍♀

PR #73, opened 2019-07-15
Enhanced metadata for Logs

samjsong:
We expose a new `config_param` `data_type` which by default is `metrics`. If `logs`, then we attach the enhanced metadata to the `kubernetes` part of the log, and don't attach pod labels since those already exist.

Result is something like this:
```
{
  "timestamp": 1563232390248,
  "log": "2019-07-15 23:13:10 +0000 [info]: #0 fluentd worker is now running worker=0",
  "stream": "stdout",
  "time": "2019-07-15T23:13:10.248349213Z",
  "docker": {
    "container_id": "2caa84a192d156391087f632a9a0fd6458a0c283a5097e47a3cb8edf21a93003"
  },
  "kubernetes": {
    "container_name": "fluentd",
    "namespace_name": "sumologic",
    "pod_name": "fluentd-59d9c9656d-25v6z",
    "pod_id": "125d8c51-a756-11e9-b605-06604569b78c",
    "labels": {
      "k8s-app": "fluentd-sumologic",
      "pod-template-hash": "1585752128"
    },
    "host": "ip-123-45-67-890.us-west-1.compute.internal",
    "master_url": "https://100.64.0.1:443/api",
    "namespace_id": "229392a4-773a-11e9-b605-06604569b78c",
    "replicaset": "fluentd-59d9c9656d",
    "deployment": "fluentd",
    "service": "fluentd"
  }
}
```

Next step is to handle these metrics in `fluentd_kubernetes_sumologic` for `log_format="fields"` so that these plugins can send them to Sumo as headers (the above log was formatted with `log_format="json"`)


samjsong:
Merged master - last commit disables `service` enhanced metadata as it's disabled on master

PR #74, opened 2019-07-16
SUMO-113170 Support different k8s api versions

yuting-liu:
* Add support for different k8S api versions with multiple clients.
* Add unit tests.


fluent-plugin-events/lib/sumologic/kubernetes/connector.rb
```diff
@@ -7,17 +7,41 @@ module Connector
...
+
+      def create_client(base, ver)
+        url = "#{@kubernetes_url}/#{base}"
+        log.info "create client with URL: #{url} and apiVersion: #{ver}"
+        client = Kubeclient::Client.new(
+          url, ver,
           ssl_options: ssl_options,
-          auth_options: auth_options
+          auth_options: auth_options,
+          as: :parsed
```
maimaisie: This should be removed, it wasn't set before for events plugin. I remember setting the result type to `parsed` on the client won't work because for some APIs (like event watch stream) we want the raw results.

yuting-liu:
`ci/build.sh` successfully run on my machine locally but seems failed here. Will investigate it.

fluent-plugin-events/lib/sumologic/kubernetes/connector.rb
```diff
@@ -7,17 +7,41 @@ module Connector
       K8_POD_CA_CERT = 'ca.crt'.freeze
       K8_POD_TOKEN = 'token'.freeze
 
-      def connect_kubernetes
-        @client = Kubeclient::Client.new(
-          @kubernetes_url, @api_version,
+      # Need different clients to access different API groups/versions
+      CORE_API_VERSIONS = ['v1'].freeze
+      API_GROUPS = ['events.k8s.io/v1beta1'].freeze
```
samjsong: I believe since we want to support watching changes to other resource types, we would want to include the other API groups like `apps/v1` and `extensions/v1beta1` in order to watch those. For instance, the `deployment` watch endpoint is `/apis/apps/v1/watch/deployments` according to the docs [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#watch-list-all-namespaces-32).

fluent-plugin-events/lib/sumologic/kubernetes/connector.rb
yuting-liu: I noticed this too. Already fixed this locally. forgot to push it when created this PR 😊

fluent-plugin-events/lib/sumologic/kubernetes/connector.rb
maimaisie: That is a good point. Actually we don't need to create the clients for all of the available API groups here, but just the one that user wants to get events from. For example, if user specifies the config as:
```
    <source>
      @type events
      resource_name deployments
      api_version apps/v1
    </source>
```
then we put `apps/v1` in `API_GROUPS`. If it is `api_version events.k8s.io/v1beta1` then we put `events.k8s.io/v1beta1` in `API_GROUPS`.

We should however always build the core v1 client since we use it for config maps

samjsong: Aha, this makes sense, I forgot we had exposed a config param for this. In that case should we only create the core v1 client, and the client specified by the `api_version` config param in `connector.rb` ?

yuting-liu: Oh, then I think we don't need to have this `API_GROUPS` then.

fluent-plugin-events/test/helper.rb
```diff
@@ -55,6 +71,7 @@ def get_watch_resources_count_by_type_selector(type_selector, file_name)
 
 def init_globals
   @kubernetes_url = 'http://localhost:8080'
+  # @api_version = 'v1'
```
samjsong: nit: we should get rid of this comment

fluent-plugin-events/lib/sumologic/kubernetes/connector.rb
```diff
@@ -7,17 +7,40 @@ module Connector
...
+
+      def core_clients
+        CORE_API_VERSIONS.map do |ver|
+          [ver, create_client('api', ver)]
+        end.to_h
+      end
+
+      # If @api_version is not v1, we will create client for other API groups/versions
+      def group_clients
+        [@api_version].map do |ver|
```
maimaisie: `@api_version` is default to v1, so this code would create two clients that both connect to v1. we should only create a second client if it the value is not v1

fluent-plugin-events/test/helper.rb
```diff
@@ -12,9 +13,24 @@ def test_resource(name)
   File.new("test/resources/#{name}")
 end
 
-def mock_get_events
+def stub_apis
```
maimaisie: can you explain how this works?

fluent-plugin-events/lib/sumologic/kubernetes/connector.rb
yuting-liu: The `@api_version` will be checked in `connect_kubernetes`. In that method, when `@api_version` is not equals to `v1`, `group_clients` method won't be called. So there won't be two clients then.
```
def connect_kubernetes
        @clients = @api_version == 'v1' ? core_clients : core_clients.merge(group_clients)
end
```

fluent-plugin-events/test/helper.rb
yuting-liu: The reason I added this because the unit test doesn't pass before. The reason is basically that the clients cannot be created successfully. I try to print out the object `@clients`. It looks like following:
```
clients {"v1"=>"2019-07-16 21:15:40 +0000 [error]: Failed to open TCP connection to localhost:8080 (Connection refused - connect(2) for \"localhost\" port 8080)", "events.k8s.io/v1beta1"=>"2019-07-16 21:15:40 +0000 [error]: Failed to open TCP connection to localhost:8080 (Connection refused - connect(2) for \"localhost\" port 8080)", "apps/v1"=>"2019-07-16 21:15:40 +0000 [error]: Failed to open TCP connection to localhost:8080 (Connection refused - connect(2) for \"localhost\" port 8080)", "extensions/v1beta1"=>"2019-07-16 21:15:40 +0000 [error]: Failed to open TCP connection to localhost:8080 (Connection refused - connect(2) for \"localhost\" port 8080)"}, clients type Hash
```
It will get the error of connection issues when creating clients. I think the reason why it won't cause problem locally is that I have `kubectl proxy --port=8080` running locally. So I added this to make sure there won't have this TCP connection issues.

maimaisie: hmm okay, I still don't know _how_ it fixes the issue? can you explain what `stub_apis` does?

yuting-liu: What `stub_apis` does here is to mock response for `client.api_valid?`. Otherwise, method `create_client` will be stuck here with the connection error as I mentioned before.

maimaisie: ah interesting. thanks for clarifying

yuting-liu:
Already tested the plugin, with api version `apps/v1` and resourceName `deployments`. Here is what we got from watch:
```
{  
   "type":"ADDED",
   "object":{  
      "kind":"Deployment",
      "apiVersion":"apps/v1",
      "metadata":{  
         "name":"fluentd-events",
         "namespace":"sumologic",
         "selfLink":"/apis/apps/v1/namespaces/sumologic/deployments/fluentd-events",
         "uid":"64720983-a984-11e9-80e5-025000000001",
         "resourceVersion":"2771498",
         "generation":1,
         "creationTimestamp":"2019-07-18T17:49:29Z",
         "labels":{  
            "k8s-app":"fluentd-sumologic"
         },
         "annotations":{  
            "deployment.kubernetes.io/revision":"1",
            "kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"apps/v1\",\"kind\":\"Deployment\",\"metadata\":{\"annotations\":{},\"labels\":{\"k8s-app\":\"fluentd-sumologic\"},\"name\":\"fluentd-events\",\"namespace\":\"sumologic\"},\"spec\":{\"selector\":{\"matchLabels\":{\"k8s-app\":\"fluentd-sumologic\"}},\"template\":{\"metadata\":{\"labels\":{\"k8s-app\":\"fluentd-sumologic\"}},\"spec\":{\"containers\":[{\"env\":[{\"name\":\"SUMO_ENDPOINT_EVENTS\",\"valueFrom\":{\"secretKeyRef\":{\"key\":\"endpoint-events\",\"name\":\"sumologic\"}}}],\"image\":\"yuting-test\",\"imagePullPolicy\":\"Never\",\"name\":\"fluentd-events\",\"resources\":{\"limits\":{\"cpu\":\"100m\",\"memory\":\"256Mi\"},\"requests\":{\"cpu\":\"100m\",\"memory\":\"256Mi\"}},\"volumeMounts\":[{\"mountPath\":\"/fluentd/etc/\",\"name\":\"config-volume\"},{\"mountPath\":\"/mnt/pos/\",\"name\":\"pos-files\"}]}],\"serviceAccountName\":\"fluentd\",\"volumes\":[{\"hostPath\":{\"path\":\"/var/run/fluentd-pos\",\"type\":\"\"},\"name\":\"pos-files\"},{\"configMap\":{\"name\":\"fluentd-events-config\"},\"name\":\"config-volume\"}]}}}}\n"
         }
      },
      "spec":{  
         "replicas":1,
         "selector":{  
            "matchLabels":{  
               "k8s-app":"fluentd-sumologic"
            }
         },
         "template":{  
            "metadata":{  
               "creationTimestamp":null,
               "labels":{  
                  "k8s-app":"fluentd-sumologic"
               }
            },
            "spec":{  
               "volumes":[  
                  {  
                     "name":"pos-files",
                     "hostPath":{  
                        "path":"/var/run/fluentd-pos",
                        "type":""
                     }
                  },
                  {  
                     "name":"config-volume",
                     "configMap":{  
                        "name":"fluentd-events-config",
                        "defaultMode":420
                     }
                  }
               ],
               "containers":[  
                  {  
                     "name":"fluentd-events",
                     "image":"yuting-test",
                     "env":[  
                        {  
                           "name":"SUMO_ENDPOINT_EVENTS",
                           "valueFrom":{  
                              "secretKeyRef":{  
                                 "name":"sumologic",
                                 "key":"endpoint-events"
                              }
                           }
                        }
                     ],
                     "resources":{  
                        "limits":{  
                           "cpu":"100m",
                           "memory":"256Mi"
                        },
                        "requests":{  
                           "cpu":"100m",
                           "memory":"256Mi"
                        }
                     },
                     "volumeMounts":[  
                        {  
                           "name":"config-volume",
                           "mountPath":"/fluentd/etc/"
                        },
                        {  
                           "name":"pos-files",
                           "mountPath":"/mnt/pos/"
                        }
                     ],
                     "terminationMessagePath":"/dev/termination-log",
                     "terminationMessagePolicy":"File",
                     "imagePullPolicy":"Never"
                  }
               ],
               "restartPolicy":"Always",
               "terminationGracePeriodSeconds":30,
               "dnsPolicy":"ClusterFirst",
               "serviceAccountName":"fluentd",
               "serviceAccount":"fluentd",
               "securityContext":{  

               },
               "schedulerName":"default-scheduler"
            }
         },
         "strategy":{  
            "type":"RollingUpdate",
            "rollingUpdate":{  
               "maxUnavailable":"25%",
               "maxSurge":"25%"
            }
         },
         "revisionHistoryLimit":10,
         "progressDeadlineSeconds":600
      },
      "status":{  
         "observedGeneration":1,
         "replicas":1,
         "updatedReplicas":1,
         "readyReplicas":1,
         "availableReplicas":1,
         "conditions":[  
            {  
               "type":"Available",
               "status":"True",
               "lastUpdateTime":"2019-07-18T17:49:32Z",
               "lastTransitionTime":"2019-07-18T17:49:32Z",
               "reason":"MinimumReplicasAvailable",
               "message":"Deployment has minimum availability."
            },
            {  
               "type":"Progressing",
               "status":"True",
               "lastUpdateTime":"2019-07-18T17:49:32Z",
               "lastTransitionTime":"2019-07-18T17:49:29Z",
               "reason":"NewReplicaSetAvailable",
               "message":"ReplicaSet \"fluentd-events-597cbf49bf\" has successfully progressed."
            }
         ]
      }
   }
}
```

maimaisie:
is this log ingested in Sumo?

yuting-liu:
@maimaisie Yea, the log is ingested in Sumo.

PR #76, opened 2019-07-16
fix automatic setup script command in deploy README

samjsong:


deploy/README.md
```diff
@@ -66,7 +66,7 @@ This approach requires access to the Sumo Logic Collector API. It will create a
 
 ```sh
 curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/kubernetes/setup.sh \
-  | bash -s [-c collector-name] [-k cluster-name] <api-endpoint> <access-id> <access-key>
+  | bash -s - [-c collector-name] [-k cluster-name] [-n namespace] <api-endpoint> <access-id> <access-key>
```
lei-sumo: why do we need the extra bar here?

samjsong: I believe the extra bar tells the `bash` command that the result from the pipe stdin should be inserted in the command there. Otherwise the other arguments `-c -k -n` are interpreted as arguments to the `bash` command

lei-sumo: ah, makes sense.

PR #77, opened 2019-07-16
Make imagePullPolicy and latest tag explicit for now

rvmiller89:
Until we start tagging our `sumologic/kubernetes-fluentd` releases, we should be explicit about the `Always` imagePullPolicy and `0.0.0` tag.

Once we start tagging, we can change this to:
```
image: sumologic/kubernetes-fluentd:v1.blah
imagePullPolicy: IfNotPresent
```

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
+++ b/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
@@ -285,7 +285,7 @@ spec:
           name: fluentd-config
       containers:
       - name: fluentd
-        image: sumologic/kubernetes-fluentd:latest
+        image: sumologic/kubernetes-fluentd:0.0.0
         imagePullPolicy: Always
         resources:
           limits:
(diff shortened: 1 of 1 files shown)


rvmiller89:
Do we overwrite the previous tagged image `0.0.0` each time? I guess I don't really know how tags work. Should we be incrementing instead, like `0.0.1`, `0.0.2`, etc. so that each version has a unique tag?

samjsong:
Yeah since we didn't start tagging anything we've been overriding the version `0.0.0` every time we push something new... I believe the plan was to start incrementing in that way once events and enhanced metadata were complete, starting with `1.0.0`

samjsong:
In retrospect it would've been better to start incrementing much sooner, so we have a history of changes...

blaise-sumo:
FWIW, images can have multiple tags and it is really nice if at least one
of them is the first 7 or 8 chars of the git commit Sha.

PR #82, opened 2019-07-18
Cleanup Dockerfile and remove Dockerfile-debian

rvmiller89:
We're no longer using `fluent-plugin-carbon-v2` so we can skip installing that (although our CI script will still build the gem since the directory is in this repo).

We were also installing the local `fluent-plugin-sumologic_output` over the gem file from RubyGems, so this removes that extra replacement.

Lastly we want to focus on Dockerfile and remove Dockerfile-debian.


deploy/docker/Dockerfile
```diff
@@ -26,9 +27,8 @@ RUN gem install fluent-plugin-s3 -v 1.1.4 \
        && gem install fluent-plugin-prometheus -v 1.1.0 \
        && gem install fluent-plugin-kubernetes_sumologic
```
rvmiller89: actually, is this correct? Are we missing a version here? In fact, are we using this gem anywhere? This is https://github.com/SumoLogic/fluentd-kubernetes-sumologic, right?

frankreno: I missed that, you are right. We should include the version explicitly.  WE are using it

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L179

frankreno: we should use 2.4.1

lei-sumo: Why are we using this plugin? This is the fluend daemonset which should be removed from here as we have fluentbit now.

PR #87, opened 2019-07-24
Upgrade Dockerfile fluentd to v1.6.2

rvmiller89:


deploy/docker/Dockerfile
```diff
@@ -64,6 +64,7 @@ ENV VERIFY_SSL "true"
 
 RUN mkdir -p /fluentd/conf.d
 
-COPY ./fluent.conf /fluentd/conf.d/fluent.conf
+COPY ./fluent.conf /fluentd/etc/
+COPY entrypoint.sh /bin/
```
maimaisie: should we test this with a local image first to see if helps with the current bug? or do you think we should upgrade regardless?

lei-sumo: What are we trying to do here with the upgrade? Is this an alpine or debian build? I remember we had some issue with debian build

rvmiller89: Yeah, I'm testing currently to see if it helps with the other issue.  But I think we should merge this regardless.  I have verified fluentd is starting correctly.

@lei-sumo We're upgrading to the latest release of the Alpine docker image from FluentD, since there have been some bug fixes since 1.4.2: https://github.com/fluent/fluentd/blob/master/CHANGELOG.md

PR #91, opened 2019-07-26
Fix ServiceMonitor docs for missing label and add note about using `--force` with `helm upgrade.`

frankreno:
Fix ServiceMonitor docs for missing label and add note about using `--force` with `helm upgrade.`


frankreno:
A customer ran into the helm upgrade issue.  Workaround came from here:

https://github.com/helm/charts/blob/master/stable/prometheus-operator/README.md#upgrade

PR #96, opened 2019-07-27
Push alpha docker image on merge push CI

rvmiller89:
When docker password env variable is present, and we are pushing to master (merging a PR), this will push a new docker image with auto-incrementing patch number.

It will use the latest non-alpha version to determine the next patch number to use and append `-alpha` to the version.

Example output:
```
Most recent release version: 0.0.0, most recent alpha: 0.0.0-alpha
Pushing alpha docker image with version 0.0.1-alpha
```

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -4,6 +4,7 @@ VERSION="${TRAVIS_TAG:-0.0.0}"
 VERSION="${VERSION#v}"
 : "${DOCKER_TAG:=sumologic/kubernetes-fluentd}"
 : "${DOCKER_USERNAME:=sumodocker}"
+DOCKER_TAGS="https://registry.hub.docker.com/v1/repositories/sumologic/kubernetes-fluentd/tags"
 
 echo "Starting build process in: `pwd` with version tag: $VERSION"
 set -e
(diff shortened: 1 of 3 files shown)


ci/build.sh
```diff
@@ -41,12 +42,29 @@ cd ../..
...
+      echo "Most recent release version: $latest_release, most recent alpha: $latest_alpha-alpha"
+    else
+      latest_alpha="$latest_release"
+      echo "Most recent release version: $latest_release, most recent alpha does not yet exist"
+    fi
+    patch=$((`echo $latest_alpha | tr '.' $'\n' | sed -n 3p`+1))
+    new_alpha="$latest_major.$latest_minor.$patch-alpha"
+    echo "Pushing alpha docker image with version $new_alpha"
+    echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+    docker push $DOCKER_TAG:$new_alpha
```
rvmiller89: TODO:  Docker tag before push

ci/build.sh
```diff
@@ -41,12 +42,29 @@ cd ../..
...
 ruby deploy/test/test_docker.rb
 
-if [ -z "$DOCKER_PASSWORD" ] || [ -z "$TRAVIS_TAG" ]; then
-    echo "Skip Docker pushing"
-else
+if [ -n "$DOCKER_PASSWORD" ] && [ -n "$TRAVIS_TAG" ]; then
     echo "Pushing docker image with $DOCKER_TAG:$VERSION ..."
     echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
     docker push $DOCKER_TAG:$VERSION
+elif [ -n "$DOCKER_PASSWORD" ] && [ "$TRAVIS_BRANCH" == "master" ]; then
```
rvmiller89: TODO: 2nd CI still sees this as attempt to to push to master: https://travis-ci.org/SumoLogic/sumologic-kubernetes-collection/builds/564260193?utm_source=github_status&utm_medium=notification

rvmiller89: Also needs check "$TRAVIS_EVENT_TYPE" == "push"

ci/build.sh
```diff
@@ -32,21 +33,43 @@ for i in ./fluent-plugin* ; do
...
+  latest_major=`echo $latest_release | tr '.' $'\n' | sed -n 1p`
+  latest_minor=`echo $latest_release | tr '.' $'\n' | sed -n 2p`
+  latest_alpha=`wget -q $DOCKER_TAGS -O - | jq -r .[].name | grep alpha | grep "^$latest_major.$latest_minor" | sed 's/-alpha//g' | sort --version-sort --field-separator=. | tail -1`
+  if [ -n "$latest_alpha" ]; then
+    echo "Most recent release version: $latest_release, most recent alpha: $latest_alpha-alpha"
+  else
+    echo "Most recent release version: $latest_release, most recent alpha does not yet exist"
+    latest_alpha="$latest_release"
+  fi
+  new_patch=$((`echo $latest_alpha | tr '.' $'\n' | sed -n 3p`+1))
```
samjsong: In the case that we bumped the major or minor version, we could have a case like this: `latest_release=1.0.0`, `latest_alpha=0.0.10` in which case I believe `new_alpha` would be `1.0.11-alpha`, but we probably want to start the patch count from 0 again

rvmiller89: L56 also does `grep "^$latest_major.$latest_minor"` to find the most recent alpha for the most recent Major.minor release, so it will not find an alpha for `latest_release=1.0.0` and below it will create `new_alpha=1.0.1-alpha`.

samjsong: Ah gotcha, I missed that part. Thanks for clarifying.

PR #99, opened 2019-07-29
Parameterize buffer settings for Fluentd Logs/Metrics

rvmiller89:
- Raise number of threads per output plugin to 4 threads to alleviate http bottleneck
- Apply same buffer configuration settings to both logs and metrics output plugins
- Specify `gzip` compression for fluentd to use internally when buffering to maximize amount of buffering we can do
- Parameterize chunk_limit_size (http payload size) with 100KB max, and total_limit_size (amount of memory used per plugin) with 128MB max

Tested bringing up Fluentd pods and ensured log and metric collection is working.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -43,7 +43,9 @@ RUN gem sources --clear-all \
 # Default settings for log collection
 ENV LOG_FORMAT "fields"
 ENV FLUSH_INTERVAL "5s"
-ENV NUM_THREADS "1"
+ENV NUM_THREADS "4"
+ENV CHUNK_LIMIT_SIZE "100k"
+ENV TOTAL_LIMIT_SIZE "128m"
 ENV SOURCE_CATEGORY "%{namespace}/%{pod_name}"
 ENV SOURCE_CATEGORY_PREFIX "kubernetes/"
 ENV SOURCE_CATEGORY_REPLACE_DASH "/"
--- a/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
+++ b/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
@@ -90,6 +90,7 @@ data:
       data_type metrics
       metric_data_format prometheus
       disable_cookies true
+      @include buffer.output.conf
     </match>
     <match prometheus.datapoint.kubelet**>
       @type sumologic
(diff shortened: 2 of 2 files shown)


deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -260,13 +267,21 @@ data:
       endpoint "#{ENV['SUMO_ENDPOINT_LOGS']}"
       verify_ssl "#{ENV['VERIFY_SSL']}"
       log_format "#{ENV['LOG_FORMAT']}"
-      flush_interval "#{ENV['FLUSH_INTERVAL']}"
-      num_threads "#{ENV['NUM_THREADS']}"
-      open_timeout 60
       add_timestamp "#{ENV['ADD_TIMESTAMP']}"
       timestamp_key "#{ENV['TIMESTAMP_KEY']}"
       proxy_uri "#{ENV['PROXY_URI']}"
+      @include buffer.output.conf
```
lei-sumo: This is a nice one. Can we also do some similar abstraction for other configurations? e.g.       
      data_type metrics
      metric_data_format prometheus
      disable_cookies true

deploy/docker/Dockerfile
```diff
@@ -43,7 +43,9 @@ RUN gem sources --clear-all \
 # Default settings for log collection
 ENV LOG_FORMAT "fields"
 ENV FLUSH_INTERVAL "5s"
-ENV NUM_THREADS "1"
+ENV NUM_THREADS "4"
```
lei-sumo: Does it mean every plugin will have 4 threads? Does it harm if we set it to an even higher value?

rvmiller89: Yes it does.  I picked 4 because that's what the collector uses.  I recommend we try it for a while, work on instrumenting some metrics so we can see queue size, and only increase it once we can observe the effects.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
rvmiller89: Sure I can DRY this up more.  It's too bad fluentd only supports `include` and not macro templates that we could parameterize.

rvmiller89: Done, can you approve again ?

lei-sumo: LGTM. thanks

PR #102, opened 2019-07-30
Added Falco specific changes.

arunpatyal:
Added- overrides.yaml for Falco and Falco GKE. Updated Read me for Falco.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -353,7 +353,7 @@ helm repo update \
 
 ## Step 4: Deploy Falco
 
-In this step, you will deploy [Falco](https://falco.org/) to detect anomalous activity and capture Kubernetes Audit Events.
+In this step, you will deploy [Falco](https://falco.org/) to detect anomalous activity and capture Kubernetes Audit Events. This step is required only if you intend to use the Sumo Logic Kubernetes App.
 
 __NOTE__ [Falco](https://sysdig.com/blog/sysdig-falco/) needs privileged container access to insert its kernel module to process events for system calls.
 
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -291,7 +291,7 @@ spec:
   endpoints:
   - port: web
   ```
-  
+
 Replace the `name` with a name that relates to your service, and a `matchLabels` that would match the pods you want this service monitor to scrape against. By default, prometheus attempts to scrape metrics off of the `/metrics` endpoint, but if you do need to use a different url, you can override it by providing a `path` attribute in the settings like so:
 
 ```
@@ -329,7 +329,7 @@ The `prometheus-overrides.yaml` file controls what metrics get forwarded on to S
 After adding this to the `yaml`, go ahead and run a `helm upgrade prometheus-operator stable/prometheus-operator -f prometheus-overrides.yaml` to upgrade your `prometheus-operator`.
 
 Note: When executing the helm upgrade to avoid the error below is need add the argument `--force`.
-      
+
       invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/name":"kube-state-metrics"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
 
 If all goes well, you should now have your custom metrics piping into Sumo Logic.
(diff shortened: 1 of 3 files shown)


deploy/README.md
```diff
@@ -351,8 +351,49 @@ helm repo update \
    && helm install stable/fluent-bit --name fluent-bit --namespace sumologic -f fluent-bit-overrides.yaml
 ```
 
+## Step 4: Deploy Falco
+
+In this step, you will deploy Falco to detect anomalous activity and capture Kubernetes Audit Events.
+
+__NOTE__ Falco needs privileged container access to insert its kernel module to process events for system calls.
```
frankreno: Can we link to some Falco documentation that talks to this?

deploy/README.md
```diff
@@ -351,8 +351,49 @@ helm repo update \
...
+```
+
+Install `falco` using Helm:
+
+```sh
+helm repo update \
+   && helm install stable/falco --name falco --namespace sumologic -f falco-overrides.yaml
+```
+
+__NOTE__ `Google Kubernetes Engine (GKE)` uses Container-Optimized OS (COS) as the default operating system for its worker node pools. COS is a security-enhanced operating system that limits access to certain parts of the underlying OS. Because of this security constraint, Falco cannot insert its kernel module to process events for system calls. However, COS provides the ability to leverage eBPF (extended Berkeley Packet Filter) to supply the stream of system calls to the Falco engine. eBPF is currently supported only on GKE and COS.
```
frankreno: Might be good to put a link to the official Falco docs here as well so customers can get more information.

deploy/helm/falco-gke-overrides.yaml
```diff
@@ -0,0 +1,5 @@
+ebpf:
+  # Enable eBPF support for Falco
```
frankreno: I think we should have 1 YAML with the EBPF stuff commented by default and the comment can indicate this needs to be uncommented for GKE.  This will help us avoid having to maintain multiple files.

deploy/README.md
lei-sumo: What's the reason to add it to our collection process? Does our solution rely on it?

deploy/README.md
```diff
@@ -351,8 +351,49 @@ helm repo update \
    && helm install stable/fluent-bit --name fluent-bit --namespace sumologic -f fluent-bit-overrides.yaml
 ```
 
+## Step 4: Deploy Falco
```
rvmiller89: Should we say `Step 4: (Optional) Deploy Falco`

deploy/README.md
arunpatyal: > Can we link to some Falco documentation that talks to this?

There is no detailed documentation around the privileged container access. This is the most relevant page I have found, https://sysdig.com/blog/sysdig-falco/. In the section, "how Falco works" it talks a bit about it. I will link this one.

arunpatyal: > What's the reason to add it to our collection process? Does our solution rely on it?

Our collection doesn't depend on it. Falco provides additional information in addition to what we are doing, specifically behavioral activity to detect anomalous activity. Falco can detect and alert on any behavior that involves making Linux system calls. In addition, it can also retrieve k8s audit events.

deploy/README.md
arunpatyal: Will do.

deploy/helm/falco-gke-overrides.yaml
arunpatyal: Good idea. Will change.

deploy/README.md
arunpatyal: Maybe, but we have dashboards dependent on this in the core Kubernetes app.

deploy/README.md
arunpatyal: done

deploy/helm/falco-gke-overrides.yaml
arunpatyal: done

deploy/README.md
samjsong: Since Falco is not required for collection, but is required for the Kubernetes app, can we add something that communicates this in the doc? Maybe something like

`"This step is required for users who wish to use the Sumo Logic Kubernetes App."`

on line 356 would help make this clear.

arunpatyal: Makes sense, I was also thinking the same.

arunpatyal: How about this: "This step is required only if you intend to use the Sumo Logic Kubernetes App."

rvmiller89: I'm fine with that.

arunpatyal: Thanks, changed.

PR #103, opened 2019-07-30
Auto-tag alpha images in Git that we can generate full releases from

rvmiller89:
Automatically tag pushes to master with the same version we use to push the docker image. From these tags we can create releases in Github, and also perform diffs between tags/master.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -68,6 +68,10 @@ elif [ -n "$DOCKER_PASSWORD" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS
   echo "Pushing alpha docker image with version $new_alpha"
   echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
   docker push $DOCKER_TAG:$new_alpha
+
+  echo "Tagging git with v$new_alpha..."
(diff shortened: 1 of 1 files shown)


rvmiller89:
This is difficult to test without actually merging and seeing how it works.  If it messes up, we can delete the tag.

ci/build.sh
```diff
@@ -68,6 +68,10 @@ elif [ -n "$DOCKER_PASSWORD" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS
   echo "Pushing alpha docker image with version $new_alpha"
   echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
   docker push $DOCKER_TAG:$new_alpha
+
+  echo "Tagging git with v$new_alpha..."
+  git tag -a "v$new_alpha" -m "Bump version to v$new_alpha"
+  git push --tags origin master
```
lei-sumo: Would this create too many tags in github? Do we have any concern there?

rvmiller89: It will create a tag for every PR that gets merged. I think that's fine though -- we can always re-evaluate if we hit some limit.

samjsong: +1 I'm okay with this, it helps keep track of which alpha tag corresponds to which commits as well

PR #106, opened 2019-07-30
SUMO-115364 Handle exceptions thrwon from kubeclient calls

yuting-liu:
Since we are using `timer_execute` for `start_monitor`, any exceptions thrown from the method will cause the timer detached.

We are going to add exceptions handling for all methods that have kubeclient calls in events plugin, including following methods:
```
start_watcher_thread
create_config_map
patch_config_map
initialize_resource_version
pull_resource_version
```
We will only rescue possible errors and then log the error so that we can know errors from which method they throw. For all other methods except for `start_watch_thread`, we don't want errors from kubeclient blocks the thread of watching, so we just log the error and ignore it. For `start_watcher_thread`, we need `timer_execute` continues working so exceptions will be swallowed and timer won't be detached in this case.

### DIFF
--- a/fluent-plugin-events/test/plugin/test_in_events.rb
+++ b/fluent-plugin-events/test/plugin/test_in_events.rb
@@ -24,18 +24,22 @@ class EventsInputTest < Test::Unit::TestCase
     driver = Fluent::Test::Driver::Input.new(Fluent::Plugin::EventsInput).configure(conf)
   end
 
+  def configure_test_driver(config = %([]))
+    driver = create_driver(config).instance
+    connect_kubernetes_with_api_version(driver)
+    driver.instance_variable_set(:@clients, @clients)
+    driver
+  end
+
   def connect_kubernetes_with_api_version(driver)
     @api_version = driver.instance_variable_get(:@api_version)
     connect_kubernetes
   end
 
   test 'pull_resource_version correctly from eventlist' do
-    config = %([])
-    driver = create_driver(config).instance
-    connect_kubernetes_with_api_version(driver)
-    driver.instance_variable_set(:@clients, @clients)
-
+    driver = configure_test_driver()
     mock_get_events('api_list_events_v1.json')
+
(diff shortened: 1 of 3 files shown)


fluent-plugin-events/test/plugin/test_in_events.rb
```diff
@@ -186,4 +186,65 @@ def connect_kubernetes_with_api_version(driver)
     
     driver.start_monitor
   end
+
+  sub_test_case 'exception from kubeclient will not interrupt execution of start monitor' do
```
maimaisie: `create_config_map` and `initialize_resource_version` are not called within `start_monitor` right?

fluent-plugin-events/test/plugin/test_in_events.rb
```diff
@@ -186,4 +186,65 @@ def connect_kubernetes_with_api_version(driver)
...
+    test 'exception from kubeclient call in create_config_map' do
+      config = %([])
+      driver = create_driver(config).instance
+      connect_kubernetes_with_api_version(driver)
+      driver.instance_variable_set(:@clients, @clients)
+      driver.instance_variable_set(:@last_recreated, 0)
```
maimaisie: is this needed in test cases that don't call `start_monitor`? Also it looks like these 5 lines are being repeated in almost every test case, maybe we need to DRY or need something similar to the `beforeEach()` in scala

yuting-liu: Good point. Going to refactor this.

fluent-plugin-events/test/plugin/test_in_events.rb
yuting-liu: Yes. Need to change this description then.

yuting-liu:
Refactored the part of creating the test driver with configuration and instance variable setup into a method `configure_test_driver(config = %([]))`.

PR #107, opened 2019-07-30
Move Dockerfile env variables to yaml file to avoid drift

rvmiller89:
We currently have an issue where customers can use the latest yaml file from master branch (e.g. from setup.sh script) but it can reference environment variables in newer docker images so they won't get the default value.

To avoid drift, we would like to centralize this configuration in the yaml file itself.


rvmiller89:
Frank and Lei just told me that is not the plan -- they want them separated as `env` variables

samjsong:
Gotcha, I must've misunderstood.

PR #108, opened 2019-07-31
Add monitoring for FluentD plugins

rvmiller89:
Exposes Prometheus metrics (currently just output plugin metrics) for fluentd and fluentd-events deployments.
Screen Shot 2019-07-31 at 5 11 50 PM

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
+++ b/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
@@ -57,6 +57,8 @@ data:
   common.conf: |-
     <source>
       @type prometheus
+      metrics_path /metrics
+      port 24231
     </source>
     <source>
       @type monitor_agent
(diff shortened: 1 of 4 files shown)


fluent-plugin-protobuf/lib/fluent/plugin/parser_protobuf.rb
```diff
@@ -4,6 +4,8 @@
 require_relative '../../types_pb'
 require_relative '../../remote_pb'
 
+WriteRequest = Prometheus::WriteRequest
```
lei-sumo: Do you really need this change?

rvmiller89: Yes, otherwise our `module Prometheus` conflicts with the fluentd-prometheus plugin's `module Prometheus` and our parsing breaks.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -511,4 +560,23 @@ spec:
...
+  name: fluentd-events
+  labels:
+    k8s-app: fluentd-sumologic-events
+spec:
+  selector:
+    k8s-app: fluentd-sumologic-events
+  ports:
+  - name: metrics
+    port: 24231
+    targetPort: 24231
```
frankreno: Not needed for now, but we should look at using named ports for the targetPort so we dont have to update the Pod and Service each time.

PR #109, opened 2019-07-31
Add param for alpha docker images

Gourav2906:
Add is an alpha flag to set alpha fluentD image

### DIFF
--- a/deploy/kubernetes/setup.sh
+++ b/deploy/kubernetes/setup.sh
@@ -193,7 +193,7 @@ kubectl -n $NAMESPACE create secret generic sumologic \
 echo "Applying deployment 'fluentd'... $release"
 curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/v$release/deploy/kubernetes/fluentd-sumologic.yaml.tmpl | \
 sed 's/\$NAMESPACE'"/$NAMESPACE/g" | \
-sed 's|image: sumologic/kubernetes-fluentd:.*|image: sumologic/kubernetes-fluentd:$release|g' | \
+sed 's|image: sumologic/kubernetes-fluentd:.*|image: sumologic/kubernetes-fluentd:"$release"|g' | \
 tee fluentd-sumologic.yaml | \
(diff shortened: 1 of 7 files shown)


deploy/kubernetes/setup.sh
```diff
@@ -182,9 +190,10 @@ kubectl -n $NAMESPACE create secret generic sumologic \
   --from-literal=endpoint-logs=$ENDPOINT_LOGS \
   --from-literal=endpoint-events=$ENDPOINT_EVENTS
 
-echo "Applying deployment 'fluentd'..."
+echo "Applying deployment 'fluentd'... $release"
 curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/kubernetes/fluentd-sumologic.yaml.tmpl | \
```
rvmiller89: could we also change this URL? Now that we automatically tag our releases, we can make the URL like:

curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/v$release/deploy/kubernetes/fluentd-sumologic.yaml.tmpl | \

Note:  requires "v" added before $release

deploy/kubernetes/setup.sh
```diff
@@ -182,9 +190,10 @@ kubectl -n $NAMESPACE create secret generic sumologic \
...
 sed 's/\$NAMESPACE'"/$NAMESPACE/g" | \
+sed 's/\0.0.0'"/$release/g" | \
```
samjsong: I think we'd eventually want to update the master yaml to also reflect the latest stable version, in which case we'd have to remember to change it here as well, not sure how I feel about that

rvmiller89: Agreed, I'm going to assume that we will bump the version in the master Yaml every so often.  Could we try to replace this based on a regex search, or at least looking for `image: sumologic/kubernetes-fluentd:` ?

Gourav2906:
Messed up here , created new PR 
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/113

PR #112, opened 2019-07-31
enable fluent-bit metrics collection

frankreno:
This PR enables Fluent-Bit metrics, creates a serviceMonitor with the appropriate label so Prometheus discovers, and then adds the appropriate remote write configuration.


frankreno:
Yup, tested on my cluster and worked perfectly.

PR #113, opened 2019-08-01
Add param for alpha docker images

Gourav2906:
Add is an alpha flag to set alpha fluentD image

### DIFF
--- a/deploy/kubernetes/setup.sh
+++ b/deploy/kubernetes/setup.sh
@@ -4,7 +4,7 @@ set -e
 usage() {
   echo
   echo 'Usage:'
-  echo '  setup.sh [-c collector-name] [-k cluster-name] [-n namespace] [-a isAlpha] <endpoint> <access-id> <access-key>'
+  echo '  setup.sh [-c collector-name] [-k cluster-name] [-n namespace] [-a useAlpha] <endpoint> <access-id> <access-key>'
   echo
 }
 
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/setup.sh
```diff
@@ -182,9 +190,10 @@ kubectl -n $NAMESPACE create secret generic sumologic \
   --from-literal=endpoint-logs=$ENDPOINT_LOGS \
   --from-literal=endpoint-events=$ENDPOINT_EVENTS
 
-echo "Applying deployment 'fluentd'..."
-curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/kubernetes/fluentd-sumologic.yaml.tmpl | \
+echo "Applying deployment 'fluentd'... $release"
```
samjsong: nit: For consistency, can we put the version before the `...` ?
```
echo "Applying deployment 'fluentd' $release..."
```

deploy/kubernetes/setup.sh
```diff
@@ -4,7 +4,7 @@ set -e
 usage() {
   echo
   echo 'Usage:'
-  echo '  setup.sh [-c collector-name] [-k cluster-name] [-n namespace] <endpoint> <access-id> <access-key>'
+  echo '  setup.sh [-c collector-name] [-k cluster-name] [-n namespace] [-a isAlpha] <endpoint> <access-id> <access-key>'
```
rvmiller89: nit: rather than `isAlpha` can we call this `useAlpha`?

Also:  In this same PR can we update the deployment README with the new parameter: https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/master/deploy#automatic-source-creation-and-setup-script

PR #118, opened 2019-08-02
Helm Chart initial commit

maimaisie:
Create the Helm chart. Dependencies and automation of yaml generation will be in separate PRs.

Output of `helm install`:
```
$ helm install helm/sumologic
NAME:   giggly-pike
LAST DEPLOYED: Fri Aug  2 11:24:49 2019
NAMESPACE: default
STATUS: DEPLOYED

RESOURCES:
==> v1/ConfigMap
NAME                                 DATA  AGE
giggly-pike-sumologic-config         9     1s
giggly-pike-sumologic-events-config  2     1s

==> v1/Deployment
NAME                          READY  UP-TO-DATE  AVAILABLE  AGE
giggly-pike-sumologic         0/3    3           0          0s
giggly-pike-sumologic-events  0/1    1           0          0s

==> v1/Pod(related)
NAME                                           READY  STATUS             RESTARTS  AGE
giggly-pike-sumologic-d5dc9d85d-9ptld          0/1    Pending            0         0s
giggly-pike-sumologic-d5dc9d85d-b6sxd          0/1    Pending            0         0s
giggly-pike-sumologic-d5dc9d85d-kfhc2          0/1    Pending            0         0s
giggly-pike-sumologic-events-5db58c487d-9f722  0/1    ContainerCreating  0         0s

==> v1/Service
NAME                          TYPE       CLUSTER-IP     EXTERNAL-IP  PORT(S)                       AGE
giggly-pike-sumologic         ClusterIP  10.101.92.220  <none>       9888/TCP,24321/TCP,24231/TCP  0s
giggly-pike-sumologic-events  ClusterIP  10.99.214.103  <none>       24231/TCP                     0s

==> v1/ServiceAccount
NAME                   SECRETS  AGE
giggly-pike-sumologic  1        1s

==> v1beta1/ClusterRole
NAME                   AGE
giggly-pike-sumologic  1s

==> v1beta1/ClusterRoleBinding
NAME                   AGE
giggly-pike-sumologic  0s
```

### DIFF
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
+name: sumologic
+version: 0.1.0
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 appVersion: "1.0"
-description: A Helm chart for Kubernetes
+description: A Helm chart for collecting Kubernetes logs, metrics and events into Sumo Logic.
 name: sumologic
 version: 0.1.0
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.0"
+appVersion: "0.0.0"
 description: A Helm chart for collecting Kubernetes logs, metrics and events into Sumo Logic.
 name: sumologic
 version: 0.1.0
--- a/deploy/helm/sumologic/conf/events/events.conf
+++ b/deploy/helm/sumologic/conf/events/events.conf
@@ -1,0 +1,29 @@
+<source>
+  @type prometheus
+  metrics_path /metrics
+  port 24231
+</source>
+<source>
+  @type prometheus_output_monitor
+</source>
+<source>
+  @type events
+  deploy_namespace $NAMESPACE
+</source>
+<match kubernetes.**>
+  @type sumologic
+  @id sumologic.endpoint.events
+  endpoint "#{ENV['SUMO_ENDPOINT_EVENTS']}"
+  data_type logs
+  disable_cookies true
+  verify_ssl "#{ENV['VERIFY_SSL']}"
+  proxy_uri "#{ENV['PROXY_URI']}"
+  <buffer>
+    @type memory
+    compress gzip
+    flush_interval "#{ENV['FLUSH_INTERVAL']}"
+    flush_thread_count "#{ENV['NUM_THREADS']}"
+    chunk_limit_size "#{ENV['CHUNK_LIMIT_SIZE']}"
+    total_limit_size "#{ENV['TOTAL_LIMIT_SIZE']}"
+  </buffer>
+</match>
--- a/deploy/helm/sumologic/conf/events/events.conf
+++ b/deploy/helm/sumologic/conf/events/events.conf
@@ -8,7 +8,7 @@
 </source>
 <source>
   @type events
-  deploy_namespace $NAMESPACE
+  deploy_namespace {{ .Release.Namespace }}
 </source>
 <match kubernetes.**>
   @type sumologic
--- a/deploy/helm/sumologic/conf/events/events.conf
+++ b/deploy/helm/sumologic/conf/events/events.conf
@@ -1,3 +1,18 @@
+{{- if .Values.watchResourceEventsOverrides }}
+{{- range $name, $version := .Values.watchResourceEventsOverrides }}
+<source>
+  @type events
+  deploy_namespace {{ $.Release.Namespace }}
+  resource_name {{ $name }}
+  api_version {{ $version }}
+</source>
+{{- end }}
+{{- else }}
+<source>
+  @type events
+  deploy_namespace {{ .Release.Namespace }}
+</source>
+{{- end }}
 <source>
   @type prometheus
   metrics_path /metrics
@@ -5,10 +20,6 @@
 </source>
 <source>
   @type prometheus_output_monitor
-</source>
-<source>
-  @type events
-  deploy_namespace {{ .Release.Namespace }}
 </source>
 <match kubernetes.**>
   @type sumologic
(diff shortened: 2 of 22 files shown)


rvmiller89:
```
$ helm install helm/sumologic
NAME:   giggly-pike
```
what is `giggly-pike` ?

maimaisie:
That is a randomly generated release name. You can override release name by passing in the `--name` flag to helm install

rvmiller89:
Can we provide a default release name to avoid this without having the customer specify `--name` flag?

deploy/helm/sumologic/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+  tag: 0.0.0
+  pullPolicy: Always
+
+nameOverride: ""
+fullnameOverride: ""
+
+deployment:
+  replicaCount: 3
+  resources: 
+    # We usually recommend not to specify default resources and to leave this as a conscious
```
rvmiller89: should we remove these comments? the "We" here is confusing since this comment isn't coming from Sumo Logic.

deploy/helm/sumologic/conf/events/events.conf
```diff
@@ -0,0 +1,29 @@
...
+  @type prometheus
+  metrics_path /metrics
+  port 24231
+</source>
+<source>
+  @type prometheus_output_monitor
+</source>
+<source>
+  @type events
+  deploy_namespace $NAMESPACE
```
maimaisie: this is causing issue. I tried changing this line to `deploy_namespace {{ .Release.namespace }}` expecting helm would render it from the `tpl` function [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/118/files#diff-f6b30aed1be608276081c7b71149b30fR11), but it's just empty string if I do that

deploy/helm/sumologic/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
```
yuting-liu: I know that it's not required to use SemVer2 for `appVersion`, but I'm wondering how are we going to increment appVersion later?

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -0,0 +1,56 @@
...
+Expand the name of the chart.
+*/}}
+{{- define "sumologic.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If dryRun=true, we use fixed value "fluentd".
```
maimaisie: when we need to generate yaml files, we would set `dryRun` (for the lack of a better name) to true to use fixed values for release name and the app label
eg.
`helm template ./sumologic --set dryRun=true --namespace sumologic`

deploy/helm/sumologic/Chart.yaml
maimaisie: we can decide whether we want to use it since it's an optional field. if we do, it should be the same as the docker image stable release version

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -0,0 +1,56 @@
...
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified fluentd user conf name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "sumologic.fluentdUserConfig.fullname" -}}
```
frankreno: This comes from the existing Sumo Logic helm chart (fluendUUserConfg).  It was a way for customers to ingest custom configuration into FluentD to capture additional logs.  However, I do not think we need this now as the current Fluent-Bit chart provides a way to do this.

deploy/helm/sumologic/values.yaml
```diff
@@ -0,0 +1,80 @@
+image:
+  repository: sumologic/kubernetes-fluentd
+  tag: 0.0.0
+  pullPolicy: Always
```
frankreno: When we do make this officially released, this should default to IfNotPresent, not always.

rvmiller89:
What does the `helm install` command look like to also pass `prometheus-overrides.yaml` and `fluentbit-overrides.yaml` ?

deploy/helm/sumologic/conf/events/events.conf
maimaisie: oops that fixed it! 😂 thanks Frank

deploy/helm/sumologic/values.yaml
maimaisie: removed

maimaisie:
> What does the helm install command look like to also pass prometheus-overrides.yaml and fluentbit-overrides.yaml ?

@yuting-liu is looking into adding the dependencies into the chart and it will be done in a separate PR. The chart in the current PR will not install prometheus or fluent-bit

deploy/helm/sumologic/templates/_helpers.tpl
maimaisie: this is such that the generated yaml for non-helm users will be consistent with the yaml we currently provide in the repo. for example our current yaml uses `fluentd` as the name for a lot of resources but in helm we would prefix with the release name followed by the chart name

deploy/helm/sumologic/templates/_helpers.tpl
maimaisie: removed

deploy/helm/sumologic/Chart.yaml
maimaisie: changed to `0.0.0` for now

deploy/helm/sumologic/values.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+  ## Format to post logs into Sumo. json, json_merge, or text
+  logFormat: fields
+
+  ## How frequently to push logs to SumoLogic
+  ## ref: https://github.com/SumoLogic/fluentd-kubernetes-sumologic#options
+  flushInterval: 5
+  
+  ## Increase number of http threads to Sumo. May be required in heavy logging clusters
+  numThreads: 1
```
lei-sumo: I think we changed default to 4?

PR #120, opened 2019-08-05
Fix service monitor

samjsong:
We had this running on 15 Fluentd replicas over the last ~3.5 days and none of the replicas have yet to get stuck.

Ryan concurrently ran a test with calling `get_endpoints` in a tight loop over the weekend, and found this time that none of the replicas got stuck. We're not sure what changed, but it seems like `get_endpoints` may be working now? 🤞

### DIFF
--- a/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
+++ b/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
@@ -79,7 +79,6 @@ module Fluent
           metadata['service'] = {'service' => service.sort!.join('_')} if !(service.nil? || service.empty?)
 
           ['pod_labels', 'owners', 'service'].each do |metadata_type|
-          # ['pod_labels', 'owners'].each do |metadata_type|
             attachment = metadata[metadata_type]
             if attachment.nil? || attachment.empty?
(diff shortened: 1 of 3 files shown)


fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -66,23 +66,22 @@ def decorate_record(record)
...
+          log.trace "Record doesn't have [#{@in_pod_path}] field"
         else
           if record.key? 'service'
             record['prometheus_service'] = record['service']
             record.delete('service')
           end
           metadata = get_pod_metadata(namespace_name, pod_name)
-          # service = @pods_to_services[pod_name]
-          # metadata['service'] = {'service' => service.sort!.join('_')} if !(service.nil? || service.empty?)
+          service = @pods_to_services[pod_name]
```
lei-sumo: Is it possible we also use a ThreadSafe cache? Not sure if it helps though.

samjsong: I'm not sure if it helps, it's definitely something to consider - but I think that can be part of a separate PR after we verify that it works

lei-sumo: sg

PR #121, opened 2019-08-05
Bump versions of plugins used in Dockerfile

rvmiller89:
Noticed several of our fluentd plugin gems are out of date, and one is not being used.

Testing performed:
- ci/build.sh
- Redeploy fluentd and fluentd-events pods
- Confirm events logs and normal logs and metrics are still coming in

FYI: @lei-sumo @frankreno

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -16,6 +16,11 @@ RUN apk add --no-cache --update --virtual .build-deps sudo build-base ruby-dev \
        && gem install lru_redux \
        && gem install snappy
 
+# FluentD plugins to allow customers to forward data if needed to various cloud platforms
+RUN gem install fluent-plugin-s3 \
+       && gem install fluent-plugin-google-cloud \
+       && gem install fluent-plugin-azure-storage-append-blob
+
 # FluentD plugins from RubyGems
 RUN gem install fluent-plugin-systemd -v 1.0.2 \
        && gem install fluent-plugin-record-modifier -v 2.0.1 \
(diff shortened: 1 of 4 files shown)


deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -179,7 +179,6 @@ data:
         cache_size "#{ENV['K8S_METADATA_FILTER_BEARER_CACHE_SIZE']}"
         cache_ttl "#{ENV['K8S_METADATA_FILTER_BEARER_CACHE_TTL']}"
         tag_to_kubernetes_name_regexp '.+?\.containers\.(?<pod_name>[^_]+)_(?<namespace>[^_]+)_(?<container_name>.+)-(?<docker_id>[a-z0-9]{64})\.log$'
-        merge_json_log false
```
lei-sumo: Why do need to remove this one?

rvmiller89: This parameter was removed in the newer version of `kubernets_metadata` plugin, from their README:

> NOTE: As of the release 2.1.x of this plugin, it no longer supports parsing the source message into JSON and attaching it to the payload. The following configuration options are removed:

> merge_json_log
> preserve_json_log

deploy/docker/Dockerfile
```diff
@@ -17,14 +17,13 @@ RUN apk add --no-cache --update --virtual .build-deps sudo build-base ruby-dev \
        && gem install snappy
 
 # FluentD plugins from RubyGems
-RUN gem install fluent-plugin-s3 -v 1.1.4 \
```
maimaisie: is this plugin not used anywhere so we are removing it?

rvmiller89: It's not being used currently, but after talking with Frank we want to keep it around in case a customer wants to fork their data to S3 (and in the future GCP, Azure, etc.).

PR #124, opened 2019-08-06
Optionally deploy and download yaml in setup.sh

maimaisie:
This PR adds two new flags to the setup script: 
- `-d` for deploying (running `kubectl apply`); default to true and 
- `-y` for downloading yaml; default to true

Two reasons for these changes:
1. When working with @drduke1 on the docs we realize for users who wants to customize the configuration, they currently run setup.sh which deploys everything with default settings, then they'll make changes to the downloaded yaml, and then redeploy. With this change, they can set `-d false` and it will only download the yaml file so they can make changes locally and run `kubectl apply` on their own.
2. Before we add pre-installation hook to our Helm chart that sets up the collector and sources as part of `helm install`, Helm users will need to run the setup script with `-d false -y false` before they install the chart.

To summarize, these are the conditions and the behaviors:
- default: deploy and download yaml
- `-d false`: only download
- `-d false -y false`: no deploy and no download (only set up the secrets for collector and sources)

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -95,7 +95,8 @@ __NOTE__ This script will be executed in bash and requires [jq command-line JSON
 * __-k cluster-name__ - optional. Name of the Kubernetes cluster that will be attached to logs and events as metadata. If not specified, it will be named as `kubernetes-<timestamp>`. For metrics, specify the cluster name in the `prometheus-overrides.yaml` provided for the prometheus operator; further details in [step 2](#step-2-configure-prometheus).
 * __-n namespace__ - optional. Name of the Kubernetes namespace in which to deploy resources. If not specified, the namespace__ will default to `sumologic`
 * __-a use-alpha__ - optional. Use value true if you want to deploy latest alpha version. If not specified, the latest release will be deployed.
-* __-y download-yaml__ - optional. Use value false if you only want to set up the Sumo collector and sources but not download the yaml. If not specified, the yaml file with default configuration will be downloaded.
+* __-d deploy__ - optional. Use value false if you only want to set up the Sumo collector and sources and download the yaml, but do not want to deploy yet because you want to customize the yaml. If not specified, resources with default configuration will be deployed.
+* __-y download-yaml__ - optional. Use value false if you set `deploy` to false and do not want to download the yaml. If not specified, the yaml file with default configuration will be downloaded.
 * __api-endpoint__ - required. The API endpoint from [this page](https://help.sumologic.com/APIs/General-API-Information/Sumo-Logic-Endpoints-and-Firewall-Security).
 * __access-id__ - required. Sumo [access id](https://help.sumologic.com/Manage/Security/Access-Keys)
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/setup.sh
```diff
@@ -4,7 +4,7 @@ set -e
 usage() {
   echo
   echo 'Usage:'
-  echo '  setup.sh [-c collector-name] [-k cluster-name] [-n namespace] [-a useAlpha] <endpoint> <access-id> <access-key>'
+  echo '  setup.sh [-c collector-name] [-k cluster-name] [-n namespace] [-a use-alpha] [-d deploy] [-y download-yaml] <endpoint> <access-id> <access-key>'
```
rvmiller89: technically changing parameter name is a breaking change in case anyone has integrated this script into their automation, but since it's so new it's probably safe to change to `use-alpha` now.

maimaisie: it is actually not the parameter name, just a placeholder for the parameter value. the name is just `-a`

rvmiller89: oh good point.

deploy/README.md
```diff
@@ -84,20 +84,22 @@ This approach requires access to the Sumo Logic Collector API. It will create a
...
-* __access-key__ - required. Sumo [access key](https://help.sumologic.com/Manage/Security/Access-Keys)
+* __-c &lt;collector_name&gt;__ - optional. Name of Sumo collector that will be created. If not specified, it will be named as `kubernetes-<timestamp>`
+* __-k &lt;cluster_name&gt;__ - optional. Name of the Kubernetes cluster that will be attached to logs and events as metadata. If not specified, it will be named as `kubernetes-<timestamp>`. For metrics, specify the cluster name in the `prometheus-overrides.yaml` provided for the prometheus operator; further details in [step 2](#step-2-configure-prometheus).
+* __-n &lt;namespace&gt;__ - optional. Name of the Kubernetes namespace in which to deploy resources. If not specified, the namespace will default to `sumologic`.
+* __-a &lt;boolean&gt;__ - optional. Set this to true if you want to deploy with the latest alpha version. If not specified, the latest release will be deployed.
+* __-d &lt;boolean&gt;__ - optional. Set this to false to only set up the Sumo Collector and Sources and download the YAML file, but not to deploy so you can customize the YAML file. If not specified, the default configuration will deploy.
```
lei-sumo: Can we make `-a` and `-d` options instead of parameters? It doesn't make sense that user have to pass true/false there. Also the default values are not consistent with each other ..

maimaisie: From the email this morning the customer mentioned
> option parsing needs to either be all flags or all positional – I added a couple extra parameters and the option to pass in a local file since we had to override the `image:` fields (see code snippet below)

Does this mean he doesn't want a mix of parameters and flags? When I added `-d` and `-y` I wanted to be consistent with the existing ones. I'm not against changing them (and the default values)

PR #125, opened 2019-08-06
Switch to Debian fluentd Dockerfile

rvmiller89:
The [fluentd-kubernetes](https://github.com/fluent/fluentd-kubernetes-daemonset) daemonset has deprecated using Alpine as their base image, and also says

> The following repository expose images based on Alpine Linux and Debian. For production environments we strongly suggest to use Debian images.

This PR changes our fluentd Docker image to be based on Debian.

Testing performed:
- ci/build.sh
- Redeploy fluentd and fluentd-events pods
- Confirm events logs and normal logs and metrics are still coming in

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -3,8 +3,7 @@ FROM fluent/fluentd:v1.6.3-debian-1.0
 # Use root account to use apt
 USER root
 
-# below RUN includes plugin as examples elasticsearch is not required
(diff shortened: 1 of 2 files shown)


rvmiller89:
@frankreno the new image is `0.0.23-alpha` if you want to run that through the vuln scan.  Unfortunately the size is almost double using the Debian base image @ 158 MB.  Not sure how big of a deal that is for customers, but something to note.

PR #127, opened 2019-08-07
Note On Possible Remote Write Update Changes

frankreno:
For now, just updating docs, but we should think through a better solution. Update README with note on remote write API changes needed when deploying Prometheus to a different namespace or using an existing Prometheus running in a different namespace.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -204,6 +204,8 @@ Before installing `prometheus-operator`, edit `prometheus-overrides.yaml` to def
 
 __NOTE__ It’s fine to change the value of the `cluster` field, but don’t change the field name (key).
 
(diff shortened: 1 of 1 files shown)


rvmiller89:
wasn't sure if you wanted us to take a look at this, but it LGTM if you want to merge

deploy/README.md
```diff
@@ -204,6 +204,8 @@ Before installing `prometheus-operator`, edit `prometheus-overrides.yaml` to def
 
 __NOTE__ It’s fine to change the value of the `cluster` field, but don’t change the field name (key).
 
+__NOTE__ If you plan to install Prometheus in a a different namespace than you deployed FluentD to in Step 1, or you have an existing Prometheus you plan to apply our configuration to running in a different namespace,  please update the remote write API configuration to use the full service url. e.g. `http://fluentd.sumologic.svc.cluster.local:9888`.
```
lei-sumo: typo: double "a"

frankreno: Fixed :)

PR #130, opened 2019-08-07
Create pull_request_template.md

rvmiller89:


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+###### Description
+
+Fill in your description here.
```
frankreno: Should we have a section to ask for some key things? Like:

How was this deployed? Helm or YAML
What version of Kubernetes are you running?
Is this a managed service, if so which one?
Please share any customizations you have made
What version of the Prometheus Operator and Fluent-Bit charts (helm ls)

rvmiller89: that sounds more like an issue/bug report.  This is the PR template when adding new functionality or fixing issues.

frankreno: facepalm, you are right apologies, saw PR template yet my mind went to issue template (which would be a good add to save some back and forth)

PR #137, opened 2019-08-09
Generate yaml from helm chart on commits

maimaisie:
###### Description

On every commit, CI will auto-generate `fluentd-sumologic.yaml.tmpl` if it detects any diffs between the output of `helm template` and the current file.

Will need to test the generated yaml before merging.

###### Testing performed

- [x] ci/build.sh


deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -1,59 +1,27 @@
+# This file is auto-generated. Please DO NOT modify.
```
rvmiller89: I'm nervous customers will get the wrong idea reading `Please DO NOT modify.`, since yaml customers should be allowed to modify after downloading.

Could we remove `Please DO NOT modify`.  We could actually have the CI fail the Travis build if it detects changes to this file?

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -304,18 +405,19 @@ spec:
           type: ""
       - name: config-volume
         configMap:
-          name: fluentd-config
+          name: fluentd
       containers:
       - name: fluentd
-        image: sumologic/kubernetes-fluentd:0.1.0
+        image: "sumologic/kubernetes-fluentd:0.0.0"
```
rvmiller89: should be `0.1.0` .. Also could we avoid the quotes somehow? I don't normally see yaml files with these image values quoted.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
maimaisie: Sure that seems better. I forgot customers should still be able to change this file

maimaisie: We probably can't have CI to block change to this file because CI will run on the automated commit that generates this file...

rvmiller89: You can use this one-liner to check the commit's author and grep -i for "travis":
```sh
if git --no-pager show -s --format="%an" . | grep -q -i "travis"; then
  # commit is from travis
fi
```

rvmiller89: example:
```sh
$ git log -n 1
commit 479b63ab5f612679d4ec9213482c970c09d1f442 (HEAD -> maisie-yaml-automation, origin/maisie-yaml-automation)
Author: Travis CI User <travis@example.org>
Date:   Fri Aug 9 22:32:19 2019 +0000

    Generate new 'fluentd-sumologic.yaml.tmpl'

$ git --no-pager show -s --format="%an" .
Travis CI User
```

maimaisie: ah thanks Ryan

maimaisie: Added check in build.sh to fail the build if the committed files contain `fluentd-sumologic.yaml.tmpl` _and_ the author of the last commit is not Travis (since auto-generated yaml should always be the last commit)

rvmiller89:
Master build is failing, we should either revert or fix this:  https://travis-ci.org/SumoLogic/sumologic-kubernetes-collection/builds/571492109

```
$ bash ci/build.sh
Detected manual changes in 'fluentd-sumologic.yaml.tmpl', abort.
The command "bash ci/build.sh" exited with 1.
```

PR #140, opened 2019-08-12
Move fluent-bit INPUT config to overrides yaml

samjsong:
###### Description

Since the OOB fluent-bit helm chart does not expose configs for `Refresh_Interval` and `Rotate_Wait`, we decided to expose the `[INPUT]` config for fluent-bit explicitly in the overrides yaml file.

For reference, this was the helm chart's `[INPUT]` [config](https://github.com/helm/charts/blob/master/stable/fluent-bit/templates/config.yaml). We need to keep `.Values.input.systemd.enabled` and `.Values. trackOffsets` in our overrides since they are used [here](https://github.com/helm/charts/blob/master/stable/fluent-bit/templates/daemonset.yaml).

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -22,70 +22,75 @@ tolerations:
     operator: Exists
 
 input:
-  tail:
-    memBufLimit: 5MB
-    parser: docker
-    path: /var/log/containers/*.log
   systemd:
     enabled: true
-    filters:
-      systemdUnit:
-        - addon-config.service
-        - addon-run.service
-        - cfn-etcd-environment.service
-        - cfn-signal.service
-        - clean-ca-certificates.service
-        - containerd.service
-        - coreos-metadata.service
-        - coreos-setup-environment.service
-        - coreos-tmpfiles.service
-        - dbus.service
-        - docker.service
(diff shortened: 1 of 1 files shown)


deploy/helm/fluent-bit-overrides.yaml
```diff
@@ -22,70 +22,75 @@ tolerations:
...
-
-filter:
-  kubeTag: containers
 
 rawConfig: |-
   @INCLUDE fluent-bit-service.conf
-  @INCLUDE fluent-bit-input.conf
-  @INCLUDE fluent-bit-output.conf
+
+  [INPUT]
```
lei-sumo: Will `extraEntries.input` help to expose the knobs we want?

samjsong: Unfortunately not. There's two issues with `extraEntries.input`:
1. Since `Refresh_Interval 5` is already hard coded in line 33, we cannot configure this parameter.
2. Since `extraEntries.input` is on line 50, after the systemd section, any configuration will be inserted to the systemd section, not the tail section. So if we want to specify `Rotate_Wait` with `extraEntries.input`, we cannot use the systemd section (`input.systemd.enabled` must be set to false).

PR #141, opened 2019-08-12
Update doc with steps to extend fluentD image with custom plugins.

frankreno:
Update doc with steps to extend fluentD image with custom plugins.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -484,9 +484,11 @@ kubectl delete namespace sumologic
 
 ## Adding Additional FluentD Plugins
 To add additional FluentD plugins, you can simply create a new Docker image from our provided Docker image.
+ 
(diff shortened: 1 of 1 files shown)


deploy/README.md
```diff
@@ -481,6 +482,18 @@ To delete the `sumologic` namespace and all resources under it:
 kubectl delete namespace sumologic
 ```
 
+## Adding Additional FluentD Plugins
+To add additional FluentD plugins, you can simply create a new Docker image from our provided Docker image.
+
+```
+FROM sumologic/kubernetes-fluentd:0.0.0
```
samjsong: LGTM overall but not sure how I feel about hard coding the 0.0.0 here, I guess this is just an example and the customer should hopefully understand to use the latest stable version?

frankreno: thats a fair point, let me add more details on that...

frankreno: updated

PR #142, opened 2019-08-13
Update labels in prometheus-overrides.yaml

maimaisie:
###### Description

After we use helm to generate `fluentd-sumologic.yaml.tmpl`, we now use different labels than our previous ones. 
`k8s-app: fluentd-sumologic` => `app: sumologic`
`k8s-app: fluentd-sumologic-events` => `app: sumologic-events`

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -6,7 +6,7 @@ prometheus:
   additionalServiceMonitors:
     - name: fluentd
       additionalLabels: 
-        k8s-app: fluentd-sumologic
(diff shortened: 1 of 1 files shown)


maimaisie:
@yuting-liu we need to be careful about these labels and also the service names in the prometheus override values in the helm chart, since the labels are prefixed with the chart's release name which can be dynamic, so we cannot just hardcode the labels and service names. Let's talk offline when you have time

maimaisie:
This PR is not safe to merge until we update README to use versioned prometheus-override.yaml

PR #159, opened 2019-08-30
fix labels, disable Fluent-Bit Helm Chart serviceMonitor, use Prometheus Operator additionalServiceMonitors to work around expectation that Prometheus Operator is already installed.

frankreno:
fix labels, disable Fluent-Bit Helm Chart serviceMonitor, use Prometheus Operator additionalServiceMonitors to work around expectation that Prometheus Operator is already installed.

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -6,29 +6,43 @@ grafana:
 prometheus:
   additionalServiceMonitors:
     - name: fluentd
-      additionalLabels: 
+      additionalLabels:
         app: sumologic
-        release: prometheus-operator
+        release: collection
       endpoints:
       - port: metrics
       namespaceSelector:
         matchNames:
         - sumologic
       selector:
-          matchLabels:
-            app: sumologic
+        matchLabels:
+          app: fluentd
     - name: fluentd-events
-      additionalLabels: 
+      additionalLabels:
         app: sumologic-events
-        release: prometheus-operator
+        release: collection
       endpoints:
-      - port: metrics
+        - port: metrics
       namespaceSelector:
         matchNames:
-        - sumologic
+          - sumologic
       selector:
-          matchLabels:
-            app: sumologic-events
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -221,29 +217,43 @@ prometheus-operator:
   prometheus:
     additionalServiceMonitors:
       - name: collection-sumologic
-        additionalLabels: 
+        additionalLabels:
           app: sumologic
-          release: prometheus-operator
+          release: collection
```
lei-sumo: We will also need to document to remind user to override all these hard corded value if they specify a different release name cc @maimaisie

rvmiller89:
@frankreno I get this error trying this, maybe I'm doing something wrong?
```
k8s $ k get pods
No resources found.
k8s $ k get servicemonitors
No resources found.
k8s $ helm install sumologic/sumologic --name collection --namespace sumologic -f values.yaml --no-crd-hook
Error: release collection failed: servicemonitors.monitoring.coreos.com "collection-fluent-bit" already exists
```

rvmiller89:
Frank and I figured out the issue I was having with `"collection-fluent-bit" already exists` ...

The problem is that I was applying the new `values.yaml` over the repository helm chart, which still had the old fluent-bit service monitor declaration, so it was trying to declare the same service monitor twice.  This will be resolved once we update the repository version.

deploy/helm/prometheus-overrides.yaml
```diff
@@ -6,29 +6,43 @@ grafana:
...
-      - port: metrics
+        - port: metrics
       namespaceSelector:
         matchNames:
-        - sumologic
+          - sumologic
+      selector:
+        matchLabels:
+          app: fluentd-events
+    - name: collection-fluent-bit
```
maimaisie: @frankreno this file is generated from `values.yaml`. is this `name: collection-fluent-bit` correct for the `prometheus-overrides.yaml` file?

PR #166, opened 2019-08-30
Git ignore Helm chart deps

rvmiller89:
###### Description

Same as `gh-pages` branch but for `master` branch.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ Gemfile.lock
 
 # VS Code ignores
 .vscode/*
+
+# Helm chart dependencies
+deploy/helm/sumologic/charts
+deploy/helm/sumologic/charts/*


.gitignore
```diff
@@ -16,3 +16,7 @@ Gemfile.lock
 
 # VS Code ignores
 .vscode/*
+
+# Helm chart dependencies
+deploy/helm/sumologic/charts
```
lei-sumo: I don't even see these in our code base. Why do we need to exclude explicitly?

rvmiller89: they will be generated if you are developing (making changes to) our helm chart to test locally. We want to be sure to not check those in.

lei-sumo: Got it. Thanks.

PR #167, opened 2019-08-30
Migrate kubernetes_sumologic filter plugin

samjsong:
###### Description

This moves the `kubernetes_sumologic` filter plugin from the old repo to this repo, and builds locally rather than pulling the gem.

We also duplicate the `host` metadata and attach as a new field `node` since that makes more sense in the kubernetes context.

image

###### Testing performed

- [x] ci/build.sh


samjsong:
It's copy & paste, except a few minor changes. To remain consistent with the other plugins in our repo, I changed `fluent-plugin-kubernetes_sumologic` to `fluent-plugin-kubernetes-sumologic`, plus a few other very small changes for consistency (none that affect behaviour of the plugin).

Good catch on the null field value change. That's currently still in the `enhance_k8s_metadata` plugin, I'll move that over.

PR #168, opened 2019-08-30
SUMO-117261: Skip using TRAVIS_COMMIT_RANGE for yaml detection

rvmiller89:
###### Description

Recommend using "Hide whitespace changes" since most of this is just indent change.

Rather than using  `TRAVIS_COMMIT_RANGE` which is empty on initial commit, we can `git diff master...$TRAVIS_BRANCH` to detect manual changes to the yaml files.

Rather than checking for changes made to `values.yaml`, we can always generate the new files and commit them if there is a difference.

###### Testing performed

- [x] ci/build.sh
- [x] Tested initial commit with no yaml changes
- [x] Tested initial commit with changes in values.yaml for our template as well as overrides
- [x] Tested initial commit incorrectly making changes to overrides and ensure it failed

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 if [ "$TRAVIS_PULL_REQUEST" == false ] && [ "$TRAVIS_BRANCH" != "master" ] && [ "$TRAVIS_EVENT_TYPE" == "push" ]; then
-  changes=`git diff master...$TRAVIS_BRANCH | grep -i "fluentd-sumologic.yaml.tmpl\|fluent-bit-overrides.yaml\|prometheus-overrides.yaml\|falco-overrides.yaml"`
+  changes=`git diff master...$TRAVIS_BRANCH --name-only | grep -i "fluentd-sumologic.yaml.tmpl\|fluent-bit-overrides.yaml\|prometheus-overrides.yaml\|falco-overrides.yaml"`
   if [ -n "$changes" ]; then
     if git --no-pager show -s --format="%an" . | grep -v -q -i "travis"; then
       echo "Aborting due to manual changes detected in the following generated files: $changes"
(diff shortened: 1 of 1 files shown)


ci/build.sh
```diff
@@ -56,8 +51,23 @@ if [ -n "$GITHUB_TOKEN" ]; then
   git config --global user.email "travis@travis-ci.org"
   git config --global user.name "Travis CI"
   git remote add origin-repo https://${GITHUB_TOKEN}@github.com/SumoLogic/sumologic-kubernetes-collection.git > /dev/null 2>&1
+  git fetch origin-repo
+  git checkout $TRAVIS_BRANCH
+fi
+
+# Check for invalid changes to overrides yaml files
```
maimaisie: should this check be at the top like it was before? The reason we put it at the top is because we didn't want to build the gems or the image if it would fail here

rvmiller89: If we're okay with moving the `# Set up Github` part to the top too, I'm fine with having it fail faster.  In my testing it wasn't a huge deal that it built the gems/docker image before failing.

ci/build.sh
```diff
@@ -71,65 +81,57 @@ if [ "$TRAVIS_BRANCH" != "master" ] && [ "$TRAVIS_EVENT_TYPE" == "push" ] && [ -
...
-      git fetch origin-repo
-      git checkout $TRAVIS_BRANCH
-      git add deploy/helm/*-overrides.yaml
-      git commit -m "Generate new overrides yaml file(s)."
-      git push --quiet origin-repo "$TRAVIS_BRANCH"
-    else
-      echo "No changes in the generated overrides files."
-    fi
+  # Generate override yaml files for chart dependencies to determine if changes are made to overrides yaml files
+  echo "Generating overrides files..."
```
lei-sumo: Do we always generate override files now?

rvmiller89: We do.  We will only commit them if there is a difference detected, however.

lei-sumo: gtk

PR #173, opened 2019-08-30
Update helm readme

maimaisie:
###### Description

After #159 user also needs to replace all occurrences of `collection` for the override values if they use a different release name.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -34,7 +34,7 @@ Install the chart with release name `collection` and namespace `sumologic`
 helm install sumologic/sumologic --name collection --namespace sumologic
 ```
 
-NOTE: If you install the chart with a different release name or a different namespace, you will need to override remote write URLs for Prometheus and the host for fluent-bit. We recommend using an override file due to the number of URLs that need to be overridden. You can use the following command to download the override file and replace the `<RELEASE-NAME>` and `<NAMESPACE>`:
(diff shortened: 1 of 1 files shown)


deploy/helm/README.md
```diff
@@ -34,17 +34,19 @@ Install the chart with release name `collection` and namespace `sumologic`
...
 ```
 
-NOTE: If you install the chart with a different release name or a different namespace, you will need to override remote write URLs for Prometheus and the host for fluent-bit. We recommend using an override file due to the number of URLs that need to be overridden. You can use the following command to download the override file and replace the `<RELEASE-NAME>` and `<NAMESPACE>`:
+NOTE: If you install the chart with a different release name or a different namespace, you will need to override some configuration fields for both Prometheus and fluent-bit. We recommend using an override file due to the number of fields that need to be overridden. You can use the following command to download the override file and replace the `<RELEASE-NAME>` and `<NAMESPACE>`:
   
 ```bash
 curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/helm/sumologic/values.yaml | \
-sed 's/\collection-sumologic.sumologic'"/<RELEASE-NAME>-sumologic.<NAMESPACE>/g" > values.yaml
+sed 's/\-sumologic.sumologic'"/-sumologic.<NAMESPACE>/g" | \
+sed 's/\collection'"/<RELEASE-NAME>/g" > values.yaml
```
lei-sumo: would this regex for collection be too broad?

maimaisie: currently it's fine, we'll need to re-evaluate it in the future maybe

PR #176, opened 2019-09-03
Pre-install hook for helm

maimaisie:
###### Description

Adds a helm pre-install hook that runs the setup script as a job. Currently the container is built from a separate image than our fluentd one. I will look into ways to use the same 

###### Testing performed

- [x] ci/build.sh
- [x] Package new helm chart locally and deploy to make sure the hook is being run


deploy/helm/sumologic/templates/setup-job.yaml
```diff
@@ -0,0 +1,32 @@
...
+  labels:
+    app: {{ template "sumologic.labels.app" . }}
+    {{ template "sumologic.labels.common" . }}
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: setup
+        image: setup-test:local
```
maimaisie: need to change this

deploy/helm/sumologic/values.yaml
```diff
@@ -253,7 +273,7 @@ prometheus-operator:
     prometheusSpec:
       externalLabels:
         # Set this to a value to distinguish between different k8s clusters
-        cluster: kubernetes
+        cluster: *cluster_name
```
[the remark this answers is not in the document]
maimaisie: this is called an anchor used to reference a value in YAMLs. so `*cluster_name` will look for `&cluster_name` and take whatever value `&cluster_name` has

deploy/helm/sumologic/values.yaml
```diff
@@ -26,6 +26,26 @@ eventsDeployment:
       cpu: "100m"
 
 sumologic:
+  ## Setup
+
+  # If enabled, a pre-install hook will create Collector and Sources in Sumo Logic
+  setupEnabled: true
```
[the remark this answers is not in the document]
maimaisie: We are commenting out access id, key and endpoint because they are required values with no defaults. When we reference them [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/176/files#diff-6c6e2426d7625fb2a69ef04d6b81d272R29) , we use the `required` function so that if user does not override this value on `helm install` it will error out. For example:
```
$ helm install maisie-charts/sumologic --name collection --namespace sumologic --no-crd-hook
Error: render error in "sumologic/templates/setup-job.yaml": template: sumologic/templates/setup-job.yaml:28:14: executing "sumologic/templates/setup-job.yaml" at <required "A valid .Values.sumologic.endpoint entry required!" .Values.sumologic.endpoint>: error calling required: A valid .Values.sumologic.endpoint entry required!
```

The collector name is commented out because the script has the default value, so we don't need to specify it again here

deploy/docker/setup/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apk add --no-cache gnupg
+RUN apk add --no-cache curl
+
+# Install kubectl
+ADD https://storage.googleapis.com/kubernetes-release/release/v1.6.12/bin/linux/amd64/kubectl /usr/local/bin/kubectl
+RUN chmod +x /usr/local/bin/kubectl
+
+# Install jq
+ENV JQ_VERSION='1.5'
+RUN wget --no-check-certificate https://raw.githubusercontent.com/stedolan/jq/master/sig/jq-release.key -O /tmp/jq-release.key && \
```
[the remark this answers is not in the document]
maimaisie: hah let me try that

deploy/docker/setup/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM alpine:3.9
+
+RUN apk add --no-cache gnupg
+RUN apk add --no-cache curl
+
+# Install kubectl
+ADD https://storage.googleapis.com/kubernetes-release/release/v1.6.12/bin/linux/amd64/kubectl /usr/local/bin/kubectl
+RUN chmod +x /usr/local/bin/kubectl
```
lei-sumo: Where do we use the kubectl?

deploy/helm/sumologic/values.yaml
```diff
@@ -26,6 +26,26 @@ eventsDeployment:
...
+  #accesskey: ""
+
+  # Sumo API endpoint
+  #endpoint: ""
+
+  # Collector name
+  #collectorName: ""
+
+  # Cluster name
+  clusterName: &cluster_name "kubernetes"
```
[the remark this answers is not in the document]
lei-sumo: does it work in the user override case? --set sumologic.clusterName

deploy/helm/sumologic/values.yaml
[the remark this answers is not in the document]
maimaisie: found out this won't work if it's overriden in `helm install`, it still gets the default values `kubernetes` instead of the user-supplied value... I have remove the anchor from values.yaml. If user needs to use a different cluster name, they need to set two things 
```
--set sumologic.clusterName="test-cluster" 
--set prometheus-operator.prometheus.prometheusSpec.externalLabels.cluster="test-cluster"
```
this is not ideal, but @frankreno do you think it's acceptable for GA?

deploy/docker/setup/Dockerfile
maimaisie: in the setup script

maimaisie:
I think the CI has a bug, it's thinking I have manual changes in some generated files
```
Aborting due to manual changes detected in the following generated files: deploy/helm/prometheus-overrides.yaml
deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```

deploy/docker/setup/setup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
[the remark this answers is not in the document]
maimaisie: changed back

deploy/helm/sumologic/templates/setup-job.yaml
```diff
@@ -0,0 +1,32 @@
...
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        command: ["/usr/bin/setup.sh", 
+          "-n", "{{ .Release.Namespace }}", 
+          "-k", "{{ .Values.sumologic.clusterName }}", 
+          {{- if .Values.sumologic.collectorName }}
+          "-c", "{{ .Values.sumologic.collectorName }}",
+          {{- end }}
+          "-d", "false", 
+          "-y", "false", 
+          '{{ required "A valid .Values.sumologic.endpoint entry required!" .Values.sumologic.endpoint }}',
```
frankreno: What are your thoughts on instead of using the full URL, we let the customer specify the deployment name (e.g. US2)? We then do a conditional block rendering the right URL based on what they selected and can fail if they make an invalid choice. This makes it easier for people, other various open-source Sumo projects work with this way.  We have to occasionally update that as we add new deployments, but that seems minor.

maimaisie: It will be more maintenance work and I think we generally try to avoid it because it's easily forgotten. What are your thoughts on this @lei-sumo @samjsong @rvmiller89 ?

lei-sumo: I'm fine with documenting it in our steps for now.

deploy/docker/Dockerfile
```diff
@@ -49,7 +49,16 @@ USER root
...
  && apt-get update \
- && apt-get install -y --no-install-recommends $runDeps
+ && apt-get install -y --no-install-recommends $runDeps \
+ && apt-get install -y gnupg \
+ && apt-get install -y curl \
+ && apt-get install -y jq \
+ && apt-get install -y wget
+
+ # Install kubectl
+ADD https://storage.googleapis.com/kubernetes-release/release/v1.6.12/bin/linux/amd64/kubectl /usr/local/bin/kubectl
```
lei-sumo: Is this version too low?

maimaisie: updated to 1.15.0 according to their [doc page](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl-on-linux)

PR #178, opened 2019-09-05
Pin patch version of helm dependencies

samjsong:
###### Description

We want to pin the stable versions that have been tested, including the patch version.

###### Testing performed

- [ ] ci/build.sh


samjsong:
@frankreno what was the bug that you mentioned, where we'd want to pin `5.10.5` of the prometheus operator chart?

frankreno:
the issue was with the admissionWebook and `helm template`.  We have hit other issues with webhook admission controller.

frankreno:
We can also go the route of documenting the issue

PR #181, opened 2019-09-08
RBAC for pre-install hook

maimaisie:
###### Description

Fix RBAC error for pre-install hook

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-serviceaccount.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-serviceaccount.yaml
@@ -1,0 +1,12 @@
+{{- if .Values.sumologic.setupEnabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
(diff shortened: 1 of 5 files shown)


deploy/helm/sumologic/templates/setup/setup-serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.sumologic.setupEnabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name:  {{ template "sumologic.fullname" . }}-setup
```
lei-sumo: Do we want to name all these new resources with the same name?

maimaisie: yes, it's a very conventional to give resources the same name in helm as long as we don't have two or more resources of the same type with the same name

PR #185, opened 2019-09-10
Assign weight to pre-install hooks

maimaisie:
###### Description

According to docs:

> Tiller sorts hooks by weight (assigning a weight of 0 by default) and by name for those hooks with the same weight in ascending order. Tiller then loads the hook with the lowest weight first (negative to positive)

So the service account should be assigned the lowest weight to get loaded first, and the job should be assigned the highest to get loaded last.

###### Testing performed

- [ ] ci/build.sh


maimaisie:
I will push a new 0.3.0-alpha package to the helm repo to override the current one. After that let's hopefully the jenkins job will consistently pass @Gourav2906

PR #198, opened 2019-09-18
docs(README.md, deploy/README.md): Updated the README.mds to have cle…

abhi-sumo:
…aner step descriptions for exis

###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ This repo contains all required resources to collect data from Kubernetes cluste
 
 ![overview](/images/overview.png)
 
+# Installation
+
+Detailed instructions are available in our [Installation Guide](./deploy/README.md).
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -8,6 +8,10 @@ This repo contains all required resources to collect data from Kubernetes cluste
 
 ![overview](/images/overview.png)
 
+# Installation
```
samjsong: Let's delete line 5 since we're moving it down here instead

samjsong:
One follow up AI was that we wanted to include some sort of verification step, like `at this point you should be done, run ___ to verify that everything is set up correctly`, but this would have to go in several places depending on how the customer is installing

PR #205, opened 2019-09-20
Add flag to disable events collection in helm

maimaisie:
###### Description

Add `eventCollectionEnabled` default to true.

###### Testing performed

- [ ] ci/build.sh


deploy/docker/setup/setup.sh
```diff
@@ -190,21 +192,34 @@ ENDPOINT_METRICS_NODE_EXPORTER="$SOURCE_URL"
...
+  kubectl -n $NAMESPACE create secret generic sumologic \
+    --from-literal=endpoint-metrics=$ENDPOINT_METRICS \
+    --from-literal=endpoint-metrics-apiserver=$ENDPOINT_METRICS_APISERVER \
+    --from-literal=endpoint-metrics-kube-controller-manager=$ENDPOINT_METRICS_KUBE_CONTROLLER_MANAGER \
+    --from-literal=endpoint-metrics-kube-scheduler=$ENDPOINT_METRICS_KUBE_SCHEDULER \
+    --from-literal=endpoint-metrics-kube-state=$ENDPOINT_METRICS_KUBE_STATE \
+    --from-literal=endpoint-metrics-kubelet=$ENDPOINT_METRICS_KUBELET \
+    --from-literal=endpoint-metrics-node-exporter=$ENDPOINT_METRICS_NODE_EXPORTER \
+    --from-literal=endpoint-logs=$ENDPOINT_LOGS \
+    --from-literal=endpoint-events=$ENDPOINT_EVENTS
```
rvmiller89: can we DRY this up and only add/omit `--from-literal=endpoint-events=$ENDPOINT_EVENTS` without repeating the entire command?

samjsong:
Found two concerns while playing around with this:
1. When running the setup script outside of helm, when `deploy` is set to true, we apply `fluentd-sumologic.yaml`, but this yaml file deploys the events resources. This is only a problem for non-helm users.
I'm not sure how to get around this without either having to do some sort of hacky thing in the setup script itself to remove those things from the file (really don't think this is the way to go), or maintain a second fluentd-sumologic.yaml file that doesn't deploy events collection (it would also autogenerated, but still not ideal). Any ideas?

2. I noticed for this `COLLECT_EVENTS` `-e` flag, we have a default `false`, so when I tried
```
deploy/docker/setup/setup.sh -c ssong-k8s-7 -k ssong-k8s-7 -e false <ENDPOINT> <ID> <KEY>
```
it didn't work; you actually have to do 
```
deploy/docker/setup/setup.sh -c COLLECTOR -k CLUSTER -e <ENDPOINT> <ID> <KEY>
```
which is *better*, but inconsistent with our other flags `-d` and `-y` which currently require users to pass in a value. We could change the way `-d` and `-y` work, but that'd be backwards incompatible for non-helm users. What do we think?

rvmiller89:
Could we leave the setup script alone (always create events endpoint, even if nothing will be sent to it)?  Only Helm users should expect changing a value in values.yaml should have an effect -- non-Helm users won't be using values.yaml at all.

samjsong:
The setup script creates the sources on Sumo side though, and we use the setup script in the helm deploy - so if we don't touch the setup script, then helm users won't get any events resources deployed on their clusters but still have the events HTTP source created in Sumo

rvmiller89:
Right -- is that so horrible? It's just a dormant source they would only see in the Manage Collection page.

samjsong:
I see your point - I guess this also allows users to seamlessly reenable events collection if they so desire in the future, without recreating the `sumologic` secret. +1

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -26,6 +26,9 @@ spec:
           {{- if .Values.sumologic.collectorName }}
           "-c", "{{ .Values.sumologic.collectorName }}",
           {{- end }}
+          {{- if eq .Values.sumologic.eventCollectionEnabled false }}
+          "-e",
```
rvmiller89: this should be removed

samjsong: ah good catch, thanks Ryan

PR #206, opened 2019-09-23
feat(deployment.yaml): Updated to expose the EXCLUDE NAMESPACE and ot…

abhi-sumo:
…her Regex

###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -115,11 +115,6 @@ sumologic:
   ## Matching namespaces will be excluded from Sumo. The logs will still be sent to FluentD.
   excludeNamespaceRegex: ""
 
-  ## Files matching this pattern will be ignored by the in_tail plugin, 
-  ## and will not be sent to Kubernetes or Sumo. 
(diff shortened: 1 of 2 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -103,6 +103,28 @@ sumologic:
...
+  excludeHostRegex: ""
+
+  ## A regular expression for namespaces. 
+  ## Matching namespaces will be excluded from Sumo. The logs will still be sent to FluentD.
+  excludeNamespaceRegex: ""
+
+  ## Files matching this pattern will be ignored by the in_tail plugin, 
+  ## and will not be sent to Kubernetes or Sumo. 
+  ## This can be a comma-separated list as well. See in_tail documentation for more information. 
+  excludePath: ""
```
samjsong: We no longer use the `EXCLUDE_PATH` regex environment variable. Instead, users can update the `values.yaml` file directly under the `fluent-bit` section [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L159) to change which files they wish to collect. Let's get rid of this reference

PR #216, opened 2019-09-25
Fix CI Yaml validation for subsequent commits

rvmiller89:
###### Description

For non-Travis code authors, verify they have not changed any of the generated files.

This fixes the bug where you couldn't commit subsequent changes after Travis had generated some files in your PR.

FYI: @maimaisie @frankreno 

###### Testing performed

- [x] ci/build.sh
- [x] Verify failed attempt at making change to generated file
- [x] Verify successful attempt for Travis to generate file
- [x] Verify successful attempt to make subsequent commits after Travis generates file

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -22,12 +22,14 @@ if [ -n "$GITHUB_TOKEN" ]; then
   git checkout $TRAVIS_BRANCH
 fi
 
-# Check for invalid changes to overrides yaml files
+# Check for invalid changes to generated yaml files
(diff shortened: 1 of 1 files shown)


maimaisie:
Will this work if multiple commits are pushed to the branch at once and one of the commits but not the most recent one contains manual changes to the generated files?

rvmiller89:
@maimaisie Yes it should.  Initially I thought of just looking at the most recent commit, but decided not to for that reason you gave. Instead, it looks at all commits in the branch from the author (excluding Travis author) and will fail the build if any of the commits touch a generated file.

maimaisie:
Oh nice. Thanks for the fix Ryan!

PR #221, opened 2019-09-27
Pin gem dependency versions; Fix metadata version string

rvmiller89:
###### Description

Root issue for the problem:
```
ERROR:  Could not find a valid gem 'kubeclient' (~> 4.4.0) (required by 'fluent-plugin-enhance-k8s-metadata' (>= 0)) in any repository
```
was that in our metadata plugin we were defining required version as `~>4.4.0` which only applies to `>= 4.4.0` and `< 4.5`.  Overnight `kubeclient` released a `4.5.0` version which now fails the required version check.

This PR fixes that issue, and also pins our external gem dependency versions to avoid surprises since otherwise our CI could randomly start failing at any time as projects release new major versions.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/fluent-plugin-enhance-k8s-metadata/fluent-plugin-enhance-k8s-metadata.gemspec
+++ b/fluent-plugin-enhance-k8s-metadata/fluent-plugin-enhance-k8s-metadata.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'concurrent-ruby', '~> 1.1'
   spec.add_runtime_dependency 'fluentd', ['>= 0.14.10', '< 2']
-  spec.add_runtime_dependency 'kubeclient', '~> 4.4.0'
+  spec.add_runtime_dependency 'kubeclient', '~> 4.4'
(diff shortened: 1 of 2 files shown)


fluent-plugin-enhance-k8s-metadata/fluent-plugin-enhance-k8s-metadata.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'concurrent-ruby', '~> 1.1'
   spec.add_runtime_dependency 'fluentd', ['>= 0.14.10', '< 2']
-  spec.add_runtime_dependency 'kubeclient', '~> 4.4.0'
+  spec.add_runtime_dependency 'kubeclient', '~> 4.4'
```
lei-sumo: what does this file do after pinning a static version?

rvmiller89: this allows `>= 4.4 and `< 5.0`

PR #224, opened 2019-10-03
Add instructions on deploying second prometheus in the same cluster

maimaisie:
###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -15,7 +15,7 @@ This page has instructions for collecting Kubernetes logs, metrics, and events; 
 			- [Overwrite Prometheus Remote Write Configuration](#overwrite-prometheus-remote-write-configuration) 
 			- [Merge Prometheus Remote Write Configuration](#merge-prometheus-remote-write-configuration) 
 		- [How to install if you have standalone Prometheus](#how-to-install-if-you-have-standalone-prometheus) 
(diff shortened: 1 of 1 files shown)


deploy/README.md
```diff
@@ -235,6 +236,23 @@ Update your Prometheus configuration file’s `remote_write` section, as per the
 * `writeRelabelConfigs:` change to `write_relabel_configs:`
 * `sourceLabels:` change to `source_labels:`
 
+### How to install our Prometheus side by side with your existing Prometheus
+
+It is possible to have more than one Prometheus running in the same cluster when you have your existing Prometheus running in the cluster that you want to deploy our Chart with our Prometheus to. When installing our Helm Chart, set the following fields for the Prometheus node exporter to use a different port number than the default port number 9100. For example:
```
drduke1: I'm not following the logic of this paragraph. I don't like the way this sounds "that you want to deploy our Chart with our Prometheus to."
Are we providing details on how to configure our Helm Chart with an existing Prometheus, where they would need to change the port?

maimaisie: > Are we providing details on how to configure our Helm Chart with an existing Prometheus, where they would need to change the port?

Yes, basically when a user wants to deploy two Prometheus into the same cluster, either when they have an existing Prometheus and want to deploy our Chart with our Prometheus (which results in 2 Prometheus running in the cluster), or deploy our solution twice in the same cluster

PR #227, opened 2019-10-07
check exact secret name in setup script

samjsong:
###### Description

`kubectl describe secret sumologic` will return results if you have any secret whose name starts with `sumologic`

so if you have any secret named `sumologic.*`, the setup script will fail. On the other hand, `kubectl describe secret/sumologic` will only return results for an exact match on a secret named `sumologic`.

###### Testing performed

- [ ] ci/build.sh


christophla:
Thanks. We needed this patch.

PR #229, opened 2019-10-09
Bump fluent-plugin-prometheus version to fix breaking change from prometheus-client

maimaisie:
###### Description

The `fluent-plugin-prometheus` gem has a dependency on `prometheus-client` which recently introduced a breaking change in their latest release, describe in https://github.com/fluent/fluent-plugin-prometheus/issues/119. `fluent-plugin-prometheus` made a new release 1.6.1 to use a version lower than the latest. To unblock our release we should try this new version and verify it doesn't create other issues.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -30,7 +30,7 @@ RUN gem install fluent-plugin-systemd -v 1.0.2 \
        && gem install fluent-plugin-sumologic_output -v 1.5.0 \
        && gem install fluent-plugin-concat -v 2.4.0 \
        && gem install fluent-plugin-rewrite-tag-filter -v 2.2.0 \
-       && gem install fluent-plugin-prometheus -v 1.5.0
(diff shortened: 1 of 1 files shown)


rvmiller89:
I think we should still perform all 3 of the testing checkboxes before merging this

maimaisie:
It's multiple steps to run the test, building a local image, downloading the yaml / or running helm by overriding image, manually verify ingestion. In the case where things are already broken, I wonder if we could just merge and see if this helps

rvmiller89:
If the manual testing is a burden, perhaps we should prioritize automating it as part of the CD?

If we do merge this, are we planning on running the Jenkins jobs to see if they pass?

maimaisie:
@rvmiller89 Yeah I agree. And I will definitely make sure Jenkins jobs pass since we are planning to make a release and this is blocking it.

@lei-sumo I have verified the latest 0.8.0 release is not affected.

PR #232, opened 2019-10-11
remove log_format, attach fields by default

samjsong:
###### Description

I recommend reviewing commit by commit (maybe also with "hide whitespace changes"), hopefully that makes it easier to read the changes

By removing the `log_format` config, the filter plugin now always attaches the metadata as `_sumo_metadata.fields`.

We did get rid of `add_timestamp` in the output plugin, so the logs that come in no longer have `timestamp` (but still have `time`). We'll have to explicitly decide what we want to do by default to make those config changes in the fluentd yaml.

**TODO:** use `sumologic` output plugin version `2.0.0` in DockerFile when gem is released, once https://github.com/SumoLogic/fluentd-output-sumologic/pull/53 is merged

###### Testing performed

- [x] ci/build.sh
- [x] Tested with locally built `sumologic` output plugin to verify behaviour

### DIFF
--- a/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
@@ -25,7 +25,7 @@
     cache_ttl "#{ENV['K8S_METADATA_FILTER_BEARER_CACHE_TTL']}"
     tag_to_kubernetes_name_regexp '.+?\.containers\.(?<pod_name>[^_]+)_(?<namespace>[^_]+)_(?<container_name>.+)-(?<docker_id>[a-z0-9]{64})\.log$'
   </filter>
-  <filter **>
+  <filter containers.**>
     @type enhance_k8s_metadata
     in_namespace_path '$.kubernetes.namespace_name'
     in_pod_path '$.kubernetes.pod_name'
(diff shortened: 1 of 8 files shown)


deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -48,7 +48,19 @@
...
+  #     message ${record["message"]["log"]}
+  #   </record>
+  # </filter>
+  # <filter containers.**>
+  #   @type record_modifier
+  #   <record>
+  #     message ${begin; record["message"].merge(JSON.parse(record["message"]["log"])).delete("log"); rescue JSON::ParserError; record["message"]; end}
+  #   </record>
+  # </filter>
+  <match containers.**>
```
maimaisie: did we want to add before and after examples for each filter here?

samjsong: good catch, will do that

fluent-plugin-kubernetes-sumologic/lib/fluent/plugin/filter_kubernetes_sumologic.rb
```diff
@@ -156,77 +134,43 @@ def filter(tag, time, record)
           return nil
         end
 
-        sumo_metadata[:log_format] = annotations["sumologic.com/format"] if annotations["sumologic.com/format"]
-
-        if annotations["sumologic.com/sourceHost"].nil?
-          sumo_metadata[:host] = sumo_metadata[:host] % k8s_metadata
-        else
-          sumo_metadata[:host] = annotations["sumologic.com/sourceHost"] % k8s_metadata
```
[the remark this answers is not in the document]
samjsong: good question - for default source name, we use something like `%{namespace}.%{pod}.%{container}`, so if `k8s_metadata` contains keys for `namespace`, `pod`, or `container`, it will populate those values so we get a final source name like `sumologic.fluentd-xxx.fluentd` or something like that

I believe this was how we supported dynamically populating metadata values in our sumo metadata.

PR #237, opened 2019-10-18
Create Installation_with_Helm.md

drduke1:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/docs/Installation_with_Helm.md
+++ b/deploy/docs/Installation_with_Helm.md
@@ -29,7 +29,7 @@ The Helm chart installation requires three parameter overrides:
 
 ## Installation Steps
 
-These steps require that no Premotheus exists. If you already have prometheus installed select from the following options:
+These steps require that no Prometheus exists. If you already have Prometheus installed select from the following options:
 
 - [How to install if you have an existing Prometheus operator](https://github.com/SumoLogic/sumologic-kubernetes-collection/deploy/docs/existingPrometheusDoc.md) 
 - [How to install if you have standalone Prometheus](https://github.com/SumoLogic/sumologic-kubernetes-collection/deploy/docs/standAlonePrometheus.md) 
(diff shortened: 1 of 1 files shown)


deploy/docs/Installation_with_Helm.md
```diff
@@ -0,0 +1,98 @@
...
+sed 's/\collection'"/my-release/g" > values.yaml
+```
+
+Make any changes to the `values.yaml` file as needed, and run the following to install the chart with the override file.
+
+```bash
+helm install sumologic/sumologic --name my-release --namespace my-namespace -f values.yaml --set sumologic.endpoint=<SUMO_ENDPOINT> --set sumologic.accessId=<SUMO_ACCESS_ID> --set sumologic.accessKey=<SUMO_ACCESS_KEY> 
+```
+
+If you would like to use a different cluster name than the default `kubernetes`, add this to the `helm install` command:
```
samjsong: Can we move this up to line 49, right under the first `helm install` command on line 47? Most users want to specify a cluster name and the feedback we received was that this was too far down and most users have already installed by the time they read this and have to reinstall.

deploy/docs/Installation_with_Helm.md
```diff
@@ -0,0 +1,98 @@
...
+- cluster
+- container
+- deployment
+- host
+- namespace
+- node
+- pod
+- service
+
+The Helm chart installation requires three parameter overrides:
```
frankreno: Can we document the clusterName and externalLabels here or at least before the install steps? It is always missed

deploy/docs/Installation_with_Helm.md
frankreno: +1

deploy/docs/Installation_with_Helm.md
```diff
@@ -0,0 +1,98 @@
...
+```bash
+helm install sumologic/sumologic --name my-release --namespace my-namespace -f values.yaml --set sumologic.endpoint=<SUMO_ENDPOINT> --set sumologic.accessId=<SUMO_ACCESS_ID> --set sumologic.accessKey=<SUMO_ACCESS_KEY> 
+```
+
+If you would like to use a different cluster name than the default `kubernetes`, add this to the `helm install` command:
+```
+--set prometheus-operator.prometheus.prometheusSpec.externalLabels.cluster="<my-cluster-name>" --set sumologic.clusterName="<my-cluster-name>"
+```
+
+## Uninstalling the Chart
```
frankreno: Since our process is not yet idempotent we should doc deleting the secret and collector

PR #242, opened 2019-10-18
Create Troubleshoot_Collection.md

drduke1:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/docs/Troubleshoot_Collection.md
+++ b/deploy/docs/Troubleshoot_Collection.md
@@ -1,0 +1,249 @@
+# Troubleshooting Collection
+
+<!-- TOC -->
+
+- [Namespace configuration](#namespace-configuration) 
+- [Gathering logs](#gathering-logs) 
+  - [Fluentd Logs](#fluentd-logs) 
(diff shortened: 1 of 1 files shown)


deploy/docs/Troubleshoot_Collection.md
```diff
@@ -0,0 +1,249 @@
...
+
+and redeploy Prometheus:
+```
+helm del --purge prometheus-operator
+helm install stable/prometheus-operator --name prometheus-operator --namespace sumologic -f prometheus-overrides.yaml
+```
+
+### Missing `kube-controller-manager` or `kube-scheduler` metrics
+
+There’s an issue with backwards compatibility in the current version of the prometheus-operator helm chart that requires us to override the selectors for kube-scheduler and kube-controller-manager in order to see metrics from them. If you are not seeing metrics from these two targets, try running the commands in the "Configure Prometheus" section [above](#missing-metrics-for-controller-manager-or-scheduler).
```
frankreno: THis is now missing, its in the deploy docs

PR #248, opened 2019-10-21
Dockerfile changes for Terraform support

maimaisie:
###### Description

Dockerfile changes to install Terraform dependencies.

This needs to be merged and docker image built with a new Helm chart release before we can merge https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/249

###### Testing performed

- [x] ci/build.sh
- [x] Redeploy fluentd and fluentd-events pods
- [ ] Confirm events, logs, and metrics are coming in

PR #249, opened 2019-10-21
Update setup.sh to use Terraform

maimaisie:
###### Description

Adds Terraform-based setup for Sumo collector and HTTP sources, as well as Kubernetes namespace and secret.

Requires https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/248 to be merged, and a Helm chart release (minor version, non-alpha) so that the docker image contains the required dependencies.

~~Do not merge until we have a Helm chart release with the updated Dockerfile~~
Dockerfile changes are in Helm chart release 0.10.0

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
@@ -1,0 +1,143 @@
+{{- if .Values.sumologic.setupEnabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name:  {{ template "sumologic.fullname" . }}-setup
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+  labels:
+    app: {{ template "sumologic.labels.app" . }}
+    {{ template "sumologic.labels.common" . }}
+data:
+  setup.sh: |-
+    #!/bin/sh
+
+    mkdir /test
+    cp /etc/terraform/* /test/
+    cd /test
+    terraform init
+
+    terraform import -config=/test kubernetes_namespace.sumologic_collection_namespace {{ .Release.Namespace }}
+    terraform import -config=/test kubernetes_secret.sumologic_collection_secret {{ .Release.Namespace }}/sumologic
+
+    terraform apply -auto-approve
+  sumo-k8s.tf: |-
+    variable "cluster_name" {
+      type  = string
+      default = "{{ .Values.sumologic.clusterName }}"
+    }
+
+    variable "namespace_name" {
+      type  = string
+      default = "{{ .Release.Namespace }}"
+    }
+
+    locals {
+      default-metrics-source-name                 = "(default-metrics)"
+      apiserver-metrics-source-name               = "apiserver-metrics"
+      events-source-name                          = "events"
+      kube-controller-manager-metrics-source-name = "kube-controller-manager-metrics"
+      kube-scheduler-metrics-source-name          = "kube-scheduler-metrics"
+      kube-state-metrics-source-name              = "kube-state-metrics"
+      kubelet-metrics-source-name                 = "kubelet-metrics"
+      logs-source-name                            = "logs"
+      node-exporter-metrics-source-name           = "node-exporter-metrics"
+    }
+
+    provider "sumologic" {}
+
+    resource "sumologic_collector" "collector" {
+        name  = var.cluster_name
+        fields  = {
+          cluster = var.cluster_name
+        }
+    }
+
+    resource "sumologic_http_source" "default_metrics_source" {
+        name         = local.default-metrics-source-name
+        category     = "${var.cluster_name}/${local.default-metrics-source-name}"
+        collector_id = "${sumologic_collector.collector.id}"
+    }
+
+    resource "sumologic_http_source" "apiserver_metrics_source" {
+        name         = local.apiserver-metrics-source-name
+        category     = "${var.cluster_name}/${local.apiserver-metrics-source-name}"
+        collector_id = "${sumologic_collector.collector.id}"
+    }
+
+    resource "sumologic_http_source" "events_source" {
+        name         = local.events-source-name
+        category     = "${var.cluster_name}/${local.events-source-name}"
+        collector_id = "${sumologic_collector.collector.id}"
+    }
+
+    resource "sumologic_http_source" "kube_controller_manager_metrics_source" {
+        name         = local.kube-controller-manager-metrics-source-name
+        category     = "${var.cluster_name}/${local.kube-controller-manager-metrics-source-name}"
+        collector_id = "${sumologic_collector.collector.id}"
+    }
+
+    resource "sumologic_http_source" "kube_scheduler_metrics_source" {
+        name         = local.kube-scheduler-metrics-source-name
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/templates/setup/setup-configmap.yaml
```diff
@@ -0,0 +1,143 @@
...
+  labels:
+    app: {{ template "sumologic.labels.app" . }}
+    {{ template "sumologic.labels.common" . }}
+data:
+  setup.sh: |-
+    #!/bin/sh
+
+    mkdir /test
+    cp /etc/terraform/* /test/
+    cd /test
```
maimaisie: L17-19 is because of [this issue](https://wiki.onap.org/pages/viewpage.action?pageId=29787473) with mounted volume

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -16,19 +16,30 @@ spec:
...
+        volumeMounts:
+        - name: setup
+          mountPath: /etc/terraform
+        env:
+        - name: SUMOLOGIC_ACCESSID
+          value: {{ required "A valid .Values.sumologic.accessId entry required!" .Values.sumologic.accessId }}
+        - name: SUMOLOGIC_ACCESSKEY
+          value: {{ required "A valid .Values.sumologic.accessKey entry required!" .Values.sumologic.accessKey }}
+        - name: SUMOLOGIC_ENVIRONMENT
+          value: us2
```
maimaisie: One thing I forgot to mention is here Terraform expects the deployment name rather than the endpoint, so we'll need to do some if/else based on the endpoint to give the corresponding deployment name

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -16,19 +16,30 @@ spec:
...
-          "-d", "false", 
-          "-y", "false", 
-          '{{ required "A valid .Values.sumologic.endpoint entry required!" .Values.sumologic.endpoint }}',
-          '{{ required "A valid .Values.sumologic.accessId entry required!" .Values.sumologic.accessId }}', 
-          '{{ required "A valid .Values.sumologic.accessKey entry required!" .Values.sumologic.accessKey }}']
+        image: test1:latest
+        imagePullPolicy: Never
+        command: ["/etc/terraform/setup.sh"]
+        securityContext:
+          runAsUser: 0
```
maimaisie: I tried to remove this but `mkdir` wouldn't work because of permission issue. I bash'ed into the container, it looks like `fluent` user can only create directory in the `/fluentd`

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -16,19 +16,30 @@ spec:
...
+      volumes:
+      - name: setup
+        configMap:
+          name: {{ template "sumologic.fullname" . }}-setup
+          defaultMode: 0777
       containers:
+      - name:  kubectl
+        image: gcr.io/google_containers/kubectl:v1.0.7
+        command: ["/kubectl"]
+        args: ["proxy", "-p", "8001"]
```
maimaisie: Need to ask @frankreno about potential security concern on exposing a port in the container. Is there other ways to initialize the `kubernetes` provider?

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -16,19 +16,30 @@ spec:
     spec:
       restartPolicy: OnFailure
       serviceAccountName: {{ template "sumologic.fullname" . }}-setup
+      volumes:
+      - name: setup
+        configMap:
+          name: {{ template "sumologic.fullname" . }}-setup
+          defaultMode: 0777
```
maimaisie: Can be removed since kubernetes doesn't honor this field for configmap?

deploy/helm/sumologic/templates/setup/setup-job.yaml
rvmiller89: I think we just need to `chown` this directory to `fluent` user. Will try today

deploy/helm/sumologic/templates/setup/setup-job.yaml
rvmiller89: we can skip environment and use `base_url` instead: https://github.com/SumoLogic/sumologic-terraform-provider/blob/master/docs/sumologic-provider.md

deploy/helm/sumologic/templates/setup/setup-configmap.yaml
```diff
@@ -0,0 +1,143 @@
...
+    terraform init
+
+    terraform import -config=/test kubernetes_namespace.sumologic_collection_namespace {{ .Release.Namespace }}
```
rvmiller89: we need to import collectors/sources too

deploy/helm/sumologic/templates/setup/setup-job.yaml
frankreno: That was only intended to be for the demo I have you guys.  We will need to support proper authentication and see if we can do this - https://www.terraform.io/docs/providers/kubernetes/index.html#authentication

deploy/helm/sumologic/templates/setup/setup-job.yaml
rvmiller89: This one needs to stay

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -14,18 +14,14 @@ metadata:
 spec:
   template:
     spec:
-      restartPolicy: Never
+      restartPolicy: OnFailure
```
frankreno: Will this cause it to never eventually die?

deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -16,19 +16,24 @@ spec:
...
-          "-d", "false", 
-          "-y", "false", 
-          '{{ required "A valid .Values.sumologic.endpoint entry required!" .Values.sumologic.endpoint }}',
-          '{{ required "A valid .Values.sumologic.accessId entry required!" .Values.sumologic.accessId }}', 
-          '{{ required "A valid .Values.sumologic.accessKey entry required!" .Values.sumologic.accessKey }}']
+        command: ["/etc/terraform/setup.sh"]
+        volumeMounts:
+        - name: setup
+          mountPath: /etc/terraform
+        env:
```
frankreno: This may not need to be solved in this PR, but this new process (and existing) do not work when there is a proxy involved.  However, Terraform supports using a proxy by setting HTTP_PROXY/HTTPS_PROXY environment variables.  Would be good to expose that.

deploy/helm/sumologic/templates/setup/setup-job.yaml
rvmiller89: yeah, my thinking was we shouldn't proceed if the setup fails, since it will be missing collectors/sources/namespace/secret. By restarting on failure, it gives them the chance to modify the config map to fix the issue -OR- cancel the helm install and skip `setupEnabled` if they want to manage it manually.

rvmiller89:
Thanks for reviewing, I plan on adding one small change today to account for customers that set endpoint URL including `/v1` since this will actually fail against our TF provider's endpoint URL.

After that, we will need to wait for a Helm chart release that has the updated Docker image with the TF providers before merging.

PR #250, opened 2019-10-22
optimize remote write settings

frankreno:
This PR simplfies the remote write configuration and also imporves memory and CPU performance by reducing the number of remote_write blocks, which are all seperate queues that read from the WAL. Testing shows roughly a 3x reduction in memory and CPU usage.

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -295,274 +295,66 @@ prometheus-operator:
         # Set this to a value to distinguish between different k8s clusters
         cluster: kubernetes
       remoteWrite:
-      # kube state metrics
-      - url: http://collection-sumologic.sumologic.svc.cluster.local:9888/prometheus.metrics.state.statefulset
-        writeRelabelConfigs:
-        - action: keep
-          regex: kube-state-metrics;kube_statefulset_status_(?:observed_generation|replicas)
-          sourceLabels: [job, __name__]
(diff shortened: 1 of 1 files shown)


frankreno:
It also makes the configuration simpler by having a 1:1 remote write block per source/target, so its easier to add things (and exclude them)

deploy/helm/sumologic/values.yaml
```diff
@@ -295,274 +295,66 @@ prometheus-operator:
...
-            regex: 'fluentd_.*'
-            sourceLabels: [__name__]
-      # fluent-bit metrics
-      - url: http://collection-sumologic.sumologic.svc.cluster.local:9888/prometheus.metrics
-        writeRelabelConfigs:
-          - action: keep
-            regex: 'fluentbit.*'
-            sourceLabels: [__name__]
+        # kube state metrics
+        - url: http://collection-sumologic.sumologic.svc.cluster.local:9888/prometheus.metrics.state
```
lei-sumo: Will you need to change the plugin as tags are changed?

frankreno: Don't believe so, but need @samjsong to confirm.  In my testing, all things worked like I expected, data volume diffs were accounted for.  I do not believe we depend on on the tag in the plugins as we use a label to handle the datapoint stuff.

samjsong: we do depend on the tag in the plugins, we no longer do the retag based on job to make it clearer to the user where the metrics were going. In this case the tags are correct since we check for `<match prometheus.metrics.state**>` etc.

PR #262, opened 2019-10-29
Update Travis CI to support PR push

rvmiller89:
###### Description

1. Only pushing to master (merging a PR) will run CI with `push` event type
2. Otherwise, all PRs will run CI with `pull_request` event type. In this case we look at `TRAVIS_PULL_REQUEST_BRANCH` in order to detect changes made to generated files, or changes that need to be committed for generated files.

FYI: @maimaisie 

###### Testing performed

- [x] ci/build.sh
- [x] Test making invalid change and verifying it fails
- [x] Test making valid change and verify it commits the generated yaml files
- [x] Test making subsequent valid change and verify it commits the updated generated files

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -19,11 +19,11 @@ if [ -n "$GITHUB_TOKEN" ]; then
   git config --global user.name "Travis CI"
   git remote add origin-repo https://${GITHUB_TOKEN}@github.com/SumoLogic/sumologic-kubernetes-collection.git > /dev/null 2>&1
   git fetch origin-repo
-  git checkout $TRAVIS_BRANCH
+  git checkout $TRAVIS_PULL_REQUEST_BRANCH
 fi
 
 # Check for invalid changes to generated yaml files (non-Tag builds)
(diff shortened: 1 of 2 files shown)


ci/build.sh
```diff
@@ -70,7 +70,7 @@ echo "Test docker image locally..."
 ruby deploy/test/test_docker.rb
 
 # Check for changes that require re-generating overrides yaml files
-if [ "$TRAVIS_BRANCH" != "master" ] && [ "$TRAVIS_EVENT_TYPE" == "push" ] && [ -n "$GITHUB_TOKEN" ]; then
+if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
```
rvmiller89: Ah, @samjsong , PRs submitted from outside repos like https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/264 wont' have `$GITHUB_TOKEN` .  So this will require some more thought.

PR #266, opened 2019-10-30
adding matching template variables

djsly:
###### Description

missing matching template variable definition

###### Testing performed

- [ ] ci/build.sh
- [ ] Redeploy fluentd and fluentd-events pods
- [ ] Confirm events, logs, and metrics are coming in

PR #267, opened 2019-10-30
Fix helper template variables

rvmiller89:
###### Description

In case we can't get https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/266 to work

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -30,8 +30,8 @@ If dryRun=true, we do not create any chart labels.
 {{- define "sumologic.labels.common" -}}
 {{- if .Values.dryRun -}}
 {{- else -}}
-    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
-    release: "{{ .Release.Name }}"
-    heritage: "{{ .Release.Service }}"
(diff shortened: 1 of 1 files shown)


djsly:
This goes with #265, since nindent is now used, the strings should now have leading spaces

djsly:
@rvmiller89 when you get a chance to merge this, could we release an alpha tag to allow a public chart to be accessible? our logic is currently based on your public helm REPO and I don't have an easy way to inject local charts into my pipelines.

rvmiller89:
@djsly I'll bring it up to the team about releasing Helm charts for alpha tag versions, right now I think our concern is that by publishing the alpha versions in https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/gh-pages/index.yaml the default `helm install sumologic/sumologic` would download and install the latest alpha chart, rather than the stable tested release.

djsly:
Thanks. 

I haven’t tried it for say but it looks like helm will not fetch pre release version in a repo anymore unless the -version is used to explicitly chose one. 

https://github.com/helm/helm/issues/2407

My gut feeling is that it should be safe to push some -alpha as long as they follow the pre-release format from SemVer2

PR #269, opened 2019-11-01
Start pushing alpha helm chart releases with every alpha build

rvmiller89:
###### Description

This will start generating alpha helm chart releases in our Github Pages-hosted chart: https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/gh-pages

Will need to merge to test this on Travis CI

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -143,7 +143,7 @@ if [ -n "$DOCKER_PASSWORD" ] && [ -n "$TRAVIS_TAG" ] && [[ $TRAVIS_TAG != *alpha
   # Push new helm release
   git checkout -- .
   sudo helm init --client-only
-  sudo helm package deploy/helm/sumologic --version=$VERSION
+  sudo helm package deploy/helm/sumologic --version=$VERSION --app-version=$VERSION
   git fetch origin-repo
   git checkout gh-pages
   sudo helm repo index ./ --url https://sumologic.github.io/sumologic-kubernetes-collection/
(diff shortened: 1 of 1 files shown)


ci/build.sh
```diff
@@ -143,7 +143,7 @@ if [ -n "$DOCKER_PASSWORD" ] && [ -n "$TRAVIS_TAG" ] && [[ $TRAVIS_TAG != *alpha
   # Push new helm release
   git checkout -- .
   sudo helm init --client-only
-  sudo helm package deploy/helm/sumologic --version=$VERSION
+  sudo helm package deploy/helm/sumologic --version=$VERSION --app-version=$VERSION
```
lei-sumo: why do we need to change this?

rvmiller89: Helm seems to be doing this already, but it makes me nervous relying on that undocumented behavior that could change.  So to be explicit I added this.

djsly:
It's working well

```
k8s-collection          1               Fri Nov  1 14:30:22 2019        DEPLOYED        sumologic-0.10.5-alpha          0.10.5-alpha    sumologic 
```

I updated our helm_repo_url with the `/alpha` and I can install 0.10.5-alpha!

thanks @rvmiller89

rvmiller89:
@djsly I'll be deleting this `/alpha` shortly, the team decided against storing the alpha builds separately. Please hold on using the alpha builds.

rvmiller89:
@djsly we just set up the CI to generate an alpha helm chart release for every alpha build.  They're hosted at the same URL as before (the /alpha URL will be going away soon).

To use an alpha helm chart release (and alpha image), you can do:
```
$ helm repo update
$ helm install sumologic/sumologic --name collection --namespace sumologic -f values.yaml --set=image.tag=0.10.10-alpha --version=0.10.10-alpha
```

djsly:
Perfect. 
Thanks for the ends up.

rvmiller89:
@djsly just wanted to let you know we put together a small doc on the alpha releases, please let us know if you think of anything worth adding: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/Alpha_Release_Guide.md

djsly:
I think it’s looking good. 

One thing though. I’m using a alpha chart release with an official docker image. But I guess it’s simpler to ask users to always aligned since this is what you tested. 

I just think that in some cases of a chart change only. A users might be interrelated in picking up the chart release only.

PR #277, opened 2019-11-05
clean up deploy readme

samjsong:
###### Description

Delete all the duplicated sections in the deploy readme, as well as update the links in the TOC to reference the proper external readmes


deploy/README.md
```diff
@@ -5,66 +5,15 @@ This page has instructions for collecting Kubernetes logs, metrics, and events;
 <!-- TOC -->
```
rvmiller89: can you get rid of this and `<!-- /TOC -->` below as well, it will conflict with the TOC generator plugin I have

PR #282, opened 2019-11-08
Bump Dockerfile Sumo TF provider to 2.0.0

rvmiller89:
###### Description

With the new TF provider, we no longer require specifying endpoint URL since it will try to follow the 301 redirection response to the correct deployment.

###### Testing performed

- [x] ci/build.sh
- [x] Local helm chart install with local built Docker image, specifying US2 access key/id but *not* specifying endpoint URL, verified the TF Provider automatically redirected and created collector and sources on US2 🎉 

### DIFF
--- a/deploy/kubernetes/setup-sumologic.yaml.tmpl
+++ b/deploy/kubernetes/setup-sumologic.yaml.tmpl
@@ -7,7 +7,7 @@ kind: ConfigMap
 metadata:
   name:  collection-sumologic-setup
   annotations:
-    "helm.sh/hook": pre-install
+    "helm.sh/hook": pre-install,pre-upgrade
     "helm.sh/hook-weight": "2"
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -7,7 +7,7 @@ kind: ConfigMap
 metadata:
   name:  collection-sumologic-setup
   annotations:
-    "helm.sh/hook": pre-install
+    "helm.sh/hook": pre-install,pre-upgrade
```
rvmiller89: I think these generated changes were missed in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/281 which may have been built using an older master with older `build.sh` script

PR #284, opened 2019-11-11
Create Upgrading_with_Helm.md

vsinghal13:
###### Description

Create new page for helm upgrade info and steps and add it in the main README.md 

###### Testing performed

- [ ] ci/build.sh


deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,85 @@
+
+# Upgrading with Helm
+
+When a new version of a chart is released, or when you want to change the configuration of your release, you can use the helm upgrade command.
+
+This command upgrades a release to a specified version of a chart and/or updates chart values.
+
+Required arguments are release and chart.
+
+The chart argument can be one of: - a chart reference; use `–version` flag for versions other than latest, - a path to a chart directory, - a packaged chart, - a fully qualified URL.
```
drduke1: Should the arguments be in a list? The markdown isn't rendering a list when I view these.

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,85 @@
...
+```bash
+$ helm upgrade collection sumologic/sumologic --reuse-values -f myvalues.yaml -f override.yaml
+```
+
+In the above case, the `collection` release is upgraded with the same chart, but with a new  key-value pair in the YAML file:
+```bash
+Test: true
+```
+  
+We can use helm get values to see whether that new setting took effect.
```
drduke1: Can we put "get values" in a syntax format?

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,85 @@
...
+- `--set` (and its variants `--set-string` and `--set-file`): Specify overrides on the command line.
+
+
+```bash
+
+helm upgrade [RELEASE] [CHART] [flags]
+
+```  
+
+Some useful Options
```
drduke1: Can we put these in a table? The spacing isn't rendering consistent.

deploy/docs/Upgrading_with_Helm.md
vsinghal13: fixed it.

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+To edit or append to the existing customized values, add the `–reuse-values` flag, otherwise any existing customized values are ignored.
+
+If no chart value arguments are provided on the command line, any existing customized values are carried forward, i.e. the `--reuse-values` flag will be used by default. If you want to revert to just the values provided in the chart, use the `-–reset-values` flag.
+
+And if there exists `--set/--set-file/--set-string/--values/-f` , the `--reset-values` flag will be used by default.
+
+### Example using the `--set` flag in the upgrade command:
+
+```bash
+helm upgrade collection sumologic/sumologic --set key1=val1, key2=val2
```
rvmiller89: shouldn't we advise setting `--reuse-values` here? It will fail for users unless they happen to be setting access key/id/endpoint.

maimaisie:
I feel like this is more like documentation for the helm upgrade command itself instead of helm upgrade for the Sumo chart. I wonder whether we should just make our docs short and simple without having to explain fundamentals of how helm works, but include links to the helm official docs? It should be very clear to users which commands they need to run for different cases (that are related to our chart)

deploy/docs/Upgrading_with_Helm.md
samjsong: +1, let's add `--reuse-values` in the example

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+Test: true
+```
+  
+We can use helm `get values` to see whether that new setting took effect.
+```bash
+$ helm get values collection
+Test: true
+```
+
+The helm get command is a useful tool for looking at a release in the cluster. And as we can see above, it shows that our new values from `override.yaml` were deployed to the cluster.
```
samjsong: Can we also wrap "helm get" as an in-line command, like "The `helm get` command is..."?

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+$ helm upgrade collection sumologic/sumologic --reuse-values -f myvalues.yaml -f override.yaml
+```
+
+In the above case, the `collection` release is upgraded with the same chart, but with a new  key-value pair in the YAML file:
+```bash
+Test: true
+```
+  
+We can use helm `get values` to see whether that new setting took effect.
```
samjsong: Can we wrap the entire command, like `helm get values`?

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Example using the `-f` flag in the upgrade command:
+
+You can specify any of the chart value flags multiple times. The priority will be given to the last (right-most) value specified. For example, if both `myvalues.yaml` and `override.yaml` contained a key called ‘Test’, the value set in `override.yaml` would take precedence:  
+
+```bash
+$ helm upgrade collection sumologic/sumologic --reuse-values -f myvalues.yaml -f override.yaml
+```
+
+In the above case, the `collection` release is upgraded with the same chart, but with a new  key-value pair in the YAML file:
```
samjsong: There's an extra space between "new" and "key-value"

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+# Upgrading with Helm
+
+When a new version of a chart is released, or when you want to change the configuration of your release, you can use the helm upgrade command.
+
+This command upgrades a release to a specified version of a chart and/or updates chart values.
+
+Required arguments are release and chart.
+
+The chart argument can be one of: 
+- a chart reference; use `–version` flag for versions other than latest,
```
samjsong: I think there are two dashes for `--version`

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+
+The chart argument can be one of: 
+- a chart reference; use `–version` flag for versions other than latest, 
+- a path to a chart directory, 
+- a packaged chart, 
+- a fully qualified URL.
+
+An upgrade takes an existing release and upgrades it according to the information you provide. Because Kubernetes charts can be large and complex, Helm tries to perform the least invasive upgrade. It will only update things that have changed since the last release.
+
+If required, there are two ways to pass configuration data:
```
samjsong: nit: maybe `If desired` instead of `If required`?

deploy/docs/Upgrading_with_Helm.md
```diff
@@ -0,0 +1,78 @@
...
+- `--set` (and its variants `--set-string` and `--set-file`): Specify overrides on the command line.
+
+
+```bash
+
+helm upgrade [RELEASE] [CHART] [flags]
+
+```  
+
+Some useful Options
```
samjsong: nit: personally I think `options` can be lowercased

samjsong:
Hmm, Maisie makes a great point. At the least though, I think we should keep the examples that include the `--reuse-values` flag since I think those will be helpful for users

vsinghal13:
Since we were creating a new page, I thought adding details about the `helm upgrade` would be helpful. I agree with Maisie's point of linking to the official docs and keeping it simple. Will make the required changes .

frankreno:
LGTM but do agree with @maimaisie that we don't need to repeat their docs.  A simple pointer to them is fine. Definitely like the `--reuse-values` flag being included.

PR #302, opened 2019-11-19
Customizable annotations for setup resources

matthewdaltonamount:
###### Description

I think that the ability to customize the annotations for helm hooks is important enough to move it out into the values to make it override-able. I also fixed some of the formatting around setup in the values to make it expandable in the future.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-clusterrole.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-clusterrole.yaml
@@ -3,10 +3,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name:  {{ template "sumologic.fullname" . }}-setup
-  {{- if .Values.sumologic.setup.clusterRole.annotations  }}
   annotations:
 {{ toYaml .Values.sumologic.setup.clusterRole.annotations | indent 4 }}
-  {{- end }}
   labels:
     app: {{ template "sumologic.labels.app" . }}
     {{- include "sumologic.labels.common" . | nindent 4 }}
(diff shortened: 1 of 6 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -30,22 +30,43 @@ eventsDeployment:
...
+    configMap:
+      annotations:
+        helm.sh/hook: pre-install
+        helm.sh/hook-weight: 2
+        helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    job:
+      #envFromSecret: sumo-api-secret
+
+      # Sumo access ID
+      #accessId: ""
```
rvmiller89: changing these will break automation passing `--set sumologic.accessId=x` etc. It would also require a docs update.

Would prefer we keep these auth variables as they were before.

matthewdaltonamount: Sounds good

deploy/helm/sumologic/templates/setup/setup-clusterrole.yaml
```diff
@@ -3,10 +3,10 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name:  {{ template "sumologic.fullname" . }}-setup
+  {{- if .Values.sumologic.setup.clusterRole.annotations  }}
   annotations:
-    "helm.sh/hook": pre-install
-    "helm.sh/hook-weight": "1"
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+{{ toYaml .Values.sumologic.setup.clusterRole.annotations | indent 4 }}
```
rvmiller89: does this work the same without the `if` conditional? That is, if `.Values.sumologic.setup.clusterRole.annotations` doesn't exist, it simply won't add it?

Could we skip the `{{- if ... }}` and `{{- end }}` in all of these?

matthewdaltonamount: oh, true

PR #306, opened 2019-11-20
add quotes

matthewdaltonamount:
###### Description

Quotes are needed


rvmiller89:
Thanks @matthewdaltonamount we noticed this as well from our automated tests, it should be fixed by https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/307

I only found the numeric fields explicitly needing quotes, alpha-numeric seems to be handled properly already.

PR #309, opened 2019-11-21
Autoupdate kube-prometheus mixin

jrobersonaquent:
###### Description

This PR adds autogenerating `deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet` based on `values.yaml`. 

A side effect of this change is it also makes the [`yq`](https://github.com/mikefarah/yq) tool available in the CI pipeline.

I also added [`jsonnet-bundler`](https://github.com/jsonnet-bundler/jsonnet-bundler) usage example to the doc, since that's how we're going to start using it internally if/when this gets merged.

###### Testing performed

- [ ] ci/build.sh


ci/jsonnet-mixin.tmpl
```diff
@@ -0,0 +1,19 @@
+{
```
jrobersonaquent: This file is the template that the `build.sh` uses to build the Jsonnet file. It just substitutes out `[/*REMOTE_WRITE*/]` with JSON-ified prometheus.prometheusSpec.remoteWrite.

I didn't know where to put this file since there doesn't appear to be anything like it, so let me know if it needs to be moved somewhere.

ci/build.sh
```diff
@@ -129,6 +129,16 @@ if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
...
+  echo "Copy prometheus.prometheusSpec.remoteWrite from 'prometheus-overrides.yaml' and inject into 'deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet'"
+  prometheus_remote_write=`yq r deploy/helm/prometheus-overrides.yaml prometheus.prometheusSpec.remoteWrite -j | jq '.' | sed 's/^/    /'`
+  # Escaping so sed will work
+  prometheus_remote_write="${prometheus_remote_write//\\/\\\\}"
+  prometheus_remote_write="${prometheus_remote_write//\//\\/}"
+  prometheus_remote_write="${prometheus_remote_write//&/\\&}"
+  prometheus_remote_write="${prometheus_remote_write//$'\n'/\\n}"
+  echo "// This file is autogenerated" > deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
+  sed 's#\[\/\*REMOTE_WRITE\*\/\]#'$prometheus_remote_write'#' ci/jsonnet-mixin.tmpl | sed 's#"http://collection-sumologic.sumologic.svc.cluster.local:9888\/#$._config.sumologicCollectorSvc + "#g' | sed 's/+:     /+: /' | sed -r 's/"(\w*)":/\1:/g' > deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
+
```
rvmiller89: I think we  need to also add `$(git diff deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet)` to the section commiting the generated files to git below (L152-L156)

jrobersonaquent: You're right, I'll add that.

jrobersonaquent: Fixed in 61ae61963550371f465824dcaef30ba48efaef54

PR #310, opened 2019-11-22
cleanup fluentd sumologic filter plugin

samjsong:
###### Description

Cherry picking filter plugin cleanup from #232 since this doesn't change behavior and will help make the coming changes easier

###### Testing performed

- [x] ci/build.sh


fluent-plugin-kubernetes-sumologic/lib/fluent/plugin/filter_kubernetes_sumologic.rb
```diff
@@ -158,23 +137,18 @@ def filter(tag, time, record)
...
-          sumo_metadata[:category] = sumo_metadata[:category] % k8s_metadata
-        else
-          sumo_metadata[:category] = (annotations["sumologic.com/sourceCategory"] % k8s_metadata).prepend(@source_category_prefix)
+        unless annotations["sumologic.com/sourceCategory"].nil?
+          sumo_metadata[:category] = annotations["sumologic.com/sourceCategory"].dup.prepend(@source_category_prefix)
         end
+        sumo_metadata[:host] = sumo_metadata[:host] % k8s_metadata
+        sumo_metadata[:source] = sumo_metadata[:source] % k8s_metadata
+        sumo_metadata[:category] = sumo_metadata[:category] % k8s_metadata
         sumo_metadata[:category].gsub!("-", @source_category_replace_dash)
```
rvmiller89: can this be removed due to L74

samjsong: I think the reason I left it is that `sumo_metadata[:category]` is modified above, so that if any of `k8s_metadata` includes a dash we replace it.

I believe this would be consistent with current behavior, but I'm not actually sure this is the behavior we want to enforce. For instance, if pod name is
```
kube-apiserver-ip-172-20-51-167.us-west-1.compute.internal
```
then the source category given our default of `sourceCategory: "%{namespace}/%{pod_name}"` would end up like this:
```
kubernetes/kube/system/kube/apiserver/ip/172/20/51/167.us/west
```
(as a side note, I'm not actually sure what happened to `-1.compute.internal`, will need to look into that)

For this PR I vote for keeping this behavior to be backwards compatible, but @frankreno can you confirm what behavior we want to enforce for `source_category_replace_dash`, and what the usecase is/was?

PR #312, opened 2019-11-25
Pin Travis CI Snap Helm version to 2.16

rvmiller89:
###### Description

This is needed because our CI doesn't currently work with Helm v3, it will fail with `Error: unknown flag: --client-only` (among other issues).

For now, pin to the last stable Helm v2 release channel.

###### Testing performed

- [x] ci/build.sh


rvmiller89:
Can we go ahead and get this in (even though it deletes the libsonnet file)? This is needed to unblock our CI to fix the libsonnet stuff

maimaisie:
Thanks Ryan

PR #313, opened 2019-11-25
Fix Multiline support for kubernetes collection

vsinghal13:
###### Description

In order to fix the multiline support the following config was added:
  1. Add a new regex parser for multiline detection 
  2. Switch on the Multiline feature of fluentbit for multiline log detection 
  3. Replace FluentD concat plugin with record_transformer plugin to process the multiline logs in proper format.

Multiline log : 
multiline message

Single line log: 
single line message

###### Testing performed

- [x] ci/build.sh


deploy/helm/sumologic/values.yaml
```diff
@@ -210,13 +203,20 @@ fluent-bit:
     systemd:
       enabled: true
 
+  parsers:
+    enabled: true
+    regex:
+      - name: multi_line
+        regex: (?<log>.*\d{4}-\d{1,2}-\d{1,2} \d{2}:\d{2}:\d{2}.*)
```
rvmiller89: can we add a comment above this with the format it will match, something like
```
# Matches "2019-11-17 07:14:12"
```

rvmiller89: Also how does this handle a timestamp with that format appearing somewhere in the middle of a multi-line message? Will it split it incorrectly?

vsinghal13: We can change the regex to ```(?<log>^{"log":"\d{4}-\d{1,2}-\d{1,2} \d{2}:\d{2}:\d{2}.*)``` to avoid that case.

rvmiller89:
Can you merge master as well, I think your branch hasn't picked up the libsonnet stuff

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -1,9 +1,12 @@
...
-  @type concat
-  key log
-  multiline_start_regexp "#{ENV['MULTILINE_START_REGEXP']}"
-  separator "#{ENV['CONCAT_SEPARATOR']}"
-  timeout_label @NORMAL
+  @type record_transformer
+  enable_ruby
+  renew_record true
+  <record>
+    log    ${record["log"].split(/[\n\t]+/).map! {|item| JSON.parse(item)["log"]}.join("")}
```
rvmiller89: just confirming: This won't cause issues if somehow the log coming from `containers.**` doesn't have `log`/`stream`/`time` keys?

vsinghal13: This is only expecting the `log` key and using that to generate the `stream` and `time` key value pairs from that, so it should be fine.

samjsong: But if the string contained in `log` doesn't have `stream` and `time`, i.e. if the user is not using docker, then this wouldn't work, correct? Maybe we should document somewhere that we only support docker for now?

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -507,9 +510,9 @@ spec:
         - name: KUBERNETES_META_REDUCE
           value: "false"
         - name: MULTILINE_START_REGEXP
-          value: "/^\\w{3} \\d{1,2}, \\d{4}/"
+          value:
```
samjsong: Since these aren't being used anymore, can you also remove these two from `deploy/helm/sumologic/templates/deployment.yaml` so that they're no longer rendered here?

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
vsinghal13: Actually it would work even if the `log` doesn't have `stream` and `time`keys.  Its just that if the key is not present, the JSON.parse value will be empty string for the key.

deploy/helm/sumologic/values.yaml
```diff
@@ -210,13 +203,21 @@ fluent-bit:
     systemd:
       enabled: true
 
+  parsers:
+    enabled: true
+    # This regex matches the first line of a multiline log starting with a date of the format : "2019-11-17 07:14:12"
+    regex:
+      - name: multi_line
+        regex: (?<log>^{"log":"\d{4}-\d{1,2}-\d{1,2} \d{2}:\d{2}:\d{2}.*)
+
```
prateekladha: @vsinghal13 are we using these values somewhere else? because we have defined same thing in `deploy/helm/fluent-bit-overrides.yaml`

vsinghal13: The `fluent-bit-overrides.yaml` is an auto-generated file for all the fluent-bit values that have been overridden in `values.yaml` . This is done for non-helm users.

PR #314, opened 2019-11-25
Revert "Autoupdate kube-prometheus mixin"

rvmiller89:
Reverts SumoLogic/sumologic-kubernetes-collection#309

Causes issue where the libsonnet file actually gets deleted. I think it has something to do with the sed replace not working on our Travis CI, but I haven't looked into further:
```
Copy prometheus.prometheusSpec.remoteWrite from 'prometheus-overrides.yaml' and inject into 'deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet'
sed: -e expression #1, char 27: unterminated `s' command
```

CC: @jrobersonaquent


jrobersonaquent:
Interesting. Do you know if you're using GNU `sed` or something else like BSD's `sed`?

jrobersonaquent:
Should be fixed in #315

PR #317, opened 2019-11-25
Update non-helm installation docs

maimaisie:
###### Description

Since we are deprecating the setup.sh and instead providing a YAML file to run a container that does all setup, we need to update our docs to reflect that.

###### Testing performed

- [x] ci/build.sh
- [x] Test setup command and verify setup job is completed after a minute or two


maimaisie:
I'm wondering if the conflicting `setup-sumologic.yaml.tmpl` is because I branched off from a fair old master. But I don't think I can manually fix the conflict since the CI will fail as soon as it detects the change is from me.

maimaisie:
I am closing this PR and re-opening another one from a newer master branch.

PR #319, opened 2019-11-26
Increase fluent-bit flush interval to lessen DNS load

rvmiller89:
###### Description

From the CoreDNS investigation, we are using the default fluent-bit flush interval of 1s, which leads to DNS lookup per flush. We can increase the flush interval to lessen the DNS lookups, and also the number of network calls from fluent-bit to fluentd.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -1,4 +1,6 @@
 # This file is auto-generated.
+service:
+  flush: 5
 metrics:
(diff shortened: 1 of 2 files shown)


lei-sumo:
I think we decreased it to 1s to avoid aother issue? @samjsong Did you remember?

samjsong:
I believe that was the `Refresh_Interval` rather than flush interval. I think this is fine to change

PR #321, opened 2019-11-26
Use empty endpoint URL by default for redirection

maimaisie:
###### Description

We decided to use empty string for the `endpoint` config param so by default it will do auto URL redirection unless the user overrides it. This will work for both helm and non-helm installations. 

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/kubernetes/setup-sumologic.yaml.tmpl
+++ b/deploy/kubernetes/setup-sumologic.yaml.tmpl
@@ -269,7 +269,7 @@ spec:
           value: $SUMOLOGIC_ACCESSID
         - name: SUMOLOGIC_ACCESSKEY
           value: $SUMOLOGIC_ACCESSKEY
-        # - name: SUMOLOGIC_BASE_URL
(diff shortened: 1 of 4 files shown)


deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -269,7 +269,7 @@ spec:
           value: $SUMOLOGIC_ACCESSID
         - name: SUMOLOGIC_ACCESSKEY
           value: $SUMOLOGIC_ACCESSKEY
-        # - name: SUMOLOGIC_BASE_URL
-        #  value: $SUMOLOGIC_BASE_URL
+        - name: SUMOLOGIC_BASE_URL
+          value: $SUMOLOGIC_BASE_URL
```
rvmiller89: this doesn't look right. We need to remove `--set sumologic.endpoint="\$SUMOLOGIC_BASE_URL"` from `build.sh`

PR #322, opened 2019-11-26
FluentD file buffer

vsinghal13:
###### Description

Add option to use fluentD file buffer instead of in-memory buffer. Default will be `memory` . 
Later on, we might add the functionality to write to an external volume, instead of local file system.

###### Testing performed

- [x] ci/build.sh


deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -270,6 +270,6 @@ spec:
         - name: SUMOLOGIC_ACCESSKEY
           value: $SUMOLOGIC_ACCESSKEY
         # - name: SUMOLOGIC_BASE_URL
```
rvmiller89: can you rebase master? this looks wrong

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -115,7 +152,7 @@ data:
     </source>
     @include logs.source.containers.conf
     @include logs.source.systemd.conf
-  logs.output.conf: |
+  logs.output.conf: |+
```
rvmiller89: should `|+` be `|-` ?

vsinghal13: yeah, even I noticed that. Upon searching found that its because of an extra trailing blank line. ``` |+: "keep": keep the line feed, keep trailing blank lines.```
Will remove that.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -57,55 +54,95 @@ data:
...
         @include metrics.output.conf
+        <buffer>
+          @type file
+          path /fluentd/buffer/metrics.all
+          @include buffer.output.conf
+        </buffer>
       </match>
     </label>
-  metrics.output.conf: |-
+  metrics.output.conf: |
```
rvmiller89: change back to `|-`

deploy/helm/sumologic/conf/metrics/metrics.conf
```diff
@@ -22,47 +22,120 @@
     @id sumologic.endpoint.metrics.apiserver
     endpoint "#{ENV['SUMO_ENDPOINT_METRICS_APISERVER']}"
     @include metrics.output.conf
+    <buffer>
+      {{- if eq .Values.sumologic.fluentdBuffer "file" }}
+        @type file
```
rvmiller89: indent looks off on these, e.g. https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/322/files#diff-77aa48b7684fe90d242ce9fdae9da300R58

I think they need to be something like
```
      {{- if eq .Values.sumologic.fluentdBuffer "file" }}
      @type file
      path /fluentd/buffer/metrics.kubelet
      {{- else }}
      @type memory
      {{- end }}
```

deploy/helm/sumologic/conf/logs/logs.source.systemd.conf
```diff
@@ -17,6 +17,15 @@
     @type sumologic
     @id sumologic.endpoint.logs.kubelet
     @include logs.output.conf
+    <buffer>
+      {{- if eq .Values.sumologic.fluentdBuffer "file" }}
+        @type file
```
rvmiller89: missed indent here?

deploy/helm/sumologic/conf/metrics/metrics.conf
vsinghal13: good catch! Changed the indentation.

deploy/helm/sumologic/values.yaml
```diff
@@ -159,6 +159,9 @@ sumologic:
   # watchResourceEventsOverrides:
   #   pods: "v1"
   #   events: "events.k8s.io/v1beta1"
+  
+  ## Option to specify the fluentD buffer as file/memory.
```
rvmiller89: nit: "fluentD" --> "Fluentd" to match our docs

rvmiller89: also can we make this
```
fluentd:
  buffer: "memory"
```

vsinghal13: "FluentD" seems to be used in the `values.yaml`. Should update all those as well?

lei-sumo:
Is this PR sufficient to be released separately? Or do we need to waiting external volume? Also is there a doc/spec for the expected behavior (flush at shutdown, etc)?

vsinghal13:
> Is this PR sufficient to be released separately? Or do we need to waiting external volume? Also is there a doc/spec for the expected behavior (flush at shutdown, etc)?

I will be creating a separate PR for the doc/spec. This PR will use the local file system to store the buffer files. In future, we can provide additional option to use external volumes to store the buffer.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -57,55 +54,87 @@ data:
         @id sumologic.endpoint.metrics.apiserver
         endpoint "#{ENV['SUMO_ENDPOINT_METRICS_APISERVER']}"
         @include metrics.output.conf
+        <buffer>
+          @type memory
```
prateekladha: @vsinghal13 is this change is only for helm based installation? For non-helm, we are still going to buffer to memory?

deploy/helm/sumologic/conf/events/events.conf
```diff
@@ -30,12 +30,13 @@
   verify_ssl "#{ENV['VERIFY_SSL']}"
   proxy_uri "#{ENV['PROXY_URI']}"
   <buffer>
+    {{- if eq .Values.sumologic.fluentd.buffer "file" }}
+    @type file
+    path /fluentd/buffer/events
```
prateekladha: @vsinghal13 is this path local to node? If yes, all pods inside that node will store buffer in the same file?

prateekladha: Screenshot 2019-12-18 at 8 48 25 PM

vsinghal13: Yes, the path is local to node. Each buffer section needs to have a unique path. The pods will store will store in different paths based on the config path provided in the respective buffer section.

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
vsinghal13: I think the non-helm users can customize config by following these steps: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/Non_Helm_Installation.md#customize-configuration

PR #329, opened 2019-12-03
Add doc for Advanced Configuration / Best Practices

vsinghal13:
###### Description

Add doc for Advanced Configuration / Best Practices to cover information for file based buffer and multiline support.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/docs/Best_Practices.md
+++ b/deploy/docs/Best_Practices.md
@@ -85,4 +85,10 @@ retry_timeout 2h
 retry_exponential_backoff_base 2
 ```
 
+Once the config has been modified in the `values.yaml` file,  we need to run the `helm upgrade` in order to apply the changes.
+
+```bash
+$ helm upgrade collection sumologic/sumologic --reuse-values -f values.yaml
+```
+
 Reference link to official Fluentd buffer documentation: https://docs.fluentd.org/configuration/buffer-section
--- a/deploy/docs/Best_Practices.md
+++ b/deploy/docs/Best_Practices.md
@@ -1,0 +1,68 @@
+
+# Advanced Configuration / Best Practices
+
+
+### Multiline Log Support
+
+By default, we use a regex that matches the first line of a multiline log starting with a date of the format : `2019-11-17 07:14:12`.
+
+Users can specify their custom regex to detect first line of a multiline log and parse the logs correctly.
+
+New parsers can be defined under the `parsers` key of the fluent-bit configuration section in `values.yaml` as follows:
+
+```
+parsers:
+  enabled: true
+  regex:
+    - name: new_parser_name
+    regex: (?<log>^{"log":"\d{4}-\d{1,2}-\d{1,2} \d{2}:\d{2}:\d{2}.*)
+```
+
+The regex needs to have at least one named capture group.
+
+In order to use the newly define parser, change the `Parser_Firstline` parameter in the `Input plugin` configuration of fluent-bit:
+
+```bash
+Parser_Firstline new_parser_name
+```
+
+
+### Fluentd File-based buffer
+
+By default, we use the in-memory buffer for the Fluentd buffer, but for production environments we recommend users to use the file-based buffer instead.
+
+Buffer configuration can be set in `values.yaml` under the `sumologic` key as follows:
+
+```
+fluentd:
+## Option to specify the Fluentd buffer as file/memory.
+   buffer: "file"
+```
+
+Additional buffering and flushing parameters can be added in the `buffer.output.conf`config file, to be used by all buffer sections.
+
+We recommend the following configurations for the file-based buffer:
+
+#### Buffering parameters:
+```
+flush_interval 5s
+
+flush_thread_count 4
+
+chunk_limit_size 100k
+
+total_limit_size 128m
+```
+
+####  Flushing parameters:
+```
+retry_wait 5
+
+retry_forever false
+
+retry_timeout 2h
+
+retry_exponential_backoff_base 2
+```
+
+Reference link to official Fluentd buffer documentation: https://docs.fluentd.org/configuration/buffer-section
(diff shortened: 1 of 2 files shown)


rvmiller89:
David's on PTO this week, should we add Marie?

deploy/docs/Best_Practices.md
```diff
@@ -0,0 +1,68 @@
+
+# Advanced Configuration / Best Practices
+
+
+### Multiline Log Support
+
+By default, we use a regex that matches the first line of a multiline log starting with a date of the format : `2019-11-17 07:14:12`.
+
+Users can specify their custom regex to detect first line of a multiline log and parse the logs correctly.
```
rvmiller89: "to detect first line" --> "to detect the first line"

deploy/docs/Best_Practices.md
```diff
@@ -0,0 +1,68 @@
...
+
+New parsers can be defined under the `parsers` key of the fluent-bit configuration section in `values.yaml` as follows:
+
+```
+parsers:
+  enabled: true
+  regex:
+    - name: new_parser_name
```
rvmiller89: in the example can we keep the current `multi_line` parser (since the customer will likely want to keep this parser and *append* their parser).  So the example should show 2 parsers with the 2nd one being the newly added example.  We should also try to come up with a regex example different from the current one

deploy/docs/Best_Practices.md
```diff
@@ -0,0 +1,68 @@
...
+    regex: (?<log>^{"log":"\d{4}-\d{1,2}-\d{1,2} \d{2}:\d{2}:\d{2}.*)
+```
+
+The regex needs to have at least one named capture group.
+
+In order to use the newly define parser, change the `Parser_Firstline` parameter in the `Input plugin` configuration of fluent-bit:
+
+```bash
+Parser_Firstline new_parser_name
```
rvmiller89: same here, let's show how they could *append* their new parser while keeping the existing config

deploy/docs/Best_Practices.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+### Fluentd File-based buffer
+
+By default, we use the in-memory buffer for the Fluentd buffer, but for production environments we recommend users to use the file-based buffer instead.
+
+Buffer configuration can be set in `values.yaml` under the `sumologic` key as follows:
+
+```
+fluentd:
```
rvmiller89: I guess this is in preparation for merging https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/330

deploy/docs/Best_Practices.md
```diff
@@ -0,0 +1,68 @@
...
+chunk_limit_size 100k
+
+total_limit_size 128m
+```
+
+####  Flushing parameters:
+```
+retry_wait 5
+
+retry_forever false
```
rvmiller89: we recommend dropping data? for production environments this seems like a bad idea.  We should recommend sizing the volume appropriately to not drop data.

vsinghal13: For  `retry_forever, If `true`, plugin will ignore `retry_timeout` and `retry_max_times` options and retry flushing forever. 

There is another buffering parameter which defines how output plugin behaves when its buffer queue is full.
Default : throw_exception
`overflow_action` [enum:
throw_exception/block/drop_oldest_chunk]

deploy/docs/Best_Practices.md
```diff
@@ -0,0 +1,68 @@
...
+
+Buffer configuration can be set in `values.yaml` under the `sumologic` key as follows:
+
+```
+fluentd:
+## Option to specify the Fluentd buffer as file/memory.
+   buffer: "file"
+```
+
+Additional buffering and flushing parameters can be added in the `buffer.output.conf`config file, to be used by all buffer sections.
```
samjsong: Correct me if I'm wrong, but I believe changing these files will not do anything since the files will have already been loaded into the configmap data. I think currently the only option users currently have is to manually update the configmap, but that won't persist on `helm upgrade`. Once we merge #330 they'll be able to edit `values.yaml` to change or add parameters, and we'll have to update the docs accordingly

vsinghal13: Updated the doc to reflect the change in #330 for editing the `values.yaml`

rvmiller89: +1 similar to the steps here: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/Troubleshoot_Collection.md#fluentd-logs

For fluent-bit daemonset they could perform a rolling update? https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/

vsinghal13:
> LGTM, though I do think we need to add a section on how to actually apply changes made to Fluent-bit daemonset

The changes are in `values.yaml`, so will be applied via the `helm upgrade` command. Added a section to run that post making config changes in `values.yaml`

rvmiller89:
I thought when I was testing making changes to fluent-bit configuration, even with `hem upgrade` the changes weren't being picked up by fluent-bit, and I needed to actually redeploy the daemon set in order for the newly launched fluent-bit to pick up the changes. Can you confirm?

vsinghal13:
> I thought when I was testing making changes to fluent-bit configuration, even with `helm upgrade` the changes weren't being picked up by fluent-bit, and I needed to actually redeploy the daemon set in order for the newly launched fluent-bit to pick up the changes. Can you confirm?

When I tried the changes were picked up and they were redeployed in rolling update by the helm upgrade itself.

vsinghal13:
Yes, I was planning to merge only after #330

vsinghal13:
Closing this PR as most of the stuff is already there in the existing doc. Will create a new PR will the specific changes.

PR #330, opened 2019-12-03
[breaking change] Expose fluentd config without ENV vars - Take 2

samjsong:
###### Description

The goal was to expose the Fluentd config to the user, while hopefully also making the `values.yaml` easier to use. We also want to get rid of the layer of indirection that was created with the ENV vars.

`values.yaml` will be the most interesting file to look at, hopefully the rest of it is just inserting the mustache syntax in the correct places to reflect what is in `values.yaml`.

Note that we also removed several configs in the process, including `addStream`, `addTime`, and `addTimestamp`, which if desired could be done through use of a `record_modifier` plugin. The relevant code was also removed from the `kubernetes_sumologic` filter plugin code.

Reading commit by commit might be easier, I tried to break it up into reasonable chunks but sometimes I made mistakes or missed things that I fixed in later commits so it's not perfect... sorry 😞 

###### Testing performed

- [x] ci/build.sh
- [x] Redeploy fluentd and fluentd-events pods
- [x] Confirm events, logs, and metrics are coming in

PR #333, opened 2019-12-09
add CHANGELOG, include tentative 1.0.0 changes

samjsong:
###### Description

To be merged after the 1.0.0 release is cut (after #330 and #329 are merged)

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,49 @@
+# Change Log
+
+## 1.0.0
+- Expose new configs in `values.yaml` to enable users to configure Fluentd to their liking. See the new `fluentd` section in `values.yaml` for more details.
+- Make several changes to `values.yaml` for consistency and understandability:
+
+Old Config | New Config
+|:-------- |:--------
+`sumologic.eventCollectionEnabled` | `fluentd.events.enabled`
+`sumologic.logFormat` | `fluentd.logs.output.logFormat`
+`sumologic.flushInterval` | `fluentd.buffer.flushInterval`
+`sumologic.numThreads` | `fluentd.buffer.numThreads`
+`sumologic.chunkLimitSize` | `fluentd.buffer.chunkLimitSize`
+`sumologic.totalLimitSize` | `fluentd.buffer.totalLimitSize`
+`sumologic.sourceName` | `fluentd.logs.containers.sourceName`
+`sumologic.sourceCategory` | `fluentd.logs.containers.sourceCategory`
+`sumologic.sourceCategoryPrefix` | `fluentd.logs.containers.sourceCategoryPrefix`
+`sumologic.sourceCategoryReplaceDash` | `fluentd.logs.containers.sourceCategoryReplaceDash`
+`sumologic.verifySsl` | `fluentd.verifySsl`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
+# Change Log
+
+## 1.0.0
```
rvmiller89: although we don't prefix images with `v`, we do prefix the releases, so I think this should be `v1.0.0` for consistency

CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
...
+`sumologic.excludeNamespaceRegex` | `fluentd.logs.containers.excludeNamespaceRegex`
+`sumologic.excludePodRegex` | `fluentd.logs.containers.excludePodRegex`
+`sumologic.fluentdLogLevel` | `fluentd.logLevel`
+`sumologic.watchResourceEventsOverrides` | `fluentd.events.watchResourceEventsOverrides`
+`sumologic.k8sMetadataFilter` | `fluentd.logs.containers.k8sMetadataFilter`
+
+- Remove some outdated configs in `values.yaml`
+  - sumologic.kubernetesMeta
+  - sumologic.kubernetesMetaReduce
+  > To preserve the behavior of `kubernetesMeta` or `kubernetesMetaReduce`, we recommend using the `fields` log_format, which by default strips this metadata from the body of the log message.
```
rvmiller89: kinda weird this shows up as grey-ed out text. Could these just be bullet points with `-` ?

CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
...
+  ```
+  fluentd:
+    logs:
+      containers:
+        extraFilterPluginConf:
+          <filter **>
+            @type record_modifier
+            remove_keys stream,time
+            <record>
+              timestamp ${(Time.now.to_f*1000).to_i}
```
rvmiller89: 1. I thought we still have `timestamp` ? or we definitely got rid of it now?
2. Why use `Time.now` and not the `time` key? During the hackathon I was able to get this to work just doing
```
<record>
  timestamp ${time.to_i}
</record>
```

samjsong: 1. I wanted to take the opportunity of a breaking change to get rid of `timestamp`, since it's not consistent with the timestamp of the log anyway.
2. Great point, will make this change

rvmiller89: I would prefer we wait until we have the protobuf message format to properly pass timestamp before we get rid of `timestamp`.  The rationale is that we have more success parsing (and deciding to parse) the epoch timestamp vs. learning diverse timestamp formats from multiplexed timestamp formats appearing in the same HTTP source (even in the same HTTP payload).

samjsong: That's fair, in that case I will go ahead and make the relevant changes in #330 and update this changelog accordingly

samjsong:
I wanted to get some reviews from our team first to iron out the details before requesting review from @drduke1 since he's back 🙂

samjsong:
The changes in #330 have been already been completed 👍

samjsong:
closing with https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/630

PR #288, opened 2019-11-13
Revert pre-upgrade hook for now for further testing

rvmiller89:
###### Description

Fill in your description here.

###### Testing performed

- [ ] ci/build.sh
- [ ] Redeploy fluentd and fluentd-events pods
- [ ] Confirm events, logs, and metrics are coming in

PR #335, opened 2019-12-11
[breaking change] Add pre-upgrade hook for setup

maimaisie:
###### Description

Revert #288 since we have confirmed that upgrading from non-TF chart to TF chart with setup running in pre-upgrade will potentially create a new collector (breaking change). 

Will update CHANGELOG in #333.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -56,27 +56,27 @@ sumologic:
   setup:
     clusterRole:
       annotations:
-        helm.sh/hook: pre-install
+        helm.sh/hook: pre-install,pre-upgrade
         helm.sh/hook-weight: "1"
         helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
     clusterRoleBinding:
       annotations:
-        helm.sh/hook: pre-install
+        helm.sh/hook: pre-install,pre-upgrade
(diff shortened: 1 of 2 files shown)


rvmiller89:
Will this cause issues until we get a fix for https://jira.kumoroku.com/jira/browse/SUMO-125138 ?

If their collector name has a dot in it, this would run the job during pre-upgrade, which would fail.

maimaisie:
That's a great point, I'm afraid it would. Do we know the QE timeline? We might still be able to get this in if the URL fix makes to production next week.

rvmiller89:
Just merged the fix for that, so we just want to upgrade api with that fix before we do the major release with this.

rvmiller89:
To be safer, could we delay merging this until we know we want to add a new source? I'm concerned there could be other Terraform gotchas that will take some time to discover.

maimaisie:
@rvmiller89  I missed your last comment. I was going to merge this PR today, I think it should be fine to merge now since QE will be testing the upgrade flow. What do you think?

rvmiller89:
I guess if it's already in the QE test plan I'm fine with merging. If customers face issues upgrading we can tell them to remove `,pre-upgrade` from their values.yaml

maimaisie:
Sounds good, thanks @rvmiller89

PR #336, opened 2019-12-11
Fix ambiguity in helm install command

samjsong:
###### Description

###### Testing performed

- [ ] ci/build.sh


rvmiller89:
What's the issue with the quotes?

samjsong:
Synced offline, there should be nothing wrong with the quotes and in fact they are necessary if there is a space in the cluster/collector name.

PR #337, opened 2019-12-11
Add support for data persistence

maimaisie:
###### Description

This PR adds support for data persistence to a persistent volume per fluentd pod. After discussing with the team, we decided to switch from using deployments to statefulsets because statefulsets allows easier setup of persistent volume.

###### Testing performed

- [x] ci/build.sh
- [x] Throttle the account, confirm data is written to buffer and flushed after unthrottling
- [ ] ~~Confirm default behavior when buffer is full~~ QE item


deploy/helm/sumologic/values.yaml
```diff
@@ -165,6 +165,24 @@ sumologic:
     ## Option to specify the Fluentd buffer as file/memory.
     buffer: "memory"
 
+    securityContext:
+      fsGroup: 999
```
maimaisie: FYI: This is the fluentd user id. By default pvc is always mounted as root user so without setting the security context the fluentd user does not have permission to write to the mounted volume. I spent a long time trying to figure out why the fluentd chart does not need to set this. The fluentd dockerfile [link](https://github.com/fluent/fluentd-docker-image/blob/master/v1.6/debian/Dockerfile#L49) says that everything under the `/fluentd` folder should be owned by the fluentd user, but it doesn't work in our case even though our image is built on top of fluentd's. 

I looked at the helm stable repo and saw a lot of charts setting the security context fsGroup, it looks like it's very common so I'm following this pattern.

deploy/helm/sumologic/templates/deployment.yaml
```diff
@@ -33,6 +33,8 @@ spec:
       - name: config-volume
         configMap:
           name: {{ template "sumologic.fullname" . }}
+      securityContext:
+        fsGroup: {{ .Values.sumologic.fluentd.securityContext.fsGroup }}
```
rvmiller89: heads up @samjsong , this needs to be updated in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/330

deploy/helm/sumologic/templates/deployment.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
```
rvmiller89: should we rename these files (`deployment.yaml` and `events-deployment.yaml`) ? Are other templates calling these `statefulset.yaml` or something?

deploy/helm/sumologic/values.yaml
rvmiller89: should we add a `// NOTE` as well explaining `999` ?

deploy/helm/sumologic/values.yaml
```diff
@@ -165,6 +165,24 @@ sumologic:
...
+
+    ## Persist data to a persistent volume; Must be used together with Fluentd file buffer (fluentd.buffer="file")
```
rvmiller89: do we want to require them to set two values, or let this one take **precedence** over the other one? I could go either way, what are other charts doing?

deploy/helm/sumologic/templates/deployment.yaml
maimaisie: yes we should

deploy/helm/sumologic/templates/deployment.yaml
maimaisie: there's quite some changes after renaming `deployment` to `statefulset` in different places, @samjsong maybe I can work on them after this PR is merged and contribute to your PR for any `values.yaml` changes needed to avoid more conflicts

deploy/helm/sumologic/values.yaml
maimaisie: I think you are right, having this one takes precedence makes sense. I don't see any charts using two values to enable persistence

deploy/helm/sumologic/templates/deployment.yaml
samjsong: that would be great, thanks @maimaisie ! Looks like this PR and #339 should be in a minor release before #330 so that should work

maimaisie:
I brought up an issue a few days ago during standup where the PVCs and PVs do not get deleted after `helm delete --purge`, it turns out this is by design in helm because they are provisioned by the underlying infrastructure (AWS, GCP etc) and not by helm. More details in https://github.com/helm/helm/issues/3313
We probably just need to add documentations to manually delete these resources if this is what the users intend to do.

maimaisie:
The commits in this branch are kinda messed up right now, I'll open a new PR

PR #339, opened 2019-12-12
Add HorizontalPodAutoscaler for fluentd

vsinghal13:
###### Description

 This PR add the capability to autoscale fluentd based on CPU usage. Metrics Server is used to expose the metrics API . HPA uses the API to read the metrics and make a decision to autoscale.

Below is a graph to show the autoscaler behavior. 

Two tests scenarios: 
1. Gradual increase in load: The load was increased in increments of 2 from 0  to 6 and then dropped to 2 and eventually 0. 
2. Immediate spike: Load goes from 0 to 10 in a short span of time.

In the first scenario, the autoscaler scales the number of replicas in multiples of 2 i.e. 1,2,4,8.
When the the load is dropped down to 0, autoscaler gradually drops with a much smaller step size and stabilizes at 3.

In the second scenario, the autoscaler again scales from 3 to 6 and then 10.
 
Screen Shot 2019-12-12 at 11 52 32 PM

Default value set to be "true". 
###### Testing performed

- [x] ci/build.sh


deploy/helm/sumologic/values.yaml
```diff
@@ -164,6 +164,12 @@ sumologic:
   fluentd:
     ## Option to specify the Fluentd buffer as file/memory.
     buffer: "memory"
+    ## Option to turn autoscaling on for fluentd and specify metrics for HPA.
+    autoscaling:
+      enabled: true
+      minReplicas: 1
```
Gourav2906: I think min should be 3

deploy/helm/sumologic/templates/hpa.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.sumologic.fluentd.autoscaling.enabled}}
```
rvmiller89: what's the reason for `and` here? We're not checking two conditions.

deploy/helm/sumologic/values.yaml
rvmiller89: do we want to set `minReplicas` to `3` to match our current 3 replicas?

deploy/helm/sumologic/values.yaml
```diff
@@ -181,6 +187,15 @@ sumologic:
     ## ref: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#configuration
     bearerCacheTtl: "3600"
 
+
+## Configure metrics-server
+## ref: https://github.com/helm/charts/blob/master/stable/metrics-server/values.yaml
+metrics-server:
```
rvmiller89: do we need to teach our CI to generate `metrics-server-overrides.yaml` for the non-Helm installation?

deploy/helm/sumologic/values.yaml
```diff
@@ -181,6 +187,15 @@ sumologic:
...
+  enabled: true
+  args:
+    - --kubelet-insecure-tls
```
rvmiller89: I could see customers having an issue with using this in an insecure way.  Is there any other way to run this?

deploy/helm/sumologic/templates/hpa.yaml
vsinghal13: Good catch! Will fix it.

deploy/helm/sumologic/values.yaml
vsinghal13: Is was wondering if that's necessary? I was thinking that we keep the min to be 1 and let the autoscaler do its job.

rvmiller89: Yeah, it's worth a discussion. Should get @frankreno 's opinion as well.  Would 1 replica no longer be considered high availability (HA) since there could be data loss until the next replica is brought up?

deploy/helm/sumologic/values.yaml
vsinghal13: Without using that the metrics server is unable to get metrics.  Will look into it more if we can avoid this.
https://github.com/kubernetes-sigs/metrics-server/issues/131

deploy/helm/sumologic/values.yaml
frankreno: I would suggest we have a min of 2 or 3 at least to prevent any loss of data while its spinning up or if it did scale down to 1 and the node died, it would have a fall back.  This can be our default and we just expose it they values.yaml so customers can change if needed.

vsinghal13: ok will change it to 3 then.

deploy/helm/sumologic/values.yaml
vsinghal13: I guess it will be required if we want to add the arguments.

deploy/helm/sumologic/values.yaml
```diff
@@ -164,6 +164,12 @@ sumologic:
   fluentd:
     ## Option to specify the Fluentd buffer as file/memory.
     buffer: "memory"
+    ## Option to turn autoscaling on for fluentd and specify metrics for HPA.
+    autoscaling:
+      enabled: false
```
prateekladha: @vsinghal13 will autoscaling be OFF by default?

vsinghal13: yes, by default it will be off as of now.

Issue #268, opened 2019-11-01
Prometheus-Operator(Charts) comes with different record rules for K8S >= 1.14 which renders some dashboards inoperable


djsly:
For Example

When using K8S >= 1.14, the prometheus-operator charts only installs the following rules

https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/rules-1.14/node.rules.yaml

compared to 

https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/rules/node.rules.yaml

that is 27 missing rules just for the node...

The following remoteWrite entry contains lots of deprecated recordNames that are used in the different Kubernetes App Dashboards.

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L355

Note that AKS is about to support N-2 starting in December and that 1.14/1.15 will be 2 of the three supported versions.

https://github.com/Azure/AKS/issues/1235

vsinghal13:
Closing this as it is resolved.

PR #341, opened 2019-12-16
Bump Falco to 1.0.11 and add recording rules removed for K8s >=1.14.0

frankreno:
###### Description

This PR bumps falco to 1.0.11 (tested on non managed k8s 1.16, EKS 1.14).  It also adds a new template to add in recording rules that will only be rendered if the k8s version is >=1.14.0.  This is a short term fix to fix missing metrics that the K8s apps depend on while we plan the long term fix of supporting mixins.

###### Testing performed

- [ ] ci/build.sh


samjsong:
I was under the impression that these additional recording rules were only required for EKS >= 1.14.0, is adding these rules safe for non-EKS users?

deploy/helm/sumologic/templates/prometheus-rules.yaml
```diff
@@ -0,0 +1,189 @@
+{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
+{{- if semverCompare ">=1.14.0-0" $kubeTargetVersion }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
```
rvmiller89: are we not supporting yaml based install for these? (this template is Helm-only).

frankreno:
@samjsong no it is a general k8s issue with 1.14, not specific to anything other than the version of k8s.  See #268

deploy/helm/sumologic/templates/prometheus-rules.yaml
frankreno: ahh good point, let me add that.

rvmiller89: actually I might be confused, I see it's added as a section to the generated `fluentd-sumologic.yaml.tmpl`

djsly:
LGTM, thanks for adding that as part of your official chart.  Once merged, I will be upgrading my colleciton solution

frankreno:
@samjsong @rvmiller89 - found a better way to do this.  Single prop change in the values.yaml can force the pre 1.14 rules to render. Mind reviewing when you get a chance so we can get this merged?

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -656,3 +656,192 @@ spec:
           value: "100k"
         - name: TOTAL_LIMIT_SIZE
           value: "128m"
+---
+# Source: sumologic/templates/prometheus-rules.yaml
```
rvmiller89: Huh, I think you'll need to open a new PR :( CI doesn't seem smart enough to remove this

frankreno: Boo.  will do

frankreno:
Opened #344

PR #343, opened 2019-12-19
Improve fluentd liveness probe

vsinghal13:
###### Description

Add new liveness probe to fluentd deployment. 
```
httpGet:
    path: /fluentd.pod.healthcheck?json=%7B%22log%22%3A+%22health+check%22%7D
    port: 9880
```
The rationale is that if Fluentd can accept log messages, it must be healthy.
The endpoint itself results in a new fluentd tag `fluentd.pod-healthcheck`
The query parameter in the URL defines a URL-encoded JSON object that looks like this:
{"log": "health check"}

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/logs/logs.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.conf
@@ -3,10 +3,5 @@
   port 24321
   bind 0.0.0.0
 </source>
-<source>
-  @type http
-  port 9880
-  bind 0.0.0.0
-</source>
 @include logs.source.containers.conf
 @include logs.source.systemd.conf
--- a/deploy/helm/sumologic/conf/logs/logs.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.conf
@@ -3,5 +3,10 @@
   port 24321
   bind 0.0.0.0
 </source>
+<source>
+  @type http
+  port 9880
+  bind 0.0.0.0
+</source>
 @include logs.source.containers.conf
 @include logs.source.systemd.conf
--- a/deploy/helm/sumologic/templates/deployment.yaml
+++ b/deploy/helm/sumologic/templates/deployment.yaml
@@ -54,11 +54,9 @@ spec:
           periodSeconds: 30
           timeoutSeconds: 3
         readinessProbe:
-          exec:
-            command:
-            - "/bin/sh"
-            - "-c"
-            - "[ $(pgrep ruby | wc -l) -gt 0 ]"
+          httpGet:
+            path: /fluentd.pod.healthcheck?json=%7B%22log%22%3A+%22health+check%22%7D
+            port: 9880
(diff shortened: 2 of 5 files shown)


deploy/helm/sumologic/conf/logs/logs.conf
```diff
@@ -3,5 +3,10 @@
   port 24321
   bind 0.0.0.0
 </source>
+<source>
```
rvmiller89: I'm confused why only `logs.conf` gets this liveness probe. Should this be moved to `common.conf` ?

deploy/helm/sumologic/templates/deployment.yaml
```diff
@@ -47,13 +47,12 @@ spec:
...
         livenessProbe:
-          exec:
-            command:
-            - "/bin/sh"
-            - "-c"
-            - "[ $(pgrep ruby | wc -l) -gt 0 ]"
+          httpGet:
+            path: /fluentd.pod.healthcheck?json=%7B%22log%22%3A+%22health+check%22%7D
+            port: 9880
           initialDelaySeconds: 300
```
rvmiller89: this seems very generous, do we need to wait 5 minutes before starting the liveness probe?

deploy/helm/sumologic/templates/deployment.yaml
```diff
@@ -47,13 +47,12 @@ spec:
...
-          periodSeconds: 20
+          periodSeconds: 30
+          timeoutSeconds: 3
         readinessProbe:
```
rvmiller89: do we want to change the `readinessProbe` as well?

deploy/helm/sumologic/templates/deployment.yaml
vsinghal13: Currently we do have wait of 5 minutes, so I kept that. It is recommended to be generous in that otherwise the pod might end up in infinite loop of being restarted during startup.

deploy/helm/sumologic/templates/deployment.yaml
vsinghal13: Yes, we can change the readinessProbe to be the same as well.

deploy/helm/sumologic/templates/deployment.yaml
samjsong: 5min does sound a little high, but the readiness probe will fail first and the pod will stop accepting data, and if we need more replicas in between that time then with autoscaler enabled I think we'll be okay. So I'm okay with leaving it at 5min

deploy/helm/sumologic/templates/deployment.yaml
rvmiller89: did you want to make this change

rvmiller89:
+1 for doing on events deployment, missed that part.

PR #344, opened 2019-12-19
1.14+-fixes-support

frankreno:
###### Description

This PR bumps falco to 1.0.11 (tested on non managed k8s 1.16, EKS 1.14).It also ensures that we always render the pre 1.14 rules that our apps depend on. This is a short term fix to fix missing metrics that the K8s apps depend on while we plan the long term fix of supporting mixins.

###### Testing performed

- [ ] ci/build.sh


deploy/helm/metrics-server-overrides.yaml
```diff
@@ -0,0 +1,183 @@
+# This file is auto-generated.
```
rvmiller89: @vsinghal13  was this missed before? Should be OK to check in with this unrelated PR though?

rvmiller89:
Huh, there are conflicts for `deploy/helm/metrics-server-overrides.yaml`.  @vsinghal13 can you work with Frank to resolve?

deploy/helm/metrics-server-overrides.yaml
vsinghal13: The file was added in the autoscaler PR.

vsinghal13: In this PR, the overrides section seems to be covering all the config?

PR #349, opened 2019-12-30
Remove source categories from source creation in terraform script

maimaisie:
###### Description

The terraform script currently specifies the source category for each source to be `{cluster-name}/{source-name}` during source creation, however we did not used to specify source category when using the bash script. 

This change creates confusion to users running queries directly from UI by clicking the query icon next to the source category since it returns no results. The reason is because output plugin sends different source category through HTTP header and that takes precedence over the static source category.

###### Testing performed

- [x] ci/build.sh
- [x] Verify sources created no longer have predefined source categories through UI

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
@@ -84,55 +84,46 @@ data:
 
     resource "sumologic_http_source" "default_metrics_source" {
         name         = local.default-metrics-source-name
-        category     = "${var.cluster_name}/${local.default-metrics-source-name}"
         collector_id = "${sumologic_collector.collector.id}"
     }
 
     resource "sumologic_http_source" "apiserver_metrics_source" {
         name         = local.apiserver-metrics-source-name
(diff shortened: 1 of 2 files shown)


rvmiller89:
I think we did used to specify category in the setup script, https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docker/setup/setup.sh#L40

Category used to be the 3rd argument + "/" + 1st argument, so something like
```
$CLUSTER_NAME/(default-metrics)
$CLUSTER_NAME/apiserver-metrics
$CLUSTER_NAME/kube-controller-manager-metrics
...
```

maimaisie:
Oh, I saw some old collectors with the name `kubernetes-xxxx` on nite without any source categories and I checked the setup script but missed that line, so I thought we didn't used to have them. But it is still kind of confusing to users, what's everyone's opinion on removing them anyway?

rvmiller89:
I think concatenating `CLUSTER_NAME` in source category is necessary as long as we're training customers to query based on that.

So it's more of a product decision. Will defer to Frank/Bryan.

maimaisie:
Discussed in the k8s sync up meeting today and decided it's okay to remove

PR #351, opened 2020-01-07
Add support for data persistence

maimaisie:
###### Description

This PR adds support for data persistence to a persistent volume per fluentd pod. After discussing with the team, we decided to switch from using deployments to statefulsets because statefulsets allows easier setup of persistent volume.

###### Testing performed

- [x] ci/build.sh
- [x] Throttle the account, confirm data is written to buffer and flushed after unthrottling


deploy/helm/sumologic/conf/buffer.output.conf
```diff
@@ -1,5 +1,4 @@
-compress gzip
 flush_interval "#{ENV['FLUSH_INTERVAL']}"
 flush_thread_count "#{ENV['NUM_THREADS']}"
 chunk_limit_size "#{ENV['CHUNK_LIMIT_SIZE']}"
```
rvmiller89: should we increase chunk_limit_size to 1mb knowing that the chunks will no longer be compressed?

maimaisie: we do still do compression for memory buffer and file buffer without persistence, it might be really tricky to have two defaults depending on the buffer type..

rvmiller89: OK I'm fine with keeping this for now. Ideally we figure out how to continue to use gzip since these metrics will compress heavily (repeated dimensions + metadata)

rvmiller89: Could this be the issue we're seeing? https://github.com/fluent/fluentd/issues/2619

Have we tried upgrading our fluentd image from v1.6.3 to v1.8.1 and see if it resolves the issue?

maimaisie: will try that

maimaisie: yep, v1.8.1 fixed the issue! will revert the compression changes

deploy/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fluent/fluentd:v1.6.3-debian-1.0 AS builder
+FROM fluent/fluentd:v1.8.1-debian-1.0 AS builder
```
rvmiller89: could we do this as a separate PR (merged before this one) to isolate the change?

maimaisie: sure

deploy/helm/sumologic/conf/events/events.conf
```diff
@@ -30,11 +30,13 @@
   verify_ssl "#{ENV['VERIFY_SSL']}"
   proxy_uri "#{ENV['PROXY_URI']}"
   <buffer>
-    {{- if eq .Values.sumologic.fluentd.buffer "file" }}
+    {{- if or .Values.sumologic.fluentd.persistence.enabled (eq .Values.sumologic.fluentd.buffer "file") }}
     @type file
     path /fluentd/buffer/events
     {{- else }}
     @type memory
+    compress gzip
```
maimaisie: need to remove this

PR #358, opened 2020-01-09
disable metrics-server dependency by default

vsinghal13:
###### Description

This PR disabled the `metrics-server` helm chart dependency by default. 
`metrics-server` is only required for the `fluentd` HPA , which is disabled by default, so we are also defaulting this dependency to `false`.

It is primarily done to cover the scenario, where the user already has `metrics-server` installed in the k8s cluster, then our helm install command would fail with the error : 

```Error: release collection failed: apiservices.apiregistration.k8s.io "v1beta1.metrics.k8s.io" already exists```

Now, if the user wants to enable autoscaling for fluentd they will have to do the following steps:
1. Enable `metrics-server` dependency 
Note: if `metrics-server` is already installed , this step is not required

2.Enable autoscaling for fluentd

###### Testing performed

- [x] ci/build.sh
- [x] Installed metrics server before running the `helm install` command and verified that helm install and upgrade are successful

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -193,7 +193,7 @@ sumologic:
 ## Configure metrics-server
 ## ref: https://github.com/helm/charts/blob/master/stable/metrics-server/values.yaml
 metrics-server:
-  enabled: true
+  enabled: false
   args:
     - --kubelet-insecure-tls
     - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname


samjsong:
Should we include the documentation changes as part of this PR as well?

vsinghal13:
> Should we include the documentation changes as part of this PR as well?

Actually there is an existing  documentation PR #329 , which is as per the breaking change, so I did not include it now.

samjsong:
Hmm, that PR does not include any docs for autoscaling and metrics server, correct?

vsinghal13:
Yes, that PR does not include this as of now.

PR #364, opened 2020-01-13
Re-add tiller-rbac.yaml

samjsong:
###### Description

Noticed that our jenkins jobs actually downloads this file, and without it I believe the helm installation fails.

I don't think this is enough to fix the jenkins job for helm since it's been failing for a few days, we'll have to keep investigating once this is merged

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/tiller-rbac.yaml
+++ b/deploy/helm/tiller-rbac.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tiller
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: tiller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+  - kind: ServiceAccount
+    name: tiller
+    namespace: kube-system


rvmiller89:
Why does our Jenkins job download this file? I couldn't find us linking to it in our docs.

samjsong:
It looks like it starts from a clean slate each time the job is run, and it reinstalls tiller and helm:
```
sudo: ./get_helm.sh
out: Downloading https://get.helm.sh/helm-v2.16.1-linux-amd64.tar.gz
out: Preparing to install helm and tiller into /usr/local/bin
out: helm installed into /usr/local/bin/helm
out: tiller installed into /usr/local/bin/tiller
out: Run 'helm init' to configure helm.

sudo: curl -LJO https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/master/deploy/helm/tiller-rbac.yaml

sudo: kubectl apply -f tiller-rbac.yaml && helm init --service-account tiller --wait
```

rvmiller89:
Do we really want to be hosting the tiller RBAC needed when installing Helm? I saw we removed that from our docs https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/Non_Helm_Installation.md and instead just link to https://helm.sh/docs/intro/install/

I'm fine with re-adding to give QE a chance to fix it, but I don't think we should be hosting the Tiller RBAC for installing Helm.

samjsong:
cc @Gourav2906 we'll be merging this to unblock for now but I think Ryan's point makes sense that we might not want to host this file in our repo

PR #377, opened 2020-01-27
Add back and expose source category for events source

maimaisie:
###### Description

Since the events data doesn't go through metadata enrichment plugin currently, the source category for events source is just the default `http` after we removed static source category for all sources in #349. 

This PR adds it back just for events source and expose it in values.yaml file. Since after Sam's breaking change, [all events configuration fields](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/330/files#diff-1afd7b203a6373f36b3fdb4d68d7ddb5R273) go under `events` I am adding it here in case this goes out in a minor release before the major one to avoid one more rename that customer needs to do when we do make the major release. 

###### Testing performed

- [ ] ci/build.sh
- [x] Ran `helm template` to verify overwriting the field

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -85,8 +85,9 @@ sumologic:
   # If enabled, collect K8s events
   eventCollectionEnabled: true
 
-  # Source category for the Events source. Default: "{clusterName}/events"
-  #eventsSourceCategory: ""
+  events:
+    # Source category for the Events source. Default: "{clusterName}/events"
+    sourceCategory: ""
 
   ## Format to post logs into Sumo. json, json_merge, or text
   logFormat: fields
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -85,6 +85,9 @@ sumologic:
   # If enabled, collect K8s events
   eventCollectionEnabled: true
 
+  # Source category for the Events source. Default: "{clusterName}/events"
+  #eventsSourceCategory: ""
+
   ## Format to post logs into Sumo. json, json_merge, or text
   logFormat: fields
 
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -85,6 +85,10 @@ sumologic:
   # If enabled, collect K8s events
   eventCollectionEnabled: true
 
+  events:
```
rvmiller89: should we move `eventCollectionEnabled` under this, maybe as:
```
events:
  collectionEnabled: true
  sourceCategory: ""
```
?

deploy/helm/sumologic/values.yaml
```diff
@@ -85,6 +85,10 @@ sumologic:
...
+    # Source category for the Events source. Default: "{clusterName}/events"
+    sourceCategory: ""
```
rvmiller89: why not use what we had below to set the default `sourceCategory: "%{namespace}/%{pod_name}"`

deploy/helm/sumologic/values.yaml
maimaisie: this will be changed after #349

deploy/helm/sumologic/values.yaml
maimaisie: I am using the source category we had for events before for consistency. Also what would pod name be in this case? For logs it's the name of the pod that logs are coming from I assume, but events are not only about pods. We can't use the pod name of the events pod itself either, since we don't know the pod name at the time of the source creation in TF

deploy/helm/sumologic/values.yaml
rvmiller89: I'm confused, #349 is already merged

maimaisie: oh sorry wrong one, it is #330, specifically [this change](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/330/files#diff-1afd7b203a6373f36b3fdb4d68d7ddb5R273)

rvmiller89: Great, thanks for confirming

PR #379, opened 2020-01-29
expose cache params for enhance_k8s_metadata plugin

vsinghal13:
###### Description

Expose cache params for enhance_k8s_metadata plugin and change default cache size to 10000

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
@@ -15,6 +15,8 @@
 <label @NORMAL>
   <filter containers.**>
     @type kubernetes_metadata
+    cache_size  {{ .Values.sumologic.k8sMetadataFilter.cacheSize }}
+    cache_ttl  {{ .Values.sumologic.k8sMetadataFilter.cacheTtl }}
     @log_level warn
     annotation_match ["sumologic\.com.*"]
     de_dot false
--- a/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
@@ -32,6 +32,8 @@
   </filter>
   <filter **>
     @type enhance_k8s_metadata
+    cache_size  {{ .Values.sumologic.k8sMetadataFilter.cacheSize }}
+    cache_ttl  {{ .Values.sumologic.k8sMetadataFilter.cacheTtl }}
     in_namespace_path '$.kubernetes.namespace_name'
     in_pod_path '$.kubernetes.pod_name'
     data_type logs
--- a/deploy/helm/sumologic/conf/metrics/metrics.conf
+++ b/deploy/helm/sumologic/conf/metrics/metrics.conf
@@ -12,8 +12,8 @@
 <label @DATAPOINT>
   <filter prometheus.metrics**>
     @type enhance_k8s_metadata
-    cache_size  {{ .Values.sumologic.k8sMetadataFilter.bearerCacheSize }}
-    cache_ttl  {{ .Values.sumologic.k8sMetadataFilter.bearerCacheTtl }}
+    cache_size  {{ .Values.sumologic.k8sMetadataFilter.cacheSize }}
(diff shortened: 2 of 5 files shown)


deploy/helm/sumologic/conf/metrics/metrics.conf
```diff
@@ -12,6 +12,8 @@
 <label @DATAPOINT>
   <filter prometheus.metrics**>
     @type enhance_k8s_metadata
+    cache_size  {{ .Values.sumologic.k8sMetadataFilter.bearerCacheSize }}
+    cache_ttl  {{ .Values.sumologic.k8sMetadataFilter.bearerCacheTtl }}
```
rvmiller89: should we also add these when we create a filter of type `enhance_k8s_metadata` in `logs.source.containers.conf`?

deploy/helm/sumologic/values.yaml
```diff
@@ -202,7 +202,7 @@ sumologic:
 
     ## Option to control the enabling of metadata filter plugin cache_size. 
     ## ref: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#configuration
-    bearerCacheSize: "1000"
+    bearerCacheSize: "10000"
```
rvmiller89: where did `bearer` come from here? Can we change it to just be `cacheSize` (I'm assuming no one's touched this yet).

rvmiller89: same with `bearerCacheTtl` below.

deploy/helm/sumologic/values.yaml
```diff
@@ -202,7 +202,7 @@ sumologic:
...
 
     ## Option to control the enabling of metadata filter plugin cache_ttl.
```
rvmiller89: can we add `(in seconds)` so to this comment so it's clear

deploy/helm/sumologic/conf/metrics/metrics.conf
vsinghal13: good point. Will add it in the logs pipeline as well.

deploy/helm/sumologic/templates/statefulset.yaml
```diff
@@ -146,9 +146,9 @@ spec:
         - name: K8S_METADATA_FILTER_VERIFY_SSL
           value: {{ .Values.sumologic.k8sMetadataFilter.verifySsl | quote }}
         - name: K8S_METADATA_FILTER_BEARER_CACHE_SIZE
```
rvmiller89: can we remove "BEARER" from these env variable names as well?

rvmiller89: (for the non-Helm users' benefit)

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -15,6 +15,8 @@
 <label @NORMAL>
   <filter containers.**>
     @type kubernetes_metadata
+    cache_size  {{ .Values.sumologic.k8sMetadataFilter.cacheSize }}
```
samjsong: These also need to be added to the `enhance_k8s_metadata` plugin below

deploy/helm/sumologic/templates/statefulset.yaml
vsinghal13: ok. wondering why was it added in the first place as the param name is `cache_size` for the metadata filter plugin as well.

rvmiller89: yeah, not sure

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
samjsong: Just noticed these are defined below, line 29 and line 30.
Since we're still using ENV vars until the major release, we should stay consistent with this and use the new ENV var `K8S_METADATA_FILTER_CACHE_SIZE` and `K8S_METADATA_FILTER_CACHE_TTL` in all of the plugins as well, for consistency

vsinghal13: Thanks Sam! Will float a PR to fix that.

PR #382, opened 2020-01-30
fix discrepancy in metadata cache ENV name

vsinghal13:
###### Description

fix discrepancy in metadata cache ENV name

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
@@ -15,8 +15,6 @@
 <label @NORMAL>
   <filter containers.**>
     @type kubernetes_metadata
-    cache_size  {{ .Values.sumologic.k8sMetadataFilter.cacheSize }}
-    cache_ttl  {{ .Values.sumologic.k8sMetadataFilter.cacheTtl }}
     @log_level warn
     annotation_match ["sumologic\.com.*"]
     de_dot false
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -26,8 +24,8 @@
     client_cert "#{ENV['K8S_METADATA_FILTER_CLIENT_CERT']}"
     client_key "#{ENV['K8S_METADATA_FILTER_CLIENT_KEY']}"
     bearer_token_file "#{ENV['K8S_METADATA_FILTER_BEARER_TOKEN_FILE']}"
-    cache_size "#{ENV['K8S_METADATA_FILTER_BEARER_CACHE_SIZE']}"
-    cache_ttl "#{ENV['K8S_METADATA_FILTER_BEARER_CACHE_TTL']}"
+    cache_size "#{ENV['K8S_METADATA_FILTER_CACHE_SIZE']}"
+    cache_ttl "#{ENV['K8S_METADATA_FILTER_CACHE_TTL']}"
```
samjsong: Can we also change lines 33-34 to use the env var, as well as in the `metrics.conf`, for consistency? Once we make the major release we can use the templating this way as it is cleaner but until then I think we should stay consistent with the use of env vars

vsinghal13: Sure

PR #386, opened 2020-01-31
Add async cache and expose refresh interval in values.yaml

maimaisie:
###### Description

This PR uses the fluentd timer helper plugin to asynchronously refresh the cache entries at a configurable interval. 
(I tried to put the timer logic inside `cache_strategy.rb` together with all the existing caching logic but since it is a module that does not extend fluentd, I couldn't import the fluentd timer there.)

###### Testing performed

- [x] ci/build.sh
- [x] Confirm cache refresh works and reduces prometheus send latency

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -212,6 +212,7 @@ sumologic:
     ## ref: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#configuration
     cacheTtl: "3600"
 
+    cacheRefresh: "600"
 
 ## Configure metrics-server
 ## ref: https://github.com/helm/charts/blob/master/stable/metrics-server/values.yaml
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -212,6 +212,7 @@ sumologic:
     ## ref: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#configuration
     cacheTtl: "3600"
 
+    ## Option to control the interval at which metadata cache is asynchronously refreshed (in seconds).
     cacheRefresh: "600"
 
 ## Configure metrics-server
--- a/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
+++ b/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
@@ -130,13 +130,7 @@ module Fluent
               namespace_name = split[0]
               pod_name = split[1]
               metadata = fetch_pod_metadata(namespace_name, pod_name)
-
-              if !metadata.nil? && !metadata.empty?
-                @cache[entry[0]] = metadata
-              else
-                log.info "Deleting entry for key #{entry[0]}"
-                @cache.delete(entry[0])
-              end
+              @cache[entry[0]] = metadata unless metadata.empty?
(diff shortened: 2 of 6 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -212,6 +212,7 @@ sumologic:
     ## ref: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#configuration
     cacheTtl: "3600"
 
+    cacheRefresh: "600"
```
rvmiller89: can we add a comment explaining this and include `(in seconds)` like we did for TTL?

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -113,7 +115,33 @@ def normalize_param
...
+      end
+
+      def start_cache_timer
+        timer_execute(:"cache_refresher", @cache_refresh) {
+          entries = @cache.to_a
+          log.info "Refreshing metadata for #{entries.count} entries"
+
+          entries.each { |entry|
+            begin
+              log.info "Refreshing metadata for key #{entry[0]}"
```
rvmiller89: could this log be too verbose? Should we move it to `debug` ?

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -113,7 +115,33 @@ def normalize_param
...
+              split = entry[0].split("::")
```
rvmiller89: should we wrap the body of the `entries.each` with a begin/rescue to avoid one bad entry preventing the refresh of subsequent entries?

rvmiller89: Never mind, you've already done this, Nice

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
maimaisie: ah, I forgot to change it

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -113,7 +115,33 @@ def normalize_param
...
+              namespace_name = split[0]
+              pod_name = split[1]
+              metadata = fetch_pod_metadata(namespace_name, pod_name)
+
+              if !metadata.nil? && !metadata.empty?
+                @cache[entry[0]] = metadata
+              else
+                log.info "Deleting entry for key #{entry[0]}"
+                @cache.delete(entry[0])
```
samjsong: I don't think we should delete the entry here - the pod itself might be deleted, but we might still have logs or metrics to collect for that pod due to backpressuring or something. I think we can just let it expire on its own with the cache TTL

maimaisie: that's a great point. I guess I can still check for an empty metadata after looking up the deleted pod, but just not update the entry in the cache so that it can expire on its own

PR #387, opened 2020-01-31
Rename statefulset configs to avoid breaking change

maimaisie:
###### Description

Rename `statefulset` and `eventsStatefulset` back to `deployment` and `eventsDeployment` to avoid breaking change in the minor release. Will revert this PR for major release.

###### Testing performed

- [ ] ci/build.sh


rvmiller89:
Don't recommend changing the template names back since we intend to move to StatefulSet

PR #372, opened 2020-01-17
Modify regexes to match both deprecated and new metrics in 1.16

maimaisie:
###### Description

**Deprecated apiserver metrics:**
`apiserver_request_count` -> `apiserver_request_total`
`apiserver_request_latencies_summary` -> `apiserver_request_duration_seconds`
`etcd_request_cache_get_latencies_summary` -> `etcd_request_cache_get_duration_seconds`
`etcd_request_cache_add_latencies_summary` -> `etcd_request_cache_add_duration_seconds`
`etcd_helper_cache_hit_count` -> `etcd_helper_cache_hit_total`
`etcd_helper_cache_miss_count` -> `etcd_helper_cache_miss_total`

(Verified the apiserver metrics are coming in 1.16, but the etcd ones are not, I think all of these 4 etcd metrics got removed in 1.17 but not sure why they are not showing up in 1.16 either)

**Deprecated kubelet metrics:**
`kubelet_runtime_operations_latency_microseconds` -> `kubelet_runtime_operations_duration_seconds`
`kubelet_docker_operations_latency_microseconds` -> `kubelet_docker_operations_duration_seconds`
`kubelet_docker_operations_errors` -> `kubelet_docker_operations_errors_total`

(Verified all of the old and new kubelet metrics are coming in 1.16)

###### Testing performed

- [x] ci/build.sh
- [x] Redeploy fluentd and fluentd-events pods
- [x] Confirm events, logs, and metrics are coming in

PR #391, opened 2020-02-03
Revert "Modify regexes to match both deprecated and new metrics in 1.16"

maimaisie:
Reverts SumoLogic/sumologic-kubernetes-collection#372 since this we have another minor release to make before the major release


maimaisie:
The changed regexes match both deprecated and renamed metrics, and on 1.16 both sets of metrics are being generated and collected by fluentd if we keep this change. At least one of the metrics has very high cardinality and the increase in DPM is a concern. I'm still talking to Frank if we want to fix this and how for the major release

PR #399, opened 2020-02-05
change fluentd log level to warn

vsinghal13:
###### Description

Since we have changed the default fluentd buffer `overflow_action` to `drop_oldest_chunk`, we can't leave the log level at `error` log and silently drop data. Therefore, changing the log level to  `warn` for now until we integrate the use of fluentd-throttle-plugin.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -159,7 +159,7 @@ sumologic:
 
   ## Sets the fluentd log level. The default log level, if not specified, is error.
   ## ref: https://docs.fluentd.org/deployment/logging
-  fluentdLogLevel: "error"
+  fluentdLogLevel: "warn"
 
   ## Override Kubernetes resource types you want to get events for from different Kubernetes
   ## API versions. The key represents the name of the resource type and the value represents
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -170,6 +170,9 @@ sumologic:
   
 
   fluentd:
+    ## Option to specify the Fluentd system config.
+    ## https://docs.fluentd.org/deployment/system-config#suppress_repeated_stacktrace
+    suppressRepeatedStackTrace: "true"
     ## Option to specify the Fluentd buffer as file/memory.
     buffer: "memory"
     ## Option to turn autoscaling on for fluentd and specify metrics for HPA.
--- a/deploy/helm/sumologic/conf/common.conf
+++ b/deploy/helm/sumologic/conf/common.conf
@@ -14,5 +14,6 @@
 {{- if .Values.sumologic.fluentdLogLevel }}
 <system>
   log_level {{ .Values.sumologic.fluentdLogLevel }}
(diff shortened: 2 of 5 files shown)


maimaisie:
> changing the log level to warn for now until we integrate the use of fluentd-throttle-plugin

Shouldn't we add the fluentd log throtting _before_ changing it to warn? This will just bring back the snowball effect when user is throttled in Sumo.

rvmiller89:
Between data loss and over logging, I'd pick over logging.  Unless we can get the throttling working in the next day or two, since we should try to cut 0.14 release soon.

maimaisie:
Over logging could cause more data loss in that case.. but since we increased the chunk limit size from 100k to 1m, the over logging should be less compared to before because fluentd buffer retry is per chunk. Maybe that is okay for now. I think we should let @frankreno know though

rvmiller89:
At least the customer will be able to see the dropping data (or 429 throttling) by running `kubectl logs` when they don't see data in Sumo, so I'm in favor of this for now.

maimaisie:
Sounds good then.

deploy/helm/sumologic/values.yaml
```diff
@@ -159,7 +159,7 @@ sumologic:
 
   ## Sets the fluentd log level. The default log level, if not specified, is error.
```
maimaisie: actually this comment should be changed

vsinghal13:
One thing that we can do is to configure log level `warn` just fro the output plugin and keep the default global log level to be `error`. This will avoid warning logs from other plugins, which won't be much but will still decrease the log count? 
https://docs.fluentd.org/deployment/logging#per-plugin-log

rvmiller89:
I don't think that's very helpful since 429s are reported by the output plugin. Are some other plugins spamming warn logs?

vsinghal13:
> I don't think that's very helpful since 429s are reported by the output plugin. Are some other plugins spamming warn logs?

```"2020-02-06 19:19:27 +0000 [warn]: #0 no patterns matched tag=\"fluentd.pod.healthcheck\""
```

rvmiller89:
> 2020-02-06 19:19:27 +0000 [warn]: #0 no patterns matched tag="fluentd.pod.healthcheck"

How often are we seeing the `no patterns matched tag` warnings? Is fluentd throttling them itself, or is are they 1:1 with every unmatched log message?

vsinghal13:
Based on the discussion, planning to do the following in a separate PR:
1. Turn off the default fluentd STDOUT logging and write to log file
Eg: ```CMD exec fluentd -o /tmp/fluentd.log``` 

2. Keep global log level to be `error` and just change the log level for the fluentd logs/metrics output plugin to be `warn.

3. Introduce a new `match` for catching all the fluentd logs and only keep logs which contain `drop_oldest_chunk`.
```
<match fluent**>
  @type grep
  <regexp>
    key log
    pattern /drop_oldest_chunk/
  </regexp>
</match>
```

rvmiller89:
I can't seem to find the comment I got emailed about, but I don't think we should have fluentd append to a file `/tmp/fluentd.log` since without file rotation or truncation this will grow indefinitely and could fill the disk.

PR #401, opened 2020-02-06
Support new 1.16 metrics with filter to exclude deprecated metrics

maimaisie:
###### Description

This PR brings back changes in #372 and adds a grep filter plugin to exclude the deprecated metrics by checking the k8s version (which is a built-in object in helm)

###### Testing performed

- [ ] ci/build.sh


deploy/helm/sumologic/conf/metrics/metrics.conf
```diff
@@ -10,6 +10,15 @@
   @label @DATAPOINT
 </match>
 <label @DATAPOINT>
+{{- if and (eq .Capabilities.KubeVersion.Major "1") (eq .Capabilities.KubeVersion.Minor "16") }}
```
rvmiller89: what about 1.17 and above? Or these metrics only apply for 1.16?

maimaisie: I think they are getting removed in 1.17
> Beta release target 1.17
All previously marked deprecated metrics will be removed from the codebase.

in https://github.com/kubernetes/enhancements/issues/1206

frankreno: They are getting removed, but i believe both metrics co-exist in 1.14, 1.15 and 1.16

frankreno: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#deprecated-metrics -  you can see that they deprecated the old and added the new in 1.14.  So, i think this version check should be for 1.14 and above.

maimaisie: I just realized that too, let me update the condition. Thanks!

maimaisie: hmm, what should we do for the non-helm yaml file? it looks like CI added the filter to the generated file by helm template

samjsong:
@maimaisie I just noticed a binary file tgz file got added to the PR, that's not expected is it?

maimaisie:
oh oops, thanks for noticing that! removed

PR #402, opened 2020-02-07
filter fluentd container logs

vsinghal13:
###### Description

This PR changes the default fluentd log level to be `info` and modifies the fluentd pipeline to ingest only specific fluentd logs into Sumo.

###### Testing performed

- [x] ci/build.sh
- [x] Confirm only filtered fluentd logs ingested in Sumo


deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -8,51 +8,64 @@
     time   ${[record["log"].split(/[\n\t]+/)[0]].map! {|item| JSON.parse(item)["time"]}.join("")}
   </record>
 </filter>
+#  only match fluentd logs based on fluentd container name and label them @FLUENT
```
rvmiller89: we should try to see if these logs are already tagged by Fluentd itself: https://docs.fluentd.org/deployment/logging#capture-fluentd-logs

vsinghal13: The label `@FLUENT_LOG` uses `<match fluent.**>`  which will not stop fluentd from outputting to stdout. We will then have duplicate fluentd logs. 
Ref: https://github.com/fluent/fluentd/issues/641#issuecomment-126566460

rvmiller89: Did it not work starting fluentd with `-o /dev/null` to output logs to file /dev/null instead of stdout? https://docs.fluentd.org/deployment/logging#output-to-log-file

vsinghal13:
@rvmiller89 have modified the existing pipeline `NORMAL` label to filter out the fluentd logs instead of having a different pipeline for it. 

@samjsong Still kept the filter plugin confs in separate files to keep the fluentd pipeline dry and easy to read.

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -8,53 +8,37 @@
...
   </record>
 </filter>
+# match all  container logs and label them @NORMAL
 <match containers.**>
   @type relabel
   @label @NORMAL
 </match>
 <label @NORMAL>
+  #  only match fluentd logs based on fluentd container name
+  {{ printf "<filter **%s**>" (include "sumologic.fullname" .) }}
```
vsinghal13: The filter plugin tag for matching fluentd logs is generated via the release name template, so in case user changes the release name , we would still match it.

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -8,53 +8,37 @@
...
+    # only ingest fluentd logs of levels: {error, fatal} and warning messages if buffer is full
+    @type grep
+    <regexp>
+    key log
+    pattern /\[info\]|\[fatal\]|drop_oldest_chunk/
```
samjsong: Wait, we want to ingest info logs here?

vsinghal13: good catch, I had changed that to test locally. Will fix it to be `error`.

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -8,53 +8,37 @@
...
   @type relabel
   @label @NORMAL
 </match>
 <label @NORMAL>
+  #  only match fluentd logs based on fluentd container name
+  {{ printf "<filter **%s**>" (include "sumologic.fullname" .) }}
+    # only ingest fluentd logs of levels: {error, fatal} and warning messages if buffer is full
+    @type grep
+    <regexp>
+    key log
```
maimaisie: need to indent this line and the next?

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -8,53 +8,37 @@
...
-    client_key "#{ENV['K8S_METADATA_FILTER_CLIENT_KEY']}"
-    bearer_token_file "#{ENV['K8S_METADATA_FILTER_BEARER_TOKEN_FILE']}"
-    cache_size "#{ENV['K8S_METADATA_FILTER_CACHE_SIZE']}"
-    cache_ttl "#{ENV['K8S_METADATA_FILTER_CACHE_TTL']}"
-    tag_to_kubernetes_name_regexp '.+?\.containers\.(?<pod_name>[^_]+)_(?<namespace>[^_]+)_(?<container_name>.+)-(?<docker_id>[a-z0-9]{64})\.log$'
+    @include logs.kubernetes.metadata.filter.conf
   </filter>
-  <filter **>
+  # sumologic kubernetes metadata enrichment filter plugin
+  <filter containers.**>
```
vsinghal13: @samjsong  @rvmiller89 I suspect it is because of this change.

PR #403, opened 2020-02-07
add generic fluentd pipeline to catch all logs

vsinghal13:
###### Description

Currently we only collect the container logs in the fluentd pipeline. This PR adds a generic fluentd pipeline to catch all logs. User would need to configure a new Input in the Fluent-bit config to send it to fluentd.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
+++ b/deploy/kubernetes/fluentd-sumologic.yaml.tmpl
@@ -157,6 +157,7 @@ data:
     </source>
     @include logs.source.containers.conf
     @include logs.source.systemd.conf
+    @include logs.source.generic.conf
   logs.output.conf: |-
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -233,6 +233,16 @@ data:
         </buffer>
       </match>
     </label>
+  logs.source.generic.conf: |-
```
samjsong: In order for this pipeline to take effect you'd have to add `logs.source.generic.conf` to `logs.conf`

vsinghal13: ah, missed that. Thanks

PR #408, opened 2020-02-11
Add common configuration and troubleshooting tasks

vsinghal13:
###### Description

This PR add common configuration and troubleshooting tasks which will help the customer 
to customize the solution as per their needs. 

Configs added: 
1. Excluding logs/metrics for certain namespace/container
2. Configure local fluent-bit source

###### Testing performed

- [ ] ci/build.sh


deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
```diff
@@ -4,149 +4,7 @@
     clusterName: "kubernetes"
   },
   sumologicCollector:: {
-    remoteWriteConfigs+: [
```
frankreno: Why did this file change? Something in the CI?

samjsong: weird, good catch Frank. @vsinghal13 can you take a look please?

vsinghal13: hmm.. looks like something in the CI. will take a look.

vsinghal13: The `build.sh` copies the remote write section from the `prometheus-overrides.yaml` to this file and then does a `git diff` to generate new files. But there is no change in the remote write config in this PR.

PR #411, opened 2020-02-12
Zipkin plugins

sumo-drosiek:
###### Description

Add zipkin input and output plugins. As zipkin uses HTTP both plugins are basing on `in_http` and `out_http` fluent plugins

###### Testing performed

- [x] ci/build.sh


rvmiller89:
Do we want to add @perk-sumo  to review?

deploy/helm/sumologic/templates/service.yaml
```diff
@@ -9,6 +9,10 @@ spec:
   selector:
     app: {{ template "sumologic.labels.app" . }}
   ports:
+  - name: zipkin-write
```
perk-sumo: Don't we want to add this only when traces are enabled?

sumo-drosiek: Yeah, I think so 👍

PR #415, opened 2020-02-12
Cut release 0.14.0

vsinghal13:
###### Description

Cut release 0.14.0

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,16 @@ This repo contains all required resources to collect data from Kubernetes cluste
 
 # Installation
 
-Detailed instructions are available in our [Installation Guide](./deploy/README.md).
+Detailed instructions are available in our Installation Guides below.
+
+Sumo Logic Helm Chart Version
+|:--------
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -8,7 +8,16 @@ This repo contains all required resources to collect data from Kubernetes cluste
 
 # Installation
 
-Detailed instructions are available in our [Installation Guide](./deploy/README.md).
+Detailed instructions are available in our Installation Guides below.
+
+Sumo Logic Helm Chart Version
+|:--------
+[0.14.0](./deploy/README.md)
```
rvmiller89: we should pin the current version too. Only the latest alpha should link to master branch

rvmiller89: but thank you for adding this

vsinghal13: hmm.. I was thinking of doing that as well. Will change it, but then we should be careful when doing find and replace for `v0.xx.x` because that might change this URL as well.

PR #424, opened 2020-02-18
Adds the opentelemetry collector

perk-sumo:
###### Description

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
+++ b/deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
@@ -1,0 +1,43 @@
+receivers:
+  opencensus: {}
+  jaeger:
+    protocols:
+      grpc:
+  zipkin: {}
+processors:
+  k8s_tagger:
+    passthrough: false
+    extract:
+      metadata:
+        # extract the following well-known metadata fields
+        - podName
+        - deployment
+        - cluster
+        - namespace
+        - node
+        - startTime
+  queued_retry:
+    num_workers: 16
+    queue_size: 10000
+    retry_on_failure: true
+  batch:
+    send_batch_size: 1024
+    timeout: 5s
+extensions:
+  health_check: {}
+exporters:
+  zipkin:
+    url: http://{{ template "sumologic.fullname" . }}:9411/api/v2/spans
+  # logging:
+  #   loglevel: debug
+service:
+  extensions: [health_check]
+  pipelines:
+    traces/1:
+      receivers: [jaeger, zipkin, opencensus]
+      processors: [batch, queued_retry, k8s_tagger]
+      exporters: [zipkin]
+    traces/2:
+      receivers: [jaeger, zipkin, opencensus]
+      processors: [batch, queued_retry, k8s_tagger]
+      exporters: [zipkin]
--- a/deploy/helm/sumologic/templates/otelcol-deployment.yaml
+++ b/deploy/helm/sumologic/templates/otelcol-deployment.yaml
@@ -1,0 +1,66 @@
+{{- if eq .Values.sumologic.traces.enabled true }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ printf "%s-otelcol" (include "sumologic.fullname" .) }}
+  labels:
+    app: {{ printf "%s-otelcol" (include "sumologic.labels.app" .) }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+spec:
+  minReadySeconds: 5
+  progressDeadlineSeconds: 120
+  replicas: {{ .Values.otelcolDeployment.replicas }}
+  selector:
+    matchLabels:
+      app: {{ printf "%s-otelcol" (include "sumologic.labels.app" .) }}
+  template:
+    metadata:
+      labels:
+        app: {{ printf "%s-otelcol" (include "sumologic.labels.app" .) }}
+        {{- include "sumologic.labels.common" . | nindent 8 }}
+    spec:
+      serviceAccountName: {{ template "sumologic.fullname" . }}
+      {{- if .Values.otelcolDeployment.nodeSelector }}
+      nodeSelector:
+        {{ toYaml .Values.otelcolDeployment.nodeSelector | indent 8 }}
+      {{- end }}
+      {{- if .Values.otelcolDeployment.tolerations }}
+      tolerations:
+        {{ toYaml .Values.otelcolDeployment.tolerations | indent 8 }}
+      {{- end }}
+      containers:
+      - name: otelcontrib-collector
+        image: {{ .Values.otelcolDeployment.image.name }}:{{ .Values.otelcolDeployment.image.tag }}
+        imagePullPolicy: {{ .Values.otelcolDeployment.image.pullPolicy }}
+        command:
+          - "/otelcontribcol"
+          - "--config=/conf/traces.otelcol.conf.yaml"
+          - {{ printf "--mem-ballast-size-mib=%s" .Values.otelcolDeployment.memBallastSizeMib }}
+        env:
+        - name: GOGC
+          value: "80"
+        resources:
+          {{- toYaml .Values.otelcolDeployment.resources | nindent 10 }}
+        ports:
+        - containerPort: 55678 # Default endpoint for Opencensus receiver.
+        - containerPort: 14267 # Default endpoint for Jaeger TChannel receiver.
+        - containerPort: 14268 # Default endpoint for Jaeger HTTP receiver.
+        - containerPort: 14250 # Default endpoint for Jaeger gRPC receiver.
+        - containerPort: 9411 # Default endpoint for Zipkin receiver.
+        - containerPort: 8888  # Default endpoint for querying metrics.
+        volumeMounts:
+        - name: otel-collector-config-vol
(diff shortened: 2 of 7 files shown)


deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -0,0 +1,51 @@
...
+  #   endpoint: "someopencensus.target.com:443" # Replace with a real endpoint.
+  #   num_workers: 8
+  #   reconnection_delay: 2s
+  #   secure: true
+  # zipkin:
+  #   url: "http://somezipkin.target.com:9411/api/v2/spans" # Replace with a real endpoint.
+  # logging:
+  #   loglevel: debug
+  jaeger_grpc:
+    endpoint: "jaeger:14250" # Replace with a real endpoint.
```
sumo-drosiek: we want to use zipkin => `http://jaeger:9411/api/v2/spans`

deploy/helm/sumologic/templates/otelcol-deployment.yaml
```diff
@@ -0,0 +1,70 @@
+{{- if eq .Values.sumologic.traces.enable true }}
+apiVersion: extensions/v1beta1
```
frankreno: should probably update this as this api is gone in 1.16

deploy/helm/sumologic/templates/otelcol-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+kind: Deployment
+metadata:
+  name: {{ printf "%s-otelcol" (include "sumologic.name" .) }}
+  labels:
+    app: {{ printf "%s-otelcol" (include "sumologic.labels.app" .) }}
```
frankreno: any reason not to move this logic to the tpl so you can just call template?

deploy/helm/sumologic/templates/otelcol-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+  replicas: {{ .Values.otelcolDeployment.replicas }}
+  selector:
+    matchLabels:
+      app: {{ printf "%s-otelcol" (include "sumologic.labels.app" .) }}
+  template:
+    metadata:
+      annotations:
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "8888"
+        prometheus.io/scrape: "true"
```
frankreno: we use Prometheus Operator for metrics collection.  That is what we recommend customers deploy. When using the Operator, annotation-based scraping is not supported and you must use a ServiceMonitor instead.  here is how we get the Fluentd and fluent-bit metrics we insturment - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L336

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -727,3 +727,13 @@ spec:
 ---
 # Source: sumologic/templates/hpa.yaml
 
+---
```
frankreno: We will probably need to modify the CI process, such that the YAML for Tracing is rendered seperately. Otherwise, customers will pick it up in the non-helm installation.

deploy/helm/sumologic/values.yaml
```diff
@@ -28,6 +28,24 @@ eventsDeployment:
...
+      cpu: 1
+    requests:
+      memory: 384Mi
+      cpu: "200m"
+  # Memory Ballast size should be max 1/3 to 1/2 of memory.
+  memBallastSizeMib: "683"
+  image:
+    name: "omnition/opentelemetry-collector-contrib"
+    tag: "0.0.5"
+    pullPolicy: "Always"
```
frankreno: Any reason to not use `IfNotPresent`?

deploy/helm/sumologic/values.yaml
```diff
@@ -89,6 +107,10 @@ sumologic:
     # Source category for the Events source. Default: "{clusterName}/events"
     sourceCategory: ""
 
+  # Traces configuration
+  traces:
+    enable: false
```
frankreno: nit - for consitency with other flags to control can you change to `enabled`

frankreno: also, lets make sure we make it clear that the tracing configuration is alpha and not for general use.

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
sumo-drosiek: When the fluentd-zipkin will be merge, even `http://{{ sumologic.fullname }}:9411/api/v2/spans`

deploy/helm/sumologic/values.yaml
sumo-drosiek: @perk-sumo 
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/drosiek-zipkin-plugin/deploy/helm/sumologic/values.yaml#L92

I will also change `enable` -> `enabled`

perk-sumo: fixed; I wanted to be consistent with previous PRs but I like `enabled` more :)

deploy/helm/sumologic/values.yaml
perk-sumo: We were thinking about the `latest` anyway until we have a real release to speed up the prototyping. But we can always just use helm parameters for that. Let me change that.

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
perk-sumo: Fixed.

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - cluster
+        - namespace
+        - node
+        - startTime
+  queued_retry:
+    num_workers: 16
+    queue_size: 10000
+    retry_on_failure: true
+  batch:
+    send_batch_size: 128
```
pmm-sumo: I think the default was 1024 which I decreased to make sure we are not waiting too long for packages, but I think we might up it to at least 512 (if not the OT default)

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
pmm-sumo: Note this means that both of those exporters will be enabled and trying to connect

deploy/helm/sumologic/templates/otelcol-deployment.yaml
perk-sumo: Fixed.

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -0,0 +1,51 @@
...
+  #   loglevel: debug
+  jaeger_grpc:
+    endpoint: "jaeger:14250" # Replace with a real endpoint.
+service:
+  extensions: [health_check]
+  pipelines:
+    traces/1:
+      receivers: [jaeger]
+      processors: [batch, queued_retry, k8s_tagger]
+      exporters: [jaeger_grpc]
```
pmm-sumo: Lets use zipkin pipeline here (and in traces/2)

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
perk-sumo: 👍

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
perk-sumo: Done :)

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
perk-sumo: @pmm-sumo is it OK now?

deploy/helm/sumologic/templates/otelcol-deployment.yaml
perk-sumo: I just wanted to be consistent with events deployment:  https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/events-deployment.yaml#L5

Should we change that in both places?

deploy/helm/sumologic/templates/otelcol-deployment.yaml
perk-sumo: 👍 done

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
perk-sumo: Does this action block merging?

frankreno: It looks like the YAML is only rendered when enabled is set to true, but would be good to confirm that is the behavior of the automation.  @rvmiller89 @vsinghal13 @samjsong can you confirm? If that is the case, it should be fine to merge.  we just want to make sure non-helm installs are affected.

PR #425, opened 2020-02-18
Revert "Revert "Bump fluentd image version to fix fluentd buffer bug""

maimaisie:
Reverts SumoLogic/sumologic-kubernetes-collection#361


maimaisie:
Not sure why all the override files are modified, maybe because this is reverting an old commit from a while ago? Let me cherry-pick the original change and open a new PR

PR #429, opened 2020-02-19
[1.0.0] Persistence volume

maimaisie:
###### Description

Brings back changes from #351, #388 and add headless services for statefulsets. 

###### Testing performed

- [ ] ci/build.sh
- [x] Redeploy fluentd pods during quota is exceeded and ensure data is not lost after unthrottling the account


maimaisie:
@vsinghal13 I think it's used to assign unique network identity to each pod in the statefulset because pods in statefulsets are unique and stateful, you can read more here https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id

PR #431, opened 2020-02-19
Support custom release and namespace

vsinghal13:
###### Description

This PR makes use of a new configmap to store the release name and using the k8s downward API for accessing the namespace where the pods are deployed. These two params are then exposed as env variables to fluent-bit and prometheus, to make the configuration dynamic.

###### Testing performed

- [x] ci/build.sh
- [x] Verified helm installation with a custom release name and different namespace


deploy/helm/sumologic/templates/chart-configmap.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sumologic-configmap
```
rvmiller89: do we need an entirely new configmap, or could we just put this data into the existing configmap?

vsinghal13: I did not add this in the `collection-sumologic` config map as that is primarily for fluentd config.

rvmiller89: I'm confused, a configmap can store more than one key...

And what about a map called `collection-sumologic` is specific to fluentd?

rvmiller89: spoke offline, since we're already creating separate configmaps for each deployment, and this one spans across deployments, I'm ok with a separate configmap.

In the future we may want to consolidate all these configmaps though to reduce complexity of the chart.

deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
```diff
@@ -6,7 +6,7 @@
   sumologicCollector:: {
     remoteWriteConfigs+: [
       {
-        url: $._config.sumologicCollectorSvc + "prometheus.metrics.state",
+        url: "http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.state",
```
samjsong: We'll need to fix this generated file in the CI

deploy/helm/prometheus-overrides.yaml
```diff
@@ -49,39 +49,53 @@ prometheus:
...
   prometheusSpec:
+    thanos:
+      version: v0.10.0
+    containers:
+    - name: "prometheus-config-reloader"
+      env:
+      - name: CHART
+        valueFrom:
+          configMapKeyRef:
+            name: sumologic-configmap
```
samjsong: Have we tested the non-helm case? What happens if the configmap has not been created yet at the time of prometheus install? Will this be able to pick up the values from the configmap later when it is created? Same for `fluent-bit`

vsinghal13: Since the configmap creation is part of the `fluentd-sumologic.yaml.tmpl`, it will be created as the first step as per the Non-Helm Installation docs, so it will be available by default. https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v0.14.0/deploy/docs/Non_Helm_Installation.md#deploy-fluentd

vsinghal13:
> LGTM, one question, have you tested the upgrade path?

@frankreno Yes, verified the upgrade path as well. It is working fine.

deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
vsinghal13: Done

PR #433, opened 2020-02-20
Fix fluentd zipkin json parsing

sumo-drosiek:
###### Description

FIx fluent json parsing
Fix zipkin receiver port exposition

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
+++ b/deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
@@ -3,7 +3,8 @@ receivers:
   jaeger:
     protocols:
       grpc:
-  zipkin: {}
+  zipkin:
+    endpoint: "0.0.0.0:9411"
 processors:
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -3,7 +3,8 @@ receivers:
   jaeger:
     protocols:
       grpc:
-  zipkin: {}
+  zipkin:
+    endpoint: "0.0.0.0:9411"
```
mrumian-sumo: I think we could check if hostname name from `values.yaml` will work in this place.

sumo-drosiek: I could use `{{ printf "%s-otelcol" (include "sumologic.fullname" .) }}`, but it will be easy to omit while updating collector name :)

PR #434, opened 2020-02-20
V0.14.0 audit patch

granular-ryanbonham:
###### Description

###### Testing performed

- [ ] ci/build.sh


deploy/helm/sumologic/conf/logs/logs.conf
```diff
@@ -5,4 +5,5 @@
 </source>
 @include logs.source.containers.conf
 @include logs.source.systemd.conf
-@include logs.source.generic.conf
\ No newline at end of file
+@include logs.source.generic.conf
+@include logs.source.kube-audit.conf
```
vsinghal13: The order should be reverse for these two. The generic filter should be at the end .

deploy/helm/sumologic/conf/logs/logs.source.kube-audit.conf
```diff
@@ -0,0 +1,29 @@
+<match kube-audit.**>
+    @type relabel
+    @label @KUBEAUDIT
+</match>
+<label @KUBEAUDIT>
+    <filter kube-audit.**>
```
vsinghal13: Indentation looks wrong, please correct.

deploy/helm/sumologic/values.yaml
```diff
@@ -311,7 +311,19 @@ fluent-bit:
...
+    [INPUT]
+      Name              tail
+      Path              /var/log/kube-apiserver-audit.log
+      Parser            docker
+      Tag               audit.*
+      Refresh_Interval  5
+      Rotate_Wait       60
+      Mem_Buf_Limit     35MB
+      Buffer_Chunk_Size 2MB
+      Buffer_Max_Size   10MB
```
vsinghal13: @rvmiller89 @frankreno Do we want to allow this buffer limit? I would recommend keeping the same as for the container logs input section.

rvmiller89: +1 to keeping the same for consistency

deploy/helm/sumologic/conf/logs/logs.conf
granular-ryanbonham: Fixed the ordering

deploy/helm/sumologic/conf/logs/logs.source.kube-audit.conf
granular-ryanbonham: Yep tabs were set to 4 spaces, should be 2 now..

deploy/helm/sumologic/values.yaml
granular-ryanbonham: So are you suggesting that we set the " Mem_Buf_Limit    5MB" and then remove the "Buffer_Chunk_Size" and "Buffer_Max_Size" or where do you want them set to?

granular-ryanbonham: Note that i got these setting from the upstream fluent bit helm chart, for what they are setting when tailing the kube-audit logs. See https://github.com/helm/charts/tree/master/stable/fluent-bit#configuration

frankreno: We should probably be consistent with upstream...

samjsong:
Thanks for contributing! LGTM, we'll perform some testing internally before merging.

frankreno:
the new fluent-bit chart changes how this works entirely.  so lets move to 2.1 and handle once we migrate to the new fluent-bit chart.

frankreno:
Out helm chart has changed quite a lot since this PR was originally created.  There are also changes to the fluent-bit helm chart depdency.  I am closing this PR.  BUt will create a new issue to track making it easier to collect the K8s audit logs.

PR #437, opened 2020-02-21
add clarifying doc for helm upgrades

vsinghal13:
###### Description

This PR adds a clarifying statement to use the latest`values.yaml` before running the helm upgrade command.

###### Testing performed

- [x] ci/build.sh


vsinghal13:
> Should this note be added in the beginning of the document? It seems like we want to tell them to download the latest version no matter what. Why did we just put it in the example section?

Correct, we want it to be downloaded no matter what. I wanted to keep it close to the helm upgrade commands, that's why kept it near the examples.

PR #438, opened 2020-02-21
Add more attributes extracted for traces

pmm-sumo:
###### Description

Adds more fields extracted, when used with our extended OpenTelemetry Collector build. Needs adding the image first (@mrumian-sumo, can you help?)

###### Testing performed

- [x] ci/build.sh


pmm-sumo:
@sumo-drosiek how about using the new image with our Collector distribution?
@perk-sumo could you take a look?

sumo-drosiek:
@perk-sumo Is it up to date? https://hub.docker.com/r/sumologic/opentelemetry-collector/tags

perk-sumo:
> @perk-sumo Is it up to date? https://hub.docker.com/r/sumologic/opentelemetry-collector/tags

Yes, this is the registry we're using and that's where CI is putting the images.
I can see there is the v0.2.6.1 release in there:
https://github.com/SumoLogic/opentelemetry-collector-contrib/releases/tag/v0.2.6.1
https://hub.docker.com/layers/sumologic/opentelemetry-collector/0.2.6.1/images/sha256-21cb9533bbd3b65ed3f4258516bd254ac7c221374635d547d71b99c98c473e66?context=explore

deploy/helm/sumologic/values.yaml
```diff
@@ -42,8 +42,8 @@ otelcolDeployment:
   # Memory Ballast size should be max 1/3 to 1/2 of memory.
   memBallastSizeMib: "683"
   image:
-    name: "omnition/opentelemetry-collector-contrib"
-    tag: "0.0.5"
+    name: "sumologic/opentelemetry-collector"
+    tag: "0.0.0"
```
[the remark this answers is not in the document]
perk-sumo: That's really good point - please use the tagged otelcol release.

sumo-drosiek:
> > @perk-sumo Is it up to date? https://hub.docker.com/r/sumologic/opentelemetry-collector/tags
> 
> Yes, this is the registry we're using and that's where CI is putting the images.
> I can see there is the v0.2.6.1 release in there:
> https://github.com/SumoLogic/opentelemetry-collector-contrib/releases/tag/v0.2.6.1
> https://hub.docker.com/layers/sumologic/opentelemetry-collector/0.2.6.1/images/sha256-21cb9533bbd3b65ed3f4258516bd254ac7c221374635d547d71b99c98c473e66?context=explore

I asked before I realized that it's updating after tagging

sumo-drosiek:
I think we are good to go. @pmaciolek could you review my changes?

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -4,71 +4,47 @@ receivers:
...
-    queue_size: 10000
-    retry_on_failure: true
+    num_workers: {{ .Values.otelcol.processors.queued_retry.num_workers }}
+    queue_size: {{ .Values.otelcol.processors.queued_retry.queue_size }}
+    retry_on_failure: {{ .Values.otelcol.processors.queued_retry.retry_on_failure }}
   batch:
-    send_batch_size: 1024
-    timeout: 5s
+    send_batch_size: {{ .Values.otelcol.processors.batch.send_batch_size }}
+    timeout: {{ .Values.otelcol.processors.batch.timeout }}
```
pmm-sumo: @perk-sumo The challenge here is that there is a number of processors: https://github.com/open-telemetry/opentelemetry-collector/tree/master/processor - the list is actually growing and their properties might change over time.

An ideal solution would to provide a template with the full config file, but we are not sure how to tackle it. Do you have some suggestions?

pmm-sumo: @sumo-drosiek perhaps we should minimally add sampling processors: https://github.com/open-telemetry/opentelemetry-collector/tree/master/processor#sampling-processor ?

sumo-drosiek: As I think about it more and more, we should just take full config branch and put it as a configmap. If user have own processors it's gonna add changes to values or config file anyway. Additionally we want to support `--set` flags

@perk-sumo any thoughts?

sumo-drosiek: not so nice proposal, but had to support dynamic value for url :D

https://github.com/SumoLogic/sumologic-kubernetes-collection/commit/bb9c9b08b6a8cffd2edbf2797918da737d131b4f

perk-sumo: I'd like to keep some defaults for clients who don't want to deal with the config that much.
We could do it like fluent-bit: provide sane defaults in config.yaml file with some flags in values.yaml and if someone want to have a totally different config provide a special key in values.yaml for that.
We could also provide the template config but I'm not sure if that should go to values.yaml or we should create a new file for that like values.otelcol.yaml or sth. With that client would need to copy and paste something into values.yaml by hand.
I don't like the idea of whole otelcol config inside default values.yaml because it will grow really huge and that doesn't seem right.

Please take a look at:
https://github.com/helm/charts/blob/master/stable/fluent-bit/values.yaml#L146
https://github.com/helm/charts/blob/master/stable/fluent-bit/templates/config.yaml#L1

It's similar to what @sumo-drosiek did.

sumo-drosiek: `We could do it like fluent-bit` I think that our case differs. We are able to use `yaml`. fluentbit were obligated to convert yaml configuration into fluent-bit specific format.
I would like to stick to the `config.yaml`. It's more clear and in both cases `values.yaml` will be growing

PR #440, opened 2020-02-21
Allow PR reverts in CI build script

rvmiller89:
###### Description

Skips check for touching generated files when the branch name starts with "revert-" to allow PR reverts through Github.

Will need to test by opening a revert after this is merged (e.g. https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/439)

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -23,7 +23,8 @@ if [ -n "$GITHUB_TOKEN" ]; then
 fi
 
 # Check for invalid changes to generated yaml files (non-Tag builds)
-if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
+# Exclude branches that start with "revert-" to allow reverts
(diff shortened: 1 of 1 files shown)


ci/build.sh
```diff
@@ -23,7 +23,8 @@ if [ -n "$GITHUB_TOKEN" ]; then
 fi
 
 # Check for invalid changes to generated yaml files (non-Tag builds)
-if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
+# Exclude branches that start with "revert-" to allow reverts
+if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ] && [[ ! "$TRAVIS_PULL_REQUEST_BRANCH" =~ ^revert- ]]; then
```
samjsong: Was the use of double brackets `[[ ]]` intentional? Not super familiar with bash but if it's not necessary here I vote we stick with single brackets for consistency

rvmiller89: Yes, the double brackets allow evaluating a regular expression in bash: https://unix.stackexchange.com/questions/306111/what-is-the-difference-between-the-bash-operators-vs-vs-vs
> Among its extended features, it can test whether a string matches a regular expression.

PR #444, opened 2020-02-21
Update docs for prometheus dynamic remote write urls

vsinghal13:
###### Description

With the new feature of installing the helm chart with a custom release name and a namespace, the prometheus remote write urls are dynamic. One of the variables used in these urls is sourced from a configmap and exposed via thanos in prometheus-operator.

If the user has existing prometheus/prometheus-operator, they will need to modify the urls. 

**Case 1 : Customer has existing prometheus operator in a different namespace.** 
 - Post installation of the sumologic helm chart, copy the `sumologic-configmap` to the namespace where prometheus-operator is installed.

**Case 2: Standalone Prometheus**
 - In this case the user will have to manually modify the remote write urls to point to the right location.

**Case 3: Side-by-side prometheus along with exisiting prometheus server in different namespace**
 - In this case we still deploy prometheus-operator and recommend using a different port eg. 9200. So, the original configmap should be accessible and the namespace will automatically be populated where the new prometheus server is started.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/docs/Installation_with_Helm.md
+++ b/deploy/docs/Installation_with_Helm.md
@@ -69,26 +69,10 @@ To customize your configuration, download the values.yaml file by running
 curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/v0.15.0/deploy/helm/sumologic/values.yaml
 ```
 
-NOTE: If you need to install the chart with a different release name or namespace you will need to override some configuration fields for both Prometheus and fluent-bit. We recommend using an override file due to the number of fields that need to be overridden. In the following command, replace the `<RELEASE-NAME>` and `<NAMESPACE>` variables with your values and then run it to download the override file with your replaced values:
+####To install the chart with a different release name or namespace:
 
 ```bash
-curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/v0.15.0/deploy/helm/sumologic/values.yaml | \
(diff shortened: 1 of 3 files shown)


deploy/docs/Installation_with_Helm.md
```diff
@@ -69,26 +69,10 @@ To customize your configuration, download the values.yaml file by running
 curl https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/v0.15.0/deploy/helm/sumologic/values.yaml
 ```
 
-NOTE: If you need to install the chart with a different release name or namespace you will need to override some configuration fields for both Prometheus and fluent-bit. We recommend using an override file due to the number of fields that need to be overridden. In the following command, replace the `<RELEASE-NAME>` and `<NAMESPACE>` variables with your values and then run it to download the override file with your replaced values:
+####To install the chart with a different release name or namespace:
```
samjsong: I think you need a space `#### To` for the header to be parsed correctly

Maybe we can get rid of this section altogether? since now there is nothing special about `name` or `namespace`, from the default install command the users would be able to tell that they could change the name or namespace parameters. What do we think?

vsinghal13: I think we should still keep the example command, as everywhere else we are using the default release name and namespace.

vsinghal13:
Have pushed the changes to branch `v0.15.0`. Merging to master to be consistent.

PR #445, opened 2020-02-24
Obtain configuration for tracing via terraform

sumo-drosiek:
###### Description

Get configuration for receiver endpoint via terraform

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
@@ -32,7 +32,9 @@ data:
     terraform import sumologic_http_source.kubelet_metrics_source "$COLLECTOR_NAME/kubelet-metrics"
     terraform import sumologic_http_source.logs_source "$COLLECTOR_NAME/logs"
     terraform import sumologic_http_source.node_exporter_metrics_source "$COLLECTOR_NAME/node-exporter-metrics"
+    {{- if .Values.sumologic.traces.enabled }}
     terraform import sumologic_http_source.traces_source "$COLLECTOR_NAME/traces"
+    {{- end }}
 
     # Kubernetes Namespace and Secret
     terraform import kubernetes_namespace.sumologic_collection_namespace {{ .Release.Namespace }}
@@ -72,7 +74,9 @@ data:
       kubelet-metrics-source-name                 = "kubelet-metrics"
       logs-source-name                            = "logs"
       node-exporter-metrics-source-name           = "node-exporter-metrics"
+      {{- if .Values.sumologic.traces.enabled }}
       traces-source-name                          = "traces"
+      {{- end }}
     }
 
     provider "sumologic" {}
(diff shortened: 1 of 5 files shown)


deploy/helm/sumologic/templates/setup/setup-configmap.yaml
```diff
@@ -128,6 +130,11 @@ data:
         collector_id = "${sumologic_collector.collector.id}"
     }
 
+    resource "sumologic_http_source" "traces_source" {
```
Gourav2906: I think we are exposing new source from the collection side so this is not right maybe? becuase it will be different from sumologic_http_source

frankreno:
One other comment, to avoid confusion of customers seeing a source called traces, this should only be created when tracing is enabled.

deploy/helm/sumologic/templates/setup/setup-configmap.yaml
```diff
@@ -128,6 +134,13 @@ data:
         collector_id = "${sumologic_collector.collector.id}"
     }
 
+    {{- if .Values.sumologic.traces.enabled }}
+    resource "sumologic_http_source" "traces_source" {
```
Gourav2906: I think you need to set Content-type here for making tracing type source otherwise this will make normal HTTP source cc @tanayjha

sumo-drosiek: @Gourav2906 We are going to use `application/vnd.sumologic.zipkin`

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/conf/traces/traces.conf#L16

Do we need to add sth to terraform too?

tanayjha: Yes we need to add `contentType` and we don't need to add `application/vnd.sumologic.zipkin` header anymore. 
So the json will look like to create a Trace source:
```
{
    "api.version": "v1",
    "source": {
        "messagePerRequest": false,
        "sourceType": "HTTP",
        "contentType": "Zipkin",
        "name": "TraceTrySource"
    }
}
```
(Notice the contentType field set to **Zipkin**)
This will return a trace endpoint, which will look like: `http://localhost:23660/receiver/v1/trace/Za***`
You can directly send data to this endpoint and it would accepted as trace (no need to add any headers)

sumo-drosiek:
> LGTM. Have we tested this at the scale customers will be running? Tested the extra load on fluentD as it now is processing logs, metrics and traces?

It's our main goal for nearest days

sumo-drosiek:
Related terraform PR:
https://github.com/SumoLogic/sumologic-terraform-provider/pull/123

deploy/helm/sumologic/templates/statefulset.yaml
```diff
@@ -165,4 +172,4 @@ spec:
       resources:
         requests:
           storage: {{ .Values.fluentd.persistence.size }}
-{{- end }}
\ No newline at end of file
+{{- end }}
```
sumo-drosiek: rebase leftover

sumo-drosiek:
@perk-sumo I think we should skip it for v0.16. Till tracing is not fully supported adding it to terraform and realesing is not the best choice

perk-sumo:
This is blocked by https://github.com/SumoLogic/sumologic-terraform-provider/pull/123

sumo-drosiek:
This will be resolved by #708 with values.yaml based approach

PR #446, opened 2020-02-24
[1.0.0] Revert "Revert "[breaking change] Add pre-upgrade hook for setup""

maimaisie:
Reverts SumoLogic/sumologic-kubernetes-collection#357


maimaisie:
I think this is needed especially now that we have the new `traces` source

samjsong:
cc @sumo-drosiek

PR #450, opened 2020-02-26
[1.0.0] change default DB path for fluentbit

vsinghal13:
###### Description

Fluent bit uses a DB to keep track of monitored files and offsets. This PR modifies the default DB path which is specified in the fluent-bit helm chart to avoid having `database error` in case a user already has fluent-bit installed in the cluster with the default DB path. 

###### Testing performed

- [x] ci/build.sh


samjsong:
since this is easily configurable in `values.yaml`, let's hold on merging until after major release (or with the major release, if it can cause a spike in logs due to re-ingestion).

samjsong:
Did we ever confirm whether this will lead to duplicate ingest?

PR #457, opened 2020-03-02
pin falco image tag

vsinghal13:
###### Description

This PR pins the falco image tag to use `0.18.0` for resoving the precomiped kernel modules error.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -605,6 +605,8 @@ prometheus-operator:
 ## ref: https://github.com/helm/charts/blob/master/stable/falco/values.yaml
 falco:
   enabled: true
+  image:
+    tag: 0.18.0
   #ebpf:
   #  enabled: true
   falco:


rvmiller89:
Is this recommended with Helm, or is it better practice to use the image provided in the Helm chart (whose subchart dependency is already pinned in `requirements.xml`)?

vsinghal13:
> Is this recommended with Helm, or is it better practice to use the image provided in the Helm chart (whose subchart dependency is already pinned in `requirements.yaml`)?

This is done as a work-around for now to get the precompiled kernel modules. We would have downgrade the falco helm chart version to get this particular app version and then upgrade it again later on. 

With this change, we can override the image tag and still keep the helm chart version intact.

rvmiller89:
> With this change, we can override the image tag and still keep the helm chart version intact.

Not opposed to this, but why not downgrade the helm chart version then?

perk-sumo:
> > With this change, we can override the image tag and still keep the helm chart version intact.
> 
> Not opposed to this, but why not downgrade the helm chart version then?

That is an excellent point by @rvmiller89 - I think we should just downgrade the whole falco chart to `1.1.0`
Otherwise we could run into a problem when Docker image does not fit the helm chart.

vsinghal13:
@rvmiller89 @perk-sumo Thanks for the inputs. Modified it to use the older helm chart version.

deploy/helm/sumologic/requirements.yaml
```diff
@@ -8,7 +8,7 @@ dependencies:
     repository: https://kubernetes-charts.storage.googleapis.com/
     condition: prometheus-operator.enabled
   - name: falco
-    version: 1.1.1
+    version: 1.1.0
```
rvmiller89: should also update the dependency table in master branch so it gets included when we cut the next release?

PR #467, opened 2020-03-05
Semver ci adjustments

perk-sumo:
###### Description

Do not release alpha version with every commit to master
Use git describe for master branch development build tags
Extract duplicated code into functions

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -151,7 +151,7 @@ if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
   fi
 fi
 
-if [ -n "$DOCKER_PASSWORD" ] && [ -n "$TRAVIS_TAG" ] && [[ $TRAVIS_TAG != *alpha* ]]; then
+if [ -n "$DOCKER_PASSWORD" ] && [ -n "$TRAVIS_TAG" ]; then
   echo "Tagging docker image $DOCKER_TAG:local with $DOCKER_TAG:$VERSION..."
   docker tag $DOCKER_TAG:local $DOCKER_TAG:$VERSION
   echo "Pushing docker image $DOCKER_TAG:$VERSION..."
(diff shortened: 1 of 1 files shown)


ci/build.sh
```diff
@@ -151,61 +151,40 @@ if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
...
+  push_docker_image "$VERSION"
+  push_helm_chart "$VERSION"
 
-  echo "Tagging git with v$new_alpha..."
-  git tag -a "v$new_alpha" -m "Bump version to v$new_alpha"
-  git push --tags --quiet --set-upstream origin-repo master
+elif [ -n "$DOCKER_PASSWORD" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_EVENT_TYPE" == "push" ]; then
+  dev_build_tag=$(git describe --tags --always)
+  push_docker_image "$dev_build_tag"
+  push_helm_chart "$dev_build_tag"
```
sumo-drosiek: @perk-sumo 
As I understand, this is still going to relase all master changes?

sumo-drosiek: Got familiar with doc, nvm :)

ci/build.sh
```diff
@@ -151,61 +151,40 @@ if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ]; then
...
-  docker push $DOCKER_TAG:$new_alpha
+if [ -n "$DOCKER_PASSWORD" ] && [ -n "$TRAVIS_TAG" ]; then
+  push_docker_image "$VERSION"
+  push_helm_chart "$VERSION"
 
-  echo "Tagging git with v$new_alpha..."
-  git tag -a "v$new_alpha" -m "Bump version to v$new_alpha"
-  git push --tags --quiet --set-upstream origin-repo master
+elif [ -n "$DOCKER_PASSWORD" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_EVENT_TYPE" == "push" ]; then
+  dev_build_tag=$(git describe --tags --always)
```
rvmiller89: if I'm merging a commit from a branch that predates one or more tags that have been created since it was branched, will this return the correct tag?

rvmiller89:
> Do not release alpha version with every commit to master

How will QE test builds?

ci/build.sh
perk-sumo: Yeah, it works, as per doc: `The command finds the most recent tag that is reachable from a commit.`
https://git-scm.com/docs/git-describe

I've just checked it locally and it's working as described for merged branches as well.

perk-sumo:
> > Do not release alpha version with every commit to master
> 
> How will QE test builds?

The idea is to release an alpha build by ourselves (create an alpha tag, eg. `v0.16.0-alpha.0`) whenever we think it's ready for testing.

rvmiller89:
> The idea is to release an alpha build by ourselves (create an alpha tag, eg. v0.16.0-alpha.0) whenever we think it's ready for testing.

Seems kind of counter to CI. Any reason we shouldn't test continuously?

perk-sumo:
> Seems kind of counter to CI. Any reason we shouldn't test continuously?

Ah, sorry, you asked about QE master builds - they will be there (both helm chart and docker image), they just won't be tagged in the git repo.
We can (and we should) trigger build tests on every `master` commit and PR.

samjsong:
cc @Gourav2906 is there anywhere else we may need to update this behavior in the e2e tests? The new behavior will be as described in the doc [here](https://docs.google.com/document/d/1661aNOoMEV3PlMSfXPbz8GqfAD5HwE4o6VsWvAoD1fw/edit)

PR #468, opened 2020-03-05
Bump otelcol and remove unused tags

pmm-sumo:
###### Description

Cleanup the unused tags


deploy/helm/sumologic/values.yaml
```diff
@@ -645,33 +645,27 @@ otelcol:
...
-            - containerImage
             - cluster
             - daemonSetName
             - deployment
             - hostName
             - namespace
-            - namespaceId
             - node
             - owners
-            - podId
```
sumo-drosiek: I think we shouldn't remove this

pmm-sumo: Reintroduced it

PR #471, opened 2020-03-06
change the dynamic namespace env variable to be fluentd namespace

vsinghal13:
###### Description

This PR changes the dynamic namespace env variable to be fluentd namespace for the prometheus remote write URL's.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -78,8 +78,9 @@ prometheus:
             key: release
       - name: NAMESPACE
         valueFrom:
-          fieldRef:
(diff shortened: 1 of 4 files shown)


samjsong:
If Prometheus and Fluentd are deployed on different namespaces, can Prometheus access the configmap that is deployed on the namespace that Fluentd is on?

vsinghal13:
> If Prometheus and Fluentd are deployed on different namespaces, can Prometheus access the configmap that is deployed on the namespace that Fluentd is on?

No, for that we copy over the configmap to the namespace where Prometheus is running. The same is documented in the existing Prometheus Operator steps.

PR #476, opened 2020-03-09
add K8S_NODE_NAME for metadata plugin

vsinghal13:
###### Description

If the name of the Kubernetes node the Kubernetes metadata plugin is running on is set as an environment variable with the name K8S_NODE_NAME, it will reduce cache misses and needless calls to the Kubernetes API.

This PR adds that env variable. 

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/deployment.yaml
+++ b/deploy/helm/sumologic/templates/deployment.yaml
@@ -107,6 +107,10 @@ spec:
         - name: pos-files
           mountPath: /mnt/pos/
         env:
+        - name: K8S_NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
         - name: SUMO_ENDPOINT_METRICS
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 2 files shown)


samjsong:
Hmm, if I understand this correctly it only works if Fluentd is running as a daemonset since it will only get metadata for the resources running on the same node, so I feel like in this case we cannot use this feature since we need metadata for all nodes, right? Please correct me if I'm wrong

vsinghal13:
> Hmm, if I understand this correctly it only works if Fluentd is running as a daemonset since it will only get metadata for the resources running on the same node, so I feel like in this case we cannot use this feature since we need metadata for all nodes, right? Please correct me if I'm wrong

Hmm, That's a good point! The metadata will be restricted to only the specific nodes. I will verify if the plugin has some concept of handling this as no constraints are mentioned in the documentation.  https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#environment-variables-for-kubernetes

vsinghal13:
Closing this as this will only work if fluentd is running as a Daemonset.

PR #481, opened 2020-03-10
Set new release 0.16.0

perk-sumo:
###### Description

This is a minor release for the opentelemetry collector related changes (disabled by default).

###### Testing performed

- [x] ci/build.sh


samjsong:
@perk-sumo we also want to update all of the doc versions so that when the `0.16.0` tag is cut we will point to the correct links in the docs.

perk-sumo:
@samjsong nice catch - let me take care of that!

PR #483, opened 2020-03-10
Add missing source_category_replace_dash to systemd conf.

bboerst:
###### Description

This PR adds source_category_replace_dash to systemd log filters.
[The default setting is "/"](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/fluent-plugin-kubernetes-sumologic/lib/fluent/plugin/filter_kubernetes_sumologic.rb#L11) and we noticed that container logs were honoring our override, but systemd logs were not.

###### Testing performed

- [ ] ci/build.sh


bboerst:
Ref: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/484

PR #492, opened 2020-03-12
Add configuration of fluent workers

sumo-drosiek:
###### Description

Allow to increase number of fluentd workers. It can help with some performance issues

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/common.conf
+++ b/deploy/helm/sumologic/conf/common.conf
@@ -16,8 +16,11 @@
   port 9880
   bind 0.0.0.0
 </source>
-{{- if .Values.fluentd.logLevel }}
 <system>
-  log_level {{ .Values.fluentd.logLevel }}
+  {{- if .Values.fluentd.logLevel }}
+    log_level {{ .Values.fluentd.logLevel }}
+  {{- end }}
+  {{- if .Values.fluentd.workers }}
+    workers {{ .Values.fluentd.workers}}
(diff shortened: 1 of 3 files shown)


sumo-drosiek:
@vsinghal13 @samjsong Do you see any cons for exposing workers to the configuration?

vsinghal13:
Also, we need to verify if all the plugins that we use are compatible with multi-process workers. https://docs.fluentd.org/deployment/multi-process-workers#fluentd-cant-start-with-multi-process-workers-why

sumo-drosiek:
Thank for pointing this out. I will create helm function for buffers to keep everything in one place as we discussed some time ago, and after that I will back to multi-workers :+1:

sumo-drosiek (commented):
Remove auto-generated files due to #1052

frankreno:
@sumo-drosiek - is this PR still relevant/needed?

sumo-drosiek:
@frankreno It can increase the performance of the fluentd instance and after moving the buffer paths to the config.yaml it's easy to achive. I will perform some comparison test

sumo-drosiek:
I updated the PR and tested the change.

So, fluentd with 2 workers and CPU limit increased to 2 works with about same performance like 2 fluentd pods, so there is no performance gain. In addition there were some errors in the plugins which don't occur for single worker instance. After taking above into consideration we decided to skip this feature

PR #511, opened 2020-03-23
change docs for existing prometheus operator to remove sed command

vsinghal13:
###### Description

Starting version `0.17` we do not need the sed command in the case of existing Prometheus operator. 
###### Testing performed

- [x] ci/build.sh


vsinghal13:
> LGTM, do we also need to update this in the 0.17 docs?

Yes, was planning to do it  once this is approved.

PR #541, opened 2020-03-31
Allow for release branches PRs builds

perk-sumo:
###### Description

###### Testing performed

- [ ] ci/build.sh


perk-sumo:
> I suppose we do not want `release-v1.0.0-alpha` and so on branches 👍

Yeah, we don't want that. It should be simple, one branch per release. Alpha/beta/rc builds should be cut from those release branches.

Issue #139, opened 2019-08-12
Add way to add custom gems without having to maintain docker image.


frankreno:
One possible way to do this is via an `init` container.  The user could provide a list of additional gems, the init container would come up and do a gem`gem install` on those.

adaughterson:
+1
We would like the ability to gracefully have additional fluentd plugins installed without having to fork the standard sumo fluentd image (what we are currently doing).

perk-sumo:
@frankreno did #542 solve the issue for you?
Can we close this one?

frankreno:
This does indeed!

PR #542, opened 2020-03-31
Allow to install plugins in docker image

sumo-drosiek:
###### Description

Allow to install plugins basing on values.yaml

Solves #139 for simple plugins (which do not require additional libraries)

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -25,4 +25,11 @@ if [ "$1" = "fluentd" ]; then
     fi
 fi
 
+# Install custom plugins if specified by user
+if [ -n "${ADDITIONAL_PLUGINS}" ]; then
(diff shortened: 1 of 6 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -25,4 +25,13 @@ if [ "$1" = "fluentd" ]; then
     fi
 fi
 
+# Install custom plugins if specified by user
+if [ -n "${ADDITIONAL_PLUGINS}" ]
+then
```
perk-sumo, line 30: Please follow the https://google.github.io/styleguide/shellguide.html#loops
`Put ; do and ; then on the same line as the while, for or if.`

It'll also be consistent throughout the file this way.

sumo-drosiek:
@drduke1 I updated documentation related to code changes. Could you take a look? :)

PR #546, opened 2020-04-01
remove sourceHost from values.yaml

samjsong:
###### Description

Since we don't want to expose sourceHost any longer, removing it from the values.yaml file

###### Testing performed

- [ ] ci/build.sh


perk-sumo:
Merged after conversation with @breinero and @frankreno

PR #548, opened 2020-04-02
Add fluentd to service name to fix urls

samjsong:
###### Description

Found that our e2e tests were failing due to logs and metrics not being sent, while events were still working. I updated the service name to include `fluentd`, tested locally and it seems to be working again

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/events-service-headless.yaml
+++ b/deploy/helm/sumologic/templates/events-service-headless.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ printf "%s-events-headless" (include "sumologic.fullname" .) }}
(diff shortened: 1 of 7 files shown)


perk-sumo:
It looks OK but what about those two files?
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/service-headless.yaml#L4
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/events-service-headless.yaml#L5

samjsong:
That's a good question. Since the logs and metrics were working with the new service name of `collection-sumologic-headless` back with the persistent volume + statefulset before, I assumed this would be the same. Maybe we can choose to add fluentd to the names there just to be clear as well.

PR #549, opened 2020-04-02
add quotes when specifying events sourceCategory

samjsong:
###### Description

When a user specifies a custom `fluentd.events.sourceCategory`, the terraform template treats it as variable and results in an error, so we need to wrap it in quotes so that TF can treat it as a string.

We will need to backport this to `0.17` as well (except the config is `sumologic.events.sourceCategory` in `0.17`), since current customers are running into this bug.

###### Testing performed

- [x] reproduced the issue with provided `values.yaml`
- [x] re-packaged helm chart, reinstalled with helm
- [x] verified the collector was created correctly, with the correct source category for events

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-configmap.yaml
@@ -94,7 +94,7 @@ data:
 
     resource "sumologic_http_source" "events_source" {
         name         = local.events-source-name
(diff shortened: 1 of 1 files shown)


perk-sumo:
Thank you @samjsong !
I've created an backport issue: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/550

sunggun-yu:
+1, @perk-sumo 
I'm currently failing with this issue in 0.17.2

PR #561, opened 2020-04-06
Control plane metrics

sumo-drosiek:
###### Description

Add control plane metrics:

 - coredns
 - etcd

Also extract match section from the metrics configuration to the helm template function

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
I skip etcd in this PR till I'm not really sure which metrics do we expect. We already send some `etcd_` metrics

sumo-drosiek:
I recommend unified view for this PR (because of helm templates 😢 )

PR #565, opened 2020-04-06
expose the k8s api version and groups as params

vsinghal13:
###### Description

###### Testing performed

- [ ] ci/build.sh


deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -85,8 +85,8 @@ data:
         cache_size  "10000"
         cache_ttl  "3600"
         cache_refresh "1800"
-        core_api_versions [v1]
-        api_groups [apps/v1 extensions/v1beta1]
+        core_api_versions "[v1]"
+        api_groups "[apps/v1 extensions/v1beta1]"
```
vsinghal13, line 89: @samjsong do we need quotes? Also, inside the brackets, the two items are with no quotes and  no comma

vsinghal13: cc @perk-sumo

samjsong: that's interesting, can you try a non-helm install and see if this works?

vsinghal13:
Converted to comma-separated list based on : https://docs.fluentd.org/configuration/config-file
```
array type: the field is parsed as a JSON array. It also supports
shorthand syntax. These are same values.
normal: ["key1", "key2"]
shorthand: key1,key2
```

vsinghal13:
> LGTM :) but the `Non_Helm_Installation.md` doc changes seem out of place, is that a result of some merge from master or something?

fixed it

PR #573, opened 2020-04-08
Adjust fluentd names and app labels

perk-sumo:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/service-headless.yaml
+++ b/deploy/helm/sumologic/templates/service-headless.yaml
@@ -3,11 +3,11 @@ kind: Service
 metadata:
   name: {{ template "sumologic.fullname" . }}-fluentd-headless
   labels:
-    app: {{ template "sumologic.labels.app" . }}
+    app: {{ template "sumologic.labels.app" . }}-fluentd-headless
     {{- include "sumologic.labels.common" . | nindent 4 }}
 spec:
   selector:
-    app: {{ template "sumologic.labels.app" . }}
+    app: {{ template "sumologic.labels.app" . }}-fluentd
   clusterIP: None
   ports:
   {{- if .Values.sumologic.traces.enabled }}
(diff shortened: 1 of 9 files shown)


deploy/helm/sumologic/templates/service-headless.yaml
```diff
@@ -3,11 +3,11 @@ kind: Service
 metadata:
   name: {{ template "sumologic.fullname" . }}-fluentd-headless
   labels:
-    app: {{ template "sumologic.labels.app" . }}
+    app: {{ printf "%s-fluentd-headless" (include "sumologic.labels.app" .) }}
```
sumo-drosiek, line 6: I think we should stick to the templates :)
We could create templates for events, fluentd, otelcol and whatever else we have. for both names and labels

perk-sumo: That's a good point but let's do it in 1.1 and not change the 1.0 scope any further.
I've created an issue to keep track of that one: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/574

samjsong:
There's a few other places that we use `collection-sumologic`, for example `ServiceAccount`, `ClusterRole`, `ClusterRoleBinding`
Will we want/need to change these as well? Some of these also reference a label `app=collection-sumologic`

perk-sumo:
Actually that's a very good question @samjsong ! Firstly I've created a massive PR with all of those places refactored. But then when I looked at how many things I needed to change and that I was still not happy with the outcome I'm not sure we'd like to change all of that just before the release.
I propose we do the smallest subset that makes sense for now and do a big refactor for 1.1 (including the #574 )

samjsong:
Hmm, I'm afraid changing the names of those other resources might technically be a breaking change for non-helm users since we'd be leaving old resources behind. The user would have to manually delete the old `ServiceAccount` etc since `kubectl apply` has no concept of deletion

but I don't believe it'll have any impact in terms of functionality, just some leftover resources... if this is okay then I'm +1 and we can proceed with this plan

perk-sumo:
Actually `ServiceAccount` is being used by couple fluentd, events and otelcol. So I think we should leave it as it is. That being said I've updated labels for HPA and ConfigMap as well, please find the second commit.

PR #577, opened 2020-04-09
Add etcd to control plane metrics

sumo-drosiek:
###### Description

Add etcd to control plane metrics

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
What is the reason that etcd metrics from apiserver are specified very detailed? Is it because we want to have strict list of metrics?

PR #581, opened 2020-04-16
Set the default Prometheus scrape interval to 30s explicitly

perk-sumo:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -65,10 +65,14 @@ prometheus:
       matchLabels:
         app: collection-sumologic-otelcol
   prometheusSpec:
+    ## Interval between consecutive scrapes.
+    ##
+    scrapeInterval: "30s"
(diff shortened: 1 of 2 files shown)


sumo-drosiek:
I cannot find information about default value of this variable equal to 30s in the docs, by via the configuration site there is 30s, so :+1:

sumo-drosiek:
I found it in the source code:
https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/promcfg.go#L175

samjsong:
As discussed on Slack, there can be some recording rules (ex: `node:node_cpu_utilisation:avg1m`) that will break if we increase the scrape interval. We should either come up with a way around that (increase the 1m in the recording rule?) or document this

perk-sumo:
Fixed upstream with https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/665

PR #584, opened 2020-04-21
K8S diagnostics image with debugging tools

pmm-sumo:
###### Description

This adds a diagnostics image that is a place to keep various tools helping to solve potential issues with K8S deployment quickly. It starts with K8S API tools but can be easily extended.

###### Testing performed

- [x] ci/build.sh


ci/build.sh
```diff
@@ -3,6 +3,7 @@
 VERSION="${TRAVIS_TAG:-0.0.0}"
 VERSION="${VERSION#v}"
 : "${DOCKER_TAG:=sumologic/kubernetes-fluentd}"
+: "${DIAGNOSTICS_TAG:=sumologic/k8s-diagnostics}"
```
sumo-drosiek, line 6: it should be `kubernetes` to keep convention with fluentd image

diagnostics/README.md
```diff
@@ -0,0 +1,30 @@
...
+Pod namespace env variable:
+Running K8S API test
+2020/04/21 17:08:19 Received data for 15 pods in the cluster
+pod "diag" deleted
+```
+
+The pod can be also run in interactive mode, i.e.:
+
+`$ kubectl run diag --generator=run-pod/v1 -it --rm --restart=Never --image sumologic/kubernetes-diagnostics -- /bin/bash -l`
+
```
sumo-drosiek, line 30: please use one image in both examples :)
`sumologic/kubernetes-diagnostics`

pmm-sumo:
> In general I love this idea and it was something that @abhi-sumo built a protoype CLI for that went a bit farther than this. It analyze the clsuter and determine the optimimal settings to install with. The intention was to reduce the cognitive overload for users. it likely needs some imporvements but we might be able to learn from the concepts of this. - https://github.com/abhi-sumo/sumo-k8s-cli

Thank you @frankreno , I think that @abhi-sumo's project serves a slightly different purpose. If I read the code right, it provides a wizard-like experience for the setup. 

What I am proposing here is more like a standalone image that can be very easily executed to check what's the environment (no installation is even needed). It can be extended in a bunch of ways as needed. I started with making it provide some answers that will help with figuring out why metadata tagging is not working for one of the alpha customers in a convenient way

sumo-drosiek:
Does it make sense to merge both ideas?.
I believe the investigation and installation wizard can coexist in one tooling image :)

I know that this code is pretty urgent right now, but we can make some effort in the future to merge both solutions

PR #586, opened 2020-04-22
Fix filtering metrics for old kubernetes versions

sumo-drosiek:
###### Description

Cast version which have suffixes to the base one before comparison: e.g. `15+` -> `15`

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
It's good candidate for function, but there are only two occurrences so I changed it in place.

PR #588, opened 2020-04-22
split logs + metrics Fluentd sts

samjsong:
###### Description

Splitting Logs and Metrics Fluentd statefulsets for scalability

Also noticed that we had not updated the service monitors for Prometheus in `values.yaml`, updated those to include the `fluentd` name

###### Testing performed

- [ ] ci/build.sh


deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
```diff
@@ -6,7 +6,7 @@
   sumologicCollector:: {
     remoteWriteConfigs+: [
       {
-        url: $._config.sumologicCollectorSvc + "prometheus.metrics.state",
+        url: "http://$(CHART)-metrics.$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.state",
```
sumo-drosiek, line 9: Do the `CHART` and `NAMESPACE` variables are going to be replaced here correctly?
No idea how the libsonnet files works, but this is kubernetes directory

deploy/helm/sumologic/templates/metrics-statefulset.yaml
```diff
@@ -0,0 +1,163 @@
...
+        configMap:
+          name: {{ template "sumologic.fullname" . }}-fluentd-metrics
+      securityContext:
+        fsGroup: {{ .Values.fluentd.securityContext.fsGroup }}
+      containers:
+      - name: fluentd
+        image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        resources:
+          {{- toYaml .Values.fluentd.statefulset.resources | nindent 10 }}
```
sumo-drosiek, line 84: I believe we should create another resources section for the metrics

deploy/kubernetes/kube-prometheus-sumo-logic-mixin.libsonnet
vsinghal13: Yes, this needs to be fixed as the variables won't be replaced.

deploy/helm/sumologic/templates/metrics-hpa.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.fluentd.autoscaling.enabled}}
```
sumo-drosiek, line 1: `.Values.fluentd.autoscaling.metrics.enabled` ?

deploy/helm/prometheus-overrides.yaml
```diff
@@ -19,28 +19,39 @@ prometheusOperator:
     enabled: false
 prometheus:
   additionalServiceMonitors:
-  - name: collection-sumologic
+  - name: collection-sumologic-fluentd
```
frankreno, line 22: I think we should explicitly call out logs, metrics and eventually traces rather than just fluentd alone.

frankreno: Which will require an update to fluent-bit output i believe.

samjsong:
Hi team! I've made the requested changes:
1. move the relevant new configs for metrics under the `fluentd.metrics` key, as well as moving the parallel configs for logs under the `fluentd.logs` key
2. add `-logs` suffix to the logs fluentd name. (However I've left the tracing config in the logs fluentd so the name is a little misleading)

samjsong:
@sumo-drosiek the CI seems to have failed due to something with the `kubernetes-tools` change, would you mind taking a look if you get the chance? thank you! :)

perk-sumo:
@samjsong I've rebased your branch on master so that building `tools` won't break the build.
Also I like the changes you've made, thank you! 👏

PR #590, opened 2020-04-24
Include serviceaccount in the usage description

pmm-sumo:
###### Description

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
Keep in mind that it can be deployment when the serviceaccount is not a `collection-sumologic` as it can be overwritten in configuration

PR #591, opened 2020-04-24
Add template generator to the kubernetes-tools

sumo-drosiek:
###### Description

This should solve the support for non-helm k8s solutions. It doesn't require anything more than customer already have (kubectl is enough to generate the template). It uses helm3 to generate the template

###### Testing performed

- [ ] ci/build.sh


tools/README.md
```diff
@@ -28,6 +28,38 @@ Running K8S API test
...
+  --namespace <NAMESPACE> \
+  --name collection \
+  --set sumologic.accessId='<ACCESS_KEY>' \
+  --set sumologic.accessKey='<ACCESS_ID>' \
+  --set sumologic.collectorName='<COLLECTOR_NAME>' \
+  --set sumologic.clusterName='<CLUSTER_NAME>' \
+  | tee setup-sumologic.yaml
+```
+
+As the result you will get `setup-sumologic.yaml` which is ready to apply for kubernetes.
```
samjsong, line 61: the name of `setup-sumologic.yaml` is already used to refer to the Sumo collector & sources setup using terraform (see under `/deploy/kubernetes/setup-sumologic.yaml.tmpl`)
I think we should change the name of the generated yaml file here, maybe to `sumologic.yaml` or something more generic to encapsulate the fact that this is for the entire chart

tools/template
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+echo '# Add sumologic repository'
+helm repo add sumologic https://sumologic.github.io/sumologic-kubernetes-collection 1>/dev/null 2>&1
+echo '# Update repository'
+helm repo update 1>/dev/null 2>&1
+echo '# Generating template'
+helm template sumologic/sumologic "$@"
```
samjsong, line 8: When we run `helm template` on the sumologic repo, does the generated template also contain the dependency chart contents? or will the user still need to run `helm template` for prometheus, fluent-bit, etc separately?

sumo-drosiek: It contains, but prometheus operator creates crd by itself, what leads to errors during first apply:
```
unable to recognize no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "/sumologic/sumologic.yaml": no matches for kind "Prometheus" in version "monitoring.coreos.com/v1"
unable to recognize "/sumologic/sumologic.yaml": no matches for kind "PrometheusRule" in version "monitoring.coreos.com/v1"
```
but after prometheus operator create the crds(~20s) next apply goes smooth

frankreno: Any chance to ensure the CRDs get applied first to prevent errors?

sumo-drosiek: As it's created by the operator 's pod we cannot. We could take it from running cluster and put into scripts, but I don't feel its safe way and leads to complications

sumo-drosiek: Also I think that is the reason why `helm delete` does not remove prometheus crds

sumo-drosiek: Here is potential solution:
https://github.com/helm/charts/tree/master/stable/prometheus-operator#helm-fails-to-create-crds

We can put the crds definitions on the top of the sumologic.yaml file

sumo-drosiek:
I did research and test for crds and seems that is better to leave it as it is, because objects definitions and validation differs between kubernetes versions, so it's better to leave the crds creation to the prometheus operator itself

sumo-drosiek:
I managed to install this in two steps without error:

step 1 (create prometheus crds):
```bash
10131  kubectl delete crd alertmanagers.monitoring.coreos.com podmonitors.monitoring.coreos.com prometheuses.monitoring.coreos.com prometheusrules.monitoring.coreos.com servicemonitors.monitoring.coreos.com
10132  curl https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.38/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagers.yaml | grep preserveUnknownFields -v | kubectl apply -f -
10133  curl https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.38/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml | grep preserveUnknownFields -v | kubectl apply -f -
10134  curl https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.38/example/prometheus-operator-crd/monitoring.coreos.com_prometheus.yaml | grep preserveUnknownFields -v | kubectl apply -f -
10135  curl https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.38/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml | grep preserveUnknownFields -v | kubectl apply -f -
10136  curl https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.38/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml | grep preserveUnknownFields -v | kubectl apply -f -
10137  curl https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.38/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml | grep preserveUnknownFields -v | kubectl apply -f -
```

step 2 (apply collection templates)
```
10138  kubectl apply -f sumologic.yaml
```

But I had to use `grep preserveUnknownFields -v` because of kubernetes version (1.14)

sumo-drosiek:
Closing due to moving tools to separate repo: https://github.com/SumoLogic/sumologic-kubernetes-tools/issues/4

PR #598, opened 2020-04-28
Add dynamic generation of terraform kubernetes object

sumo-drosiek:
###### Description

Expose terraform's kubernetes section. This is PoC as terraform object is slightly complicated and specific, so requires some tests 

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -11,6 +11,28 @@ sumologic:
   # If enabled, a pre-install hook will create Collector and Sources in Sumo Logic
   setupEnabled: true
 
+  # Configuration of kubernetes for terraform client
+  # https://www.terraform.io/docs/providers/kubernetes/index.html#argument-reference
+  cluster:
+    host: "https://kubernetes.default.svc"
+    # username:
+    # password:
+    # insecure:
+    # client_certificate:
+    # client_key:
+    cluster_ca_certificate: "${file(\"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt\")}"
+    # config_path:
+    # config_context:
+    # config_context_auth_info:
+    # config_context_cluster:
+    token: "${file(\"/var/run/secrets/kubernetes.io/serviceaccount/token\")}"
+    # load_config_file:
+    # exec:
+    #   api_version:
+    #   command:
+    #   args: []
+    #   env: {}
+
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -11,6 +11,28 @@ sumologic:
...
+  # Configuration of kubernetes for terraform client
+  # https://www.terraform.io/docs/providers/kubernetes/index.html#argument-reference
+  cluster:
+    host: "https://kubernetes.default.svc"
+    # username:
+    # password:
+    # insecure:
+    # client_certificate:
+    # client_key:
+    cluster_ca_certificate: "${file(\"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt\")}"
```
frankreno, line 23: Why is the double quote escaped with a backslash but not in the YAML below?

deploy/helm/sumologic/values.yaml
```diff
@@ -11,6 +11,28 @@ sumologic:
...
+    # config_path:
+    # config_context:
+    # config_context_auth_info:
+    # config_context_cluster:
+    token: "${file(\"/var/run/secrets/kubernetes.io/serviceaccount/token\")}"
```
frankreno, line 28: Why is the double quote escaped with a backslash but not in the YAML below?

deploy/helm/sumologic/values.yaml
sumo-drosiek: Terraform format is not friendly:
 * cannot use `'`
 * `"` between `"` cannot be escaped (`\"` is treated as `\"` instead of `"`)

sumo-drosiek: We can use in the `values.yaml`:
 * "${file(\"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt\")}"
 * or  '${file("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")}'

frankreno: AS long as that syntax works in both cases, im ok with them being slightly different :)

sumo-drosiek:
Should this go to 1.0 or 1.1 as there is no breaking changes?
@perk-sumo @frankreno

perk-sumo:
@sumo-drosiek it should go into 1.1.

sumo-drosiek:
Closing in favor of #675

PR #601, opened 2020-05-01
add migration doc

vsinghal13:
###### Description

Please review commit by commit to see the sequence of addressed PR comments. :)

###### Testing performed

- [ ] ci/build.sh


deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+| sumologic.timestampKey 	| fluentd.logs.output.timestampKey 	|
+| sumologic.verifySsl 	| fluentd.verifySsl 	|
+| sumologic.excludeContainerRegex 	| fluentd.logs.containers.excludeContainerRegex 	|
+| sumologic.excludeHostRegex 	| fluentd.logs.containers.excludeHostRegex 	|
+| sumologic.excludeNamespaceRegex 	| fluentd.logs.containers.excludeNamespaceRegex 	|
+| sumologic.excludePodRegex 	| fluentd.logs.containers.excludePodRegex 	|
+| sumologic.fluentdLogLevel 	| fluentd.logLevel 	|
+| sumologic.watchResourceEventsOverrides 	| fluentd.events.watchResourceEventsOverrides 	|
+| sumologic.fluentd.buffer 	| fluentd.buffer.type 	|
+| sumologic.fluentd.autoscaling.* 	| fluentd.autoscaling.* 	|
```
sumo-drosiek, line 53: We have now:
```
fluentd.logs.autoscaling.*
fluentd.metrics.autoscaling.*
```
instead of
```
fluentd.autoscaling.*
```

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed. 
+
+Some use-cases include : 
+ - custom log pipelines, 
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
+
+The Fluentd `deployments` have been changed to `statefulsets` to support the use of persistent volumes. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is not a breaking change for Helm users.
```
sumo-drosiek, line 84: We should mention split of logs/metrics here

sumo-drosiek:
LGTM, but we need to include information about splitting logs and metrics

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+    - [3: Apply new changes](#3-apply-new-changes)
+
+
+<!-- /TOC -->
+
+Based on the feedback from our users, we will be introducing several changes to the Sumo Logic Kubernetes Collection solution. Here we detail the changes for both Helm and Non-Helm users, as well as the exact steps for migration.
+## Helm Users
+### Changes
+
+- Falco installation disabled by Default
```
frankreno, line 24: We also bumped the helm chart version which included a fix to disable the botcoin miner rule by default - https://github.com/falcosecurity/falco/issues/835

frankreno: Can we add a note on how to enable.  also does the migration script respect the setting from the users previous values.yaml? E.g. if they had it enabled in the values.yaml, will it respect that?

deploy/docs/V1_Migration_doc.md
frankreno: +1

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+  logs:
+    containers:
+      extraFilterPluginConf:
+        <filter **>
+          @type record_modifier
+          remove_keys stream, time
+        </filter>
+```
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed.
```
frankreno, line 77: We need a document with some examples of these and should link back to this.  I think we should show some simple examples that illustrate how to use the new configuration options as well as the list of all properties in the values.yaml that were added to enable this.

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+The Fluentd `deployments` have been changed to `statefulsets` to support the use of persistent volumes. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is not a breaking change for Helm users.
+
+
+### How to upgrade
+#### 1. Upgrade to helm chart version `v0.17.1 `
+
+If the user is not already on `v0.17.1` of the helm chart, upgrade to that version first by running the below command.
+
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=0.17.1
```
frankreno, line 93: Need a helm repo update to make sure they have 0.17 and 1.0 to begin with.

frankreno: Might be good to add a step on how to check that things are working ok, as well as a rollback step if needed to revert.

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+Some use-cases include : 
+ - custom log pipelines, 
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
+
+The Fluentd `deployments` have been changed to `statefulsets` to support the use of persistent volumes. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is not a breaking change for Helm users.
+
+
+### How to upgrade
+#### 1. Upgrade to helm chart version `v0.17.1 `
```
frankreno, line 88: Lets clearly state these commands are for helm v2 as well, and remind helm3 is not yet supported since it may change.

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+```
+#### 2: Run upgrade script
+
+For Helm users, the only breaking changes are the renamed config parameters.
+For users who use a `values.yaml` file, we will provide a script that customers can run to convert their existing `values.yaml` files into ones that are compatible with the major release. 
+
+The upgrade script is present in the path : `sumologic-kubernetes-collection/deploy/helm/sumologic/`
+
+Run the upgrade script with the below command.
+```bash
```
frankreno, line 103: They will need to clone the repo :)

frankreno: or curl the file down...

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+./upgrade-1.0.0.sh values.yaml
+```
+At this point, users can then run: 
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=1.0.0 -f new_values.yaml
```
frankreno, line 108: I want to make sure that QE has tested this path? cc @perk-sumo

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+```
+For users who installed without a `values.yaml` file, if they had set any of the above config parameters they will need to re-set those parameters during their upgrade. For example, if the customer had originally run
+```bash
+helm install sumologic/sumologic --name collection --namespace sumologic --set sumologic.accessId=<SUMO_ACCESS_ID> ... --set sumologic.numThreads=16
```
frankreno, line 112: Can we add steps using helm get values? that way customers don't have to remember?

frankreno: e.g helm get values, pipe to file, run upgrade script on file then helm upgrade?

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+- The use of environment variables to set configs has been removed to avoid the extra layer of indirection and confusion. Instead, configs will be set directly within the Fluentd pipeline.
+- `kubernetesMeta` and `kubernetesMetaReduce` have been removed from `logs.kubernetes.sumologic.filter.conf` of the Fluentd pipeline for the same reason as above (Helm users)
+- Similarly `addStream` and `addTime` (default values were `true`) have been removed from `logs.kubernetes.sumologic.filter.conf` of the Fluentd pipeline; the default behavior will remain the same. To preserve the behavior of `addStream = false` or `addTime = false`, you can add: 
+```yaml
+<filter containers.**>
+  @type record_modifier
+  remove_keys stream,time
+</filter>
+```
+above the output plugin section [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L251)
```
frankreno, line 129: This is linked to master, let's fix in case this changes as master changes...once we cut the tag that is.

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+- The Fluentd deployments have been changed to **statefulsets** to support the use of **persistent volumes**. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is a breaking change for non-Helm users as the deployments will not be cleaned up upon upgrade, leading to duplicate events (logs and metrics will not experience data duplication).
```
frankreno, line 130: lets make sure we call out the split.

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+- We now support the collection of renamed metrics (for Kubernetes version `1.17+`).
+### How to upgrade for Non-helm Users
+#### 1. Move environment variable config changes to Fluentd pipeline
+- Non-Helm users who have made changes to configs in the environment variable sections of the `fluentd-sumologic.yaml` file will need to move those config changes directly into the Fluentd pipeline.
```
frankreno, line 134: can we add a link to the 0.17 yaml and highlight the effected env vars?

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+#### 2. Modify filter based on kubernetes version
+- Non-Helm users running a Kubernetes version of 1.13 or older will need to **remove the following filter plugin section from their Fluentd pipeline**
```
frankreno, line 136: Add this is required to prevent data duplication so its clear they should od it :)

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,156 @@
...
+```
+For users who installed without a `values.yaml` file, if they had set any of the above config parameters they will need to re-set those parameters during their upgrade. For example, if the customer had originally run
+```bash
+helm install sumologic/sumologic --name collection --namespace sumologic --set sumologic.accessId=<SUMO_ACCESS_ID> ... --set sumologic.numThreads=16
+```
+Then their upgrade command will look like
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=1.0.0 --set fluentd.buffer.numThreads=16
+```
+## Non-Helm Users
```
frankreno, line 118: Do these steps assume 0.17.1? What happens if they are on an older version?

deploy/docs/V1_Migration_doc.md
frankreno, line 84: Do the docs that we added in the past for persistence require any updating?

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,166 @@
...
+
+<!-- /TOC -->
+
+Based on the feedback from our users, we will be introducing several changes to the Sumo Logic Kubernetes Collection solution. Here we detail the changes for both Helm and Non-Helm users, as well as the exact steps for migration.
+## Helm Users
+### Changes
+
+- Falco installation disabled by Default
+
+- Bumped the helm Falco chart version which included a fix to disable the bitcoin miner rule by default
```
abhi-sumo, line 26: @frankreno Can you actually demarcate what version you bumped the falco dependency to?

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,166 @@
...
+          remove_keys stream, time
+        </filter>
+```
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed. 
+
+Some use-cases include : 
+ - custom log pipelines, 
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
```
abhi-sumo, line 84: Can this be used to trim metrics as well? If so, would be great to document that.

deploy/docs/V1_Migration_doc.md
abhi-sumo: Could you maybe call out which release in the future will have helm3 support? I think it might be important to help customers understand when its coming.

deploy/docs/V1_Migration_doc.md
abhi-sumo: +1 on the check and rollback. I can see this going wrong for a lot of customers potentially.

deploy/docs/V1_Migration_doc.md
vsinghal13: its mentioned as Step1 to upgrade to version 0.17.1 ```1. Upgrade to helm chart version `v0.17.1 ````

deploy/docs/V1_Migration_doc.md
vsinghal13: done

deploy/docs/V1_Migration_doc.md
vsinghal13: cc @frankreno @perk-sumo

deploy/docs/V1_Migration_doc.md
vsinghal13: Helm automatically performs certain checks for upgrade and fails if something is wrong. There are no additional checks that we can perform during the helm upgrade.  Will add `helm rollback` command to roll back to previously installed version of the helm chart.

deploy/docs/V1_Migration_doc.md
vsinghal13: will create a new document with examples.

deploy/docs/V1_Migration_doc.md
vsinghal13: falco version `1.1.6` . Updated in the doc.

deploy/docs/V1_Migration_doc.md
vsinghal13: https://github.com/helm/charts/blob/master/stable/falco/rules/falco_rules.yaml#L2652 @frankreno how will we enable this? In the `values.yaml` there is only an option to just specify the rules file path.

frankreno: On the enable, i just mean that when a user upgrades, they previously had falco.enabled=true.  Now they have falco.enabled=false.  In upgrade path, we should keep it enabled, not disable.  It should only disabled for new installs as if upgrading and it was enabled it is working most likely.  If the user set falco.enabled=false, we should keep it disabled.  Sorry I was not very clear the first time.That is what this comment was about.

vsinghal13: ok got it, thanks. cc @samjsong to add this check in the upgrade script.

vsinghal13:
@frankreno @perk-sumo @samjsong @sumo-drosiek Have modified the Non-helm installation steps to delete the existing resources and redeploy all three, `fluentd`, `fluent-bit` and `prometheus` as we have breaking changes in all the three `yamls`.

deploy/docs/V1_Migration_doc.md
vsinghal13: The `helm get values` lists all the default properties for the subcharts as well. This might be confusing for the user. Also, we will need to test this path.

deploy/docs/V1_Migration_doc.md
vsinghal13: changed to `release-v1.0.0`

deploy/docs/V1_Migration_doc.md
frankreno: Better ways to do that...whcih we will docuemnt

deploy/docs/V1_Migration_doc.md
frankreno: We do not have the exact date yet, but can put something together in an issue once we do.  For now, we can omit.

deploy/docs/V1_Migration_doc.md
frankreno: Bump on this @perk-sumo , should make sure it's in the test plan.

deploy/docs/V1_Migration_doc.md
frankreno: As we discussed, the better approach here will be

get values used in the current release (needs confirmation in helm2) -> `helm get values -o yaml`
Use this as the base to iterate through and map to changes.
The output should contain only properties from original install list

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+  enabled: true
+  ```
+
+- Bumped the helm Falco chart version to v `1.1.6` which included a fix to disable the bitcoin/crypto miner rule by default
+
+- Changes in Configuration Parameters
+	- The `values.yaml` file has had several configs moved and renamed to improve usability. Namely, we introduced a new fluentd section into which we moved all of the Fluentd specific configs, while configs for our dependency charts (`prometheus-operator`, `fluent-bit`, `metrics-server`, `falco`) have not changed.
+
+| Old Config 	| New Config 	|
+|:----------------------------------------:	|-----------------------------------------------------	|
```
sumo-drosiek, line 39: center alignment in one column and left in the second don't look well together in my opinion

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+
+### How to upgrade
+**Note: The below steps are using Helm 2. Helm 3 is not supported.**
+#### 1. Upgrade to helm chart version `v0.17.1 `
+
+Run the below command to fetch the latest helm chart:
+```bash
+helm repo update
+
+If the user is not already on `v0.17.1` of the helm chart, upgrade to that version first by running the below command.
```
samjsong, line 105: might be worth changing this to `0.17.3` now since that is the latest version

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=0.17.1
+```
+#### 2: Run upgrade script
+
+For Helm users, the only breaking changes are the renamed config parameters.
+For users who use a `values.yaml` file, we will provide a script that customers can run to convert their existing `values.yaml` files into ones that are compatible with the major release.
```
samjsong, line 113: nit: we can change `customers` to `users` to stay consistent with wording

samjsong: nit: `we provide` rather than `we will provide`
nit: singular for values.yaml files, so
```
... convert their existing `values.yaml` file into one that is compatible with ...
```

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=0.17.1
+```
+#### 2: Run upgrade script
+
+For Helm users, the only breaking changes are the renamed config parameters.
+For users who use a `values.yaml` file, we will provide a script that customers can run to convert their existing `values.yaml` files into ones that are compatible with the major release. 
+
+The upgrade script is present in the path : `sumologic-kubernetes-collection/deploy/helm/sumologic/`
+You will either need to clone the repo or `curl` the upgrade script.
```
samjsong, line 116: I would recommend removing these lines (115-116) and instead just provide the curl command, like
```
curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/release-v1.0.0/deploy/helm/sumologic/upgrade-1.0.0.sh
```

deploy/docs/V1_Migration_doc.md
samjsong: Now that the upgrade script takes only overriden configs and rewrites them in the `new_values.yaml` file, passing in `--reuse-values` here will fail the upgrade (similar to the recent escalation we had). We will need to have users run the upgrade without this flag, potentially they will need to specify flags like access id and key again.
cc @frankreno @perk-sumo @sumo-drosiek

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=1.0.0 -f new_values.yaml
+```
+For users who installed without a `values.yaml` file, if they had set any of the above config parameters they will need to re-set those parameters during their upgrade. For example, if the customer had originally run
+```bash
+helm install sumologic/sumologic --name collection --namespace sumologic --set sumologic.accessId=<SUMO_ACCESS_ID> ... --set sumologic.numThreads=16
+```
+Then their upgrade command will look like
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=1.0.0 --set fluentd.buffer.numThreads=16
```
samjsong, line 131: Again, `--reuse-values` will not work here since the `values.yaml` file has changed.

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+```
+above the output plugin section [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/release-v1.0.0/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L251)
+- The Fluentd deployments have been changed to **statefulsets** to support the use of **persistent volumes**. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is a breaking change for non-Helm users as the deployments will not be cleaned up upon upgrade, leading to duplicate events (logs and metrics will not experience data duplication).
+- The unified Fluentd `statefulsets` have been split into set of two different Fluentd's, one for `logs` and the other one for `metrics`.
+- We now support the collection of renamed metrics (for Kubernetes version `1.17+`).
+### How to upgrade for Non-helm Users
+#### 1. Tear down existing fluentd, prometheus, fluentbit resources
+Run the commands mentioned [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/Non_Helm_Installation.md#tear-down) to delete the respective resources.
+
+#### 2. Deploy Fluentd, Fluent-bit and prometheus again with the verison 1.0.0 yaml
```
samjsong, line 152: typo: `verison` -> `version`
nit: `prometheus` -> `Prometheus`

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+  @type record_modifier
+  remove_keys stream,time
+</filter>
+```
+above the output plugin section [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/release-v1.0.0/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L251)
+- The Fluentd deployments have been changed to **statefulsets** to support the use of **persistent volumes**. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is a breaking change for non-Helm users as the deployments will not be cleaned up upon upgrade, leading to duplicate events (logs and metrics will not experience data duplication).
+- The unified Fluentd `statefulsets` have been split into set of two different Fluentd's, one for `logs` and the other one for `metrics`.
+- We now support the collection of renamed metrics (for Kubernetes version `1.17+`).
+### How to upgrade for Non-helm Users
+#### 1. Tear down existing fluentd, prometheus, fluentbit resources
```
samjsong, line 149: In general I think we should capitalize Fluentd Prometheus and FluentBit in our docs

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,177 @@
...
+Run the commands mentioned [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/Non_Helm_Installation.md#tear-down) to delete the respective resources.
+
+#### 2. Deploy Fluentd, Fluent-bit and prometheus again with the verison 1.0.0 yaml
+Follow the below steps to deploy new resources. 
+##### 2.1 Deploy Fluentd
+- Non-Helm users who have made changes to configs in the [environment variable sections](#https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/release-v0.17/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L627-L678) of the `fluentd-sumologic.yaml` file will need to move those config changes directly into the Fluentd pipeline.
+
+- Run the below command to get the `fluentd-sumologic.yaml` manifest for version v. `1.0.0` and  then make the changes identified in the above step.
```
samjsong, line 157: nit: do we want to keep the `v.` outside of `v1.0.0`? Not sure here 😛

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,166 @@
...
+- The unified Fluentd `statefulsets` have been split into set of two different Fluentd's, one for `logs` and the other one for `metrics`.
+- We now support the collection of renamed metrics (for Kubernetes version `1.17+`).
+### How to upgrade for Non-helm Users
+#### 1. Move environment variable config changes to Fluentd pipeline
+- Non-Helm users who have made changes to configs in the [environment variable sections](#https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/release-v0.17/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L627-L678) of the `fluentd-sumologic.yaml` file will need to move those config changes directly into the Fluentd pipeline.
+
+#### 2. Modify filter based on kubernetes version
+- Non-Helm users running a Kubernetes version of 1.13 or older will need to **remove the following filter plugin section from their Fluentd pipeline. This is required to prevent data duplication.**
```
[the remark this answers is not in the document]
samjsong: @frankreno the steps above this show how to "get the `fluentd-sumologic.yaml` manifest for version v. `1.0.0`", so hopefully users going in order will have their new YAML by this point

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,166 @@
...
+  @type record_modifier
+  remove_keys stream,time
+</filter>
+```
+above the output plugin section [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/kubernetes/fluentd-sumologic.yaml.tmpl#L251)
+- The Fluentd deployments have been changed to **statefulsets** to support the use of **persistent volumes**. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is a breaking change for non-Helm users as the deployments will not be cleaned up upon upgrade, leading to duplicate events (logs and metrics will not experience data duplication).
+- The unified Fluentd `statefulsets` have been split into set of two different Fluentd's, one for `logs` and the other one for `metrics`.
+- We now support the collection of renamed metrics (for Kubernetes version `1.17+`).
+### How to upgrade for Non-helm Users
+#### 1. Move environment variable config changes to Fluentd pipeline
```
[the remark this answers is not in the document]
vsinghal13: #625 captures this

vsinghal13:
Thanks guys for reviewing the PR and providing recommendations. I have made the requested changes and ready to merge.

Major Change: Upgrade path for Helm Users

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,176 @@
...
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
+
+The Fluentd `deployments` have been changed to `statefulsets` to support the use of persistent volumes. This will allow better buffering behavior. They also now include `“fluentd”` in their names. This is not a breaking change for Helm users.
+
+The unified Fluentd `statefulsets` have been split into set of two different Fluentd's, one for `logs` and the other one for `metrics`.
+
+### How to upgrade
+**Note: The below steps are using Helm 2. Helm 3 is not supported.**
+#### 1. Upgrade to helm chart version `v0.17.1 `
```
samjsong, line 99: `v0.17.1` -> `v0.17.3`

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,176 @@
...
+
+Run the below command to fetch the latest helm chart:
+```bash
+helm repo update
+
+If the user is not already on `v0.17.3` of the helm chart, upgrade to that version first by running the below command.
+
+```bash
+helm upgrade collection sumologic/sumologic --reuse-values --version=0.17.1
```
samjsong, line 108: `v0.17.1` -> `v0.17.3`

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,176 @@
...
+| sumologic.fluentdLogLevel 	| fluentd.logLevel 	|
+| sumologic.watchResourceEventsOverrides 	| fluentd.events.watchResourceEventsOverrides 	|
+| sumologic.fluentd.buffer 	| fluentd.buffer.type 	|
+| sumologic.fluentd.autoscaling.* 	| fluentd.logs.autoscaling.* , fluentd.metrics.autoscaling.* 	|
+| sumologic.k8sMetadataFilter.watch 	| fluentd.logs.containers.k8sMetadataFilter.watch 	|
+| sumologic.k8sMetadataFilter.verifySsl 	| fluentd.logs.containers.k8sMetadataFilter.verifySsl 	|
+| sumologic.k8sMetadataFilter.cacheSize 	| fluentd.metadata.cacheSize 	|
+| sumologic.k8sMetadataFilter.cacheTtl 	| fluentd.metadata.cacheTtl 	|
+| sumologic.k8sMetadataFilter.cacheRefresh 	| fluentd.metadata.cacheRefresh 	|
+| deployment.* 	| fluentd.statefulset.* 	|
```
samjsong, line 68: this config has been split to `fluentd.logs.statefulset.* , fluentd.metrics.statefulset.* `

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,176 @@
...
+  logs:
+    containers:
+      extraFilterPluginConf:
+        <filter **>
+          @type record_modifier
+          remove_keys stream, time
+        </filter>
+```
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed.
```
perk-sumo, line 86: Could we link the now added doc on how to configure it?

deploy/docs/V1_Migration_doc.md
```diff
@@ -0,0 +1,176 @@
...
+For users who use a `values.yaml` file, we provide a script that users can run to convert their existing `values.yaml` file into one that is compatible with the major release. 
+
+
+- Get the existing values for the helm chart and store it as `current_values.yaml` with the below command:
+```bash
+helm get values <RELEASE-NAME> > current_values.yaml
+```
+- Run `curl` the upgrade script as follows:
+```bash
+curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/release-v1.0.0/deploy/helm/sumologic/upgrade-1.0.0.sh
```
perk-sumo, line 122: The release name is `release-v1.0` without the patch number; so the proper link is:
```
https://raw.githubusercontent.com/SumoLogic/sumologic-kubernetes-collection/release-v1.0/deploy/helm/sumologic/upgrade-1.0.0.sh
```

vsinghal13:
Closing this as somehow this has a lot of commits due to merge conflicts while rebasing. Will create a new PR

PR #604, opened 2020-05-04
Fix fluentd url for tracing

sumo-drosiek:
###### Description

Fix for version 1.0

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
+++ b/deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
@@ -1,4 +1,4 @@
 {{- $yamlFile := toYaml $.Values.otelcol.config }}
 {{- $fullname := include "sumologic.fullname" . }}
-{{- $url := list "http://" $fullname ":9411/api/v2/spans" | join "" }}
+{{- $url := list "http://" $fullname "-fluentd-logs:9411/api/v2/spans" | join "" }}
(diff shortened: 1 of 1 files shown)


deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -1,4 +1,4 @@
 {{- $yamlFile := toYaml $.Values.otelcol.config }}
 {{- $fullname := include "sumologic.fullname" . }}
-{{- $url := list "http://" $fullname ":9411/api/v2/spans" | join "" }}
+{{- $url := list "http://" $fullname "-fluentd-logs:9411/api/v2/spans" | join "" }}
```
pmm-sumo, line 3: So this appends `-fluentd-logs` to the url and the template is actually being used in both fluentd and otelcol config, right?

sumo-drosiek: We split metrics and logs in #588 so changed the service names as well.
We are using `-fluentd-logs`  and `-fluentd-metrics` suffixed services and tracing is going to use the logs one

perk-sumo: If we use fluentd for traces in the future it will probably get it's own fluentd.

PR #605, opened 2020-05-04
Update upgrade script to fix prometheus config

sumo-drosiek:
###### Description

Update upgrade script to be suit with #588 changes

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/upgrade-1.0.0.sh
+++ b/deploy/helm/sumologic/upgrade-1.0.0.sh
@@ -8,7 +8,7 @@ and return one that is compatible with v1.0.0.
 
 Requirements:
   yq (3.2.1) https://github.com/mikefarah/yq/releases/tag/3.2.1
-  git diff in case of changes to Prometheus remote write regexs
+  git diff in case of changes to Prometheus remote write regexes
 
 Usage:
   ./upgrade-1.0.0.sh /path/to/values.yaml
@@ -157,7 +157,10 @@ sumologic.kubernetesMetaReduce
 sumologic.addStream
 sumologic.addTime
 deployment
-eventsDeployment"
+eventsDeployment
+fluentd.statefulset
+fluentd.autoscaling
+fluentd.rawConfig"
 
 IFS=$'\n' read -r -d '' -a OLD_CONFIGS <<< "$OLD_CONFIGS"
 IFS=$'\n' read -r -d '' -a NEW_CONFIGS <<< "$NEW_CONFIGS"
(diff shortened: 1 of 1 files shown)


deploy/helm/sumologic/upgrade-1.0.0.sh
```diff
@@ -312,6 +312,70 @@ for i in $(seq 0 ${metrics_length}); do
...
+          matchNames:
+          - sumologic
+        selector:
+          matchLabels:
+            app: collection-sumologic-fluentd-metrics' | yq m -a -i "new.yaml" -
+service_mon_length="$(yq r -l "new.yaml" 'prometheus-operator.prometheus.additionalServiceMonitors')"
+service_mon_length="$(( ${service_mon_length} - 1))"
+for i in $(seq 0 ${service_mon_length}); do
+  service_mon_name="$(yq r "new.yaml" "prometheus-operator.prometheus.additionalServiceMonitors[${i}].name")"
+  if [[ "${service_mon_name}" = "collection-sumologic-events" ]]; then
```
samjsong, line 370: I don't understand why we have a special case for events servicemonitor, is it not possible to handle with the delete or merge you did above?

deploy/helm/sumologic/upgrade-1.0.0.sh
```diff
@@ -312,6 +312,70 @@ for i in $(seq 0 ${metrics_length}); do
...
+for i in $(seq 0 ${fb_length}); do
+  env_name="$(yq r "new.yaml" "fluent-bit.env[${i}].name")"
+  if [[ "${env_name}" = "CHART" ]]; then
+    yq w -i new.yaml "fluent-bit.env[${i}].valueFrom.configMapKeyRef.key" "fluentdLogs"
+  fi
+done
+
+# Fix prometheus service monitors
+yq d -i "new.yaml" "prometheus-operator.prometheus.additionalServiceMonitors(name==collection-sumologic)"
+yq d -i "new.yaml" "prometheus-operator.prometheus.additionalServiceMonitors(name==collection-sumologic-otelcol)"
```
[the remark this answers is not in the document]
sumo-drosiek: For some reason, the otelcol is generated twice. It was easier to remove it than spend time on investigation. I think it's not an issue :)

deploy/helm/sumologic/upgrade-1.0.0.sh
sumo-drosiek: I started from that approach, and later I get into two exceptions:
 * splitting collection-sumologic on logs and metrics
 * strange behavior of otelcol which was duplicated after merge

sumo-drosiek:
I fixed due to your doubts @samjsong.
I would be grateful for additional tests. After pulling latest master I'm not sure if I did all changes (those so far are correct anyway)

deploy/helm/sumologic/upgrade-1.0.0.sh
```diff
@@ -312,6 +315,64 @@ for i in $(seq 0 ${metrics_length}); do
...
+          - port: metrics
+        namespaceSelector:
+          matchNames:
+            - sumologic
+        selector:
+          matchLabels:
+            app: collection-sumologic-otelcol
+      - name: collection-sumologic-fluentd-logs
+        additionalLabels:
+          app: collection-sumologic-fluentd-logs
```
frankreno, line 342: These are dynamically generated no? based on the release and chart name the user uses?

frankreno: The labels and the namespace may change depending on where user installed?

perk-sumo: Yeah, you're right, we need to fix that
The simplest way is to ask the user to provide a namespace and helm release names as a script parameters I believe.

sumo-drosiek: Thats true 🤦

samjsong: however @frankreno these names are currently hardcoded in the `values.yaml` file for the prometheus service monitors, we might want to change it there if we're going to change it here

perk-sumo: I've updated the PR with commit for non-default release and namespace values; @samjsong wouldn't this work with hardcoded values as well?
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/605/files#diff-9d829ead41a2d6b7ad9d8bd01725aedbR328

PR #618, opened 2020-05-12
Add tests for upgrade scripts

sumo-drosiek:
###### Description

Add tests for upgrade scripts

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
Should I skip log comparison maybe?
@perk-sumo

perk:
`log comparison` is a very nice feature, let's keep it :)

tests/upgrade_script/static/falco_disabled.log
```diff
@@ -0,0 +1,8 @@
+
+[INFO]    falco will be disabled. Change "falco.enabled" to "true" if you want to enable it
+
+[INFO]    Replacing tail-db/tail-containers-state.db to tail-db/tail-containers-state-sumo.db
```
perk-sumo, line 4: Hm, that doesn't seem right 🤔 Why we do something with fluentbit when only falco has been changed?

tests/upgrade_script/static/basic.log
```diff
@@ -0,0 +1,8 @@
+
+[INFO]    falco will be enabled. Change "falco.enabled" to "false" if you want to disable it (default for 1.0)
+
+[INFO]    Replacing tail-db/tail-containers-state.db to tail-db/tail-containers-state-sumo.db
```
perk-sumo, line 4: Hm I'm not sure we should output anything about fluentbit if there were no changes for it 🤔

tests/upgrade_script/static/prometheus_service_monitors.log
```diff
@@ -0,0 +1,9 @@
+
+[INFO]    falco will be enabled. Change "falco.enabled" to "false" if you want to disable it (default for 1.0)
```
perk-sumo, line 2: This is not true, falco will be disabled in such case.

deploy/helm/sumologic/upgrade-1.0.0.sh
```diff
@@ -396,25 +396,30 @@ function check_user_image() {
...
+  warning 'Please ensure that new fluent-bit configuration is correct'
 
-  sed 's?tail-db/tail-containers-state.db?tail-db/tail-containers-state-sumo.db?g' ${TEMP_FILE} | \
-  sed "s/'{}'/{}/g" > new_values.yaml
+  sed -i '' 's?tail-db/tail-containers-state.db?tail-db/tail-containers-state-sumo.db?g' ${TEMP_FILE}
 }
 
-function remove_temp_file() {
-  rm ${TEMP_FILE}
+function fix_yq() {
```
perk-sumo, line 412: 👍

tests/upgrade_script/static/key_mapping_empty.output.yaml
```diff
@@ -0,0 +1 @@
+{fluentd: {}, sumologic: {}}
```
perk-sumo, line 1: @sumo-drosiek you sure that's correct? Doesn't look like 🤔

sumo-drosiek: Yes, it's valid yaml. Do you want to say that we should check all keys and remove them if they are empty objects?
What if customer have empty object in it's yaml file?

perk-sumo: You are right, thank you :)

tests/upgrade_script/static/key_mapping_empty.log
```diff
@@ -0,0 +1,22 @@
+[INFO]    Mapping deployment into:
+[INFO]    Removing deployment
+[INFO]    Removing eventsDeployment
+[INFO]    Mapping fluentd.autoscaling into:
```
perk-sumo, line 4: What happened to this `autoscaling`?

sumo-drosiek: It should be object instead of string, but it's not the case of this test

PR #625, opened 2020-05-13
v1 fluentd config examples

vsinghal13:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/docs/v1_conf_examples.md
+++ b/deploy/docs/v1_conf_examples.md
@@ -47,9 +47,9 @@ fluentd:
 
 ### Adding Fluentd Filter plugin
 
-You can add any of the Fluentd filter plugins in the `extraFilterPluginConf` section to filter data as per your needs.
+You can add any of the Fluentd filter plugins in the `extraFilterPluginConf` section to filter data based on your needs.
 
-**Note:** This is specific to container logs pipeline only and will not work for other logs.
+**Note:** This is specific to the container logs pipeline only and will not work for other logs.
 
 The below example uses the `grep` filter to match any record that satisfies the following conditions:
  - The value of the "message" field contains "cool"
--- a/deploy/docs/v1_conf_examples.md
+++ b/deploy/docs/v1_conf_examples.md
@@ -1,0 +1,130 @@
+# Configuring Fluentd
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed. 
+
+Some use-cases include : 
+ - custom log pipelines, 
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
+ - additinal configurantion for sumologic output plugin
+
+
+Below you can see a few examples of how this configuration can be set.
+
+### Custom Log Pipelines
+Now we have exposed an `extraLogs` paramter inside the `logs.containers` section of the `values.yaml` where you can add the output plugin for the custom log pipeline.
+
+```yaml
+fluentd:
+  logs:
+    extraLogs: |-
+      <match custom.tag>
+        @type sumologic
+        @id sumologic.endpoint.custom.log.endpoint
+        source_category k8s/custom
+        source_name custom
+        data_type logs
+        log_key log
+        endpoint "#{ENV['SUMO_ENDPOINT_CUSTOM_LOG']}"
+        proxy_uri "#{ENV['PROXY_URI']}"
+        verify_ssl "#{ENV['VERIFY_SSL']}"
+        log_format "json_merge"
+        add_timestamp "false"
+        <buffer>
+          @type memory
+          @include buffer.output.conf
+        </buffer>
+      </match>
+```
+
+### Adding Fluentd Filter plugin
+
+You can add any of the fluentd filter plugins in the `extraFilterPluginConf` section to filter data as per your needs.
+
+The below example uses the `grep` filter to match any event that satisfies the following conditions:
+ - The value of the "message" field contains "cool"
+ - The value of the "message" field does NOT contain "uncool"
+
+```yaml
+fluentd:
+  logs:
+    containers:
+      extraFilterPluginConf: |-
+        <filter containers.**>
+          @type grep
+          <regexp>
+            key message
+            pattern /cool/
+          </regexp>
+          <exclude>
+            key message
+            pattern /uncool/
+          </exclude> 
+        </filter>
+```
+Reference documentation: [Fluentd Filter Plugin](#https://docs.fluentd.org/filter)
+
+## Override Fluentd Output plugin to forward data to Sumo as well as S3
+The example below shows how you can override the entire output section for the container logs pipeline.
+
+```yaml
+fluentd:
+  logs:
+    containers:
+      overrideOutputConf: |-
+        <match containers.**>
+          @type copy
+          <store>
+            @type sumologic
+            @id sumologic.endpoint.logs
+            @include logs.output.conf
+            <buffer>
+              @type file
+              path /fluentd/buffer/logs.containers
+              @include buffer.output.conf
+            </buffer>
+          </store>
+          <store>
+            @type s3
+            aws_key_id YOUR_AWS_KEY_ID
+            aws_sec_key YOUR_AWS_SECRET_KEY
+            s3_bucket YOUR_S3_BUCKET_NAME
+            s3_region YOUR_S3_BUCKET_REGION
+            path logs/
+            # if you want to use ${tag} or %Y/%m/%d/ like syntax in path / s3_object_key_format,
+            # need to specify tag for ${tag} and time for %Y/%m/%d in <buffer> argument.
+            <buffer tag,time>
+              @type file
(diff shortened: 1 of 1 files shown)


deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+ - custom log pipelines, 
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
+ - additinal configurantion for sumologic output plugin
+
+
+Below you can see a few examples of how this configuration can be set.
+
+### Custom Log Pipelines
+Now we have exposed an `extraLogs` paramter inside the `logs.containers` section of the `values.yaml` where you can add the output plugin for the custom log pipeline.
```
samjsong, line 15: This Fluentd config is enough assuming FluentBit is sending the correct logs, we might want to mention that here, that this will only send the logs to Sumo if the logs are being collected correctly at the FluentBit level with an input plugin.

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+        <buffer>
+          @type memory
+          @include buffer.output.conf
+        </buffer>
+      </match>
+```
+
+### Adding Fluentd Filter plugin
+
+You can add any of the fluentd filter plugins in the `extraFilterPluginConf` section to filter data as per your needs.
```
samjsong, line 42: nit: `fluentd` -> `Fluentd`

samjsong: We'll want to point out that this `extraFilterPluginConf` is specific to the container logs pipeline, it will not work for other logs

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+
+The below example uses the `grep` filter to match any event that satisfies the following conditions:
```
samjsong, line 44: Let's be careful with the use of the word `event` since that's a separate concept in Kubernetes. We can instead use `log` or `record` here

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+              timekey_use_utc true # use utc
+              chunk_limit_size 256m
+            </buffer>
+          </store>
+        </match>
+```
+
+### Additional Buffer/Flush/Retry Config parameters for Sumologic Output Plugin
+
+The following confgig parameters are set by default and their values can be set by changing the respective cofig in `values.yaml`.
```
samjsong, line 110: nit: `confgig` -> `config` and `cofig` -> `config`

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+          <exclude>
+            key message
+            pattern /uncool/
+          </exclude> 
+        </filter>
+```
+Reference documentation: [Fluentd Filter Plugin](#https://docs.fluentd.org/filter)
+
+## Override Fluentd Output plugin to forward data to Sumo as well as S3
+The example below shows how you can override the entire output section for the container logs pipeline.
```
samjsong, line 68: Somewhere in this doc (maybe at the very top of the doc, as a note? or maybe within each of these sections? I'm not sure...) we should mention the fact that the Helm templating like
```
{{ .Values.fluentd.foo.bar | quote }}
```
will NOT work if they are specified within one of these plugin conf sections, since they are in the `values.yaml` file and are therefore interpreted as literal strings

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+```bash
+compress gzip
+flush_interval {{ .Values.fluentd.buffer.flushInterval | quote }}
+flush_thread_count {{ .Values.fluentd.buffer.numThreads | quote }}
+chunk_limit_size {{ .Values.fluentd.buffer.chunkLimitSize | quote }}
+total_limit_size {{ .Values.fluentd.buffer.totalLimitSize | quote }}
+queued_chunks_limit_size {{ .Values.fluentd.buffer.queueChunkLimitSize | quote }}
+overflow_action drop_oldest_chunk
+```
+If you want to add any new Buffer/Flush/Retry Config, you can add them in the `extraConf` section.
```
samjsong, line 121: To help distinguish between the above configs which are controlled with parameters in `values.yaml`, and the ability to add other configs through this new `extraConf` section, I think we could tweak this line to something along the lines of
```
However, if you wish to add any additional Buffer/Flush/Retry configs, you can do so in the new `extraConf` section.
```

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
+# Configuring Fluentd
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed. 
+
+Some use-cases include :
```
sumo-drosiek, line 5: `include :` -> `include:`

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
+# Configuring Fluentd
+
+Until now, Helm users have not been able to modify their Fluentd configuration outside of the specific parameters that we exposed in the `values.yaml` file. Now, we expose the ability to modify the Fluentd configuration as needed. 
+
+Some use-cases include : 
+ - custom log pipelines, 
+ - adding Fluentd filter plugins (ex: fluentd throttle plugin), or 
+ - adding Fluentd output plugins (ex: forward to both Sumo and S3)
+ - additinal configurantion for sumologic output plugin
```
sumo-drosiek, line 9: `additinal configurantion` -> `additional configuration`

deploy/docs/v1_conf_examples.md
sumo-drosiek, line 15: `paramter` -> `parameter`

deploy/docs/v1_conf_examples.md
```diff
@@ -0,0 +1,130 @@
...
+  logs:
+    extraLogs: |-
+      <match custom.tag>
+        @type sumologic
+        @id sumologic.endpoint.custom.log.endpoint
+        source_category k8s/custom
+        source_name custom
+        data_type logs
+        log_key log
+        endpoint "#{ENV['SUMO_ENDPOINT_CUSTOM_LOG']}"
```
sumo-drosiek, line 28: I would be nice to mention/link how to add custom endpoint

deploy/docs/v1_conf_examples.md
```diff
@@ -2,17 +2,25 @@
...
+will **NOT** work if they are specified within one of the below mentioned plugin conf sections, since they are in the `values.yaml` file and are therefore interpreted as literal strings. 
 
 Below you can see a few examples of how this configuration can be set.
 
 ### Custom Log Pipelines
-Now we have exposed an `extraLogs` paramter inside the `logs.containers` section of the `values.yaml` where you can add the output plugin for the custom log pipeline.
+Now we have exposed an `extraLogs` parameter inside the `logs.containers` section of the `values.yaml` where you can add the output plugin for the custom log pipeline.
+
+**NOTE:** This will only send the logs to Sumo if the logs are being collected correctly at the FluentBit level with an input plugin.
+To add a custom endpoint, please follow the steps mentioned [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/docs/additional_prometheus_configuration.md#create-a-new-http-source-in-sumo-logic)
```
frankreno, line 23: Maybe we should link to the Help docs instead, this might be confusing since we are talking about Prometheus in this section tey might think its for metrics and not for logs?

deploy/docs/v1_conf_examples.md
```diff
@@ -39,9 +47,11 @@ fluentd:
 
 ### Adding Fluentd Filter plugin
 
-You can add any of the fluentd filter plugins in the `extraFilterPluginConf` section to filter data as per your needs.
+You can add any of the Fluentd filter plugins in the `extraFilterPluginConf` section to filter data as per your needs.
+
+**Note:** This is specific to container logs pipeline only and will mnot work for other logs.
```
frankreno, line 52: nnit: mnot ->not

deploy/docs/v1_conf_examples.md
frankreno: Or maybe we just need a more detailed explanation not sure what the best approach is.

deploy/docs/v1_conf_examples.md
```diff
@@ -107,7 +117,7 @@ fluentd:
```
frankreno, line 117: In the S3 option above, should we also call out how to properly include the existing output pipieline? Since the conf file has helm template syntax they need to resolve that yeah?

deploy/docs/v1_conf_examples.md
vsinghal13: Yes, even I was thinking the same. It can be confusing. Have created #627  to expose these configs for helm users. As for now, I think linking the Help docs makes more sense as you suggested Frank.

deploy/docs/v1_conf_examples.md
vsinghal13: good point, will add that

PR #637, opened 2020-05-18
Provide default, commented out limits and resources for Prometheus - …

perk-sumo:
…consistent with limits for other components

###### Description

###### Testing performed

- [ ] ci/build.sh


perk-sumo:
@frankreno actually in the upstream there is only a request for the memory (which is 400Mi) that we had before: https://github.com/helm/charts/blob/master/stable/prometheus-operator/values.yaml#L1736

PR #645, opened 2020-05-19
no-space-in-cluster-name

### LINKED ISSUE #608: Add a conditional check on the clusterName to prevent spaces

frankreno:
###### Description

Fixes #608.  When spaces are found in `sumologicClusterName`, replace them with dashes and warn the issue via NOTES.txt.

###### Testing performed

- [ ] ci/build.sh
- [ X ] Redeploy fluentd and fluentd-events pods
- [ X ] Confirm events, logs, and metrics are coming in


deploy/helm/sumologic/values.yaml
```diff
@@ -29,7 +29,7 @@ sumologic:
   # Collector name
   #collectorName: ""
 
-  # Cluster name
+  # Cluster name: Note spaces are not allowed and will be replaced with dashes.
   clusterName: "kubernetes"
```
samjsong, line 33: We would have to change `clusterName` to `clusterNameReplaceSpaceWithDash` here, however this is "technically" a breaking change and would have to be included with 1.0. Otherwise, we can leave the name of this config the same, and just include the comment above it specifying this behavior, since up to now having a space in cluster name would have failed anyway. What do you think?

vsinghal13: I think we can keep the name same `clusterName` and just specify the comment.

frankreno: We would not have to change it actually because I created `clusterNameReplaceSpaceWithDash` as a function in the helpers.tpl file.  If customers use a space before it would work in the collection, but would break in the explore UI at which point we would tell them change it to have no spaces.  So its unlikely anyone has a space given it breaks the Explore tab.

frankreno: That is why I think it is ok to do this in 1.1 as opposed to 1.0.

samjsong: Aha, totally missed that helper file. Thanks Frank

frankreno:
Closing as I will resubmit to fix the issue with other commits in here.

PR #646, opened 2020-05-19
Use Ignore_Older 24h by default

### LINKED ISSUE #596: Set Default IgnoreOlder in FluentBit
We should set a default value for FluentBit's IgnoreOlder field to 24 hours, to prevent ingestion of excessive volumes of historical data when collection is set up

frankreno:
###### Description

Fixes #596.

Set `Ignore_Older` to `24h` by default to prevent older logs from being ingested when collection is first installed.  Relative times are supported.  Tested using shorter time range (1m) and validated that the older logs are not ingested.

https://github.com/fluent/fluent-bit/issues/221
https://docs.fluentbit.io/manual/pipeline/inputs/tail#config

Note, this is not supported on `systemd` plugin so only updating `tail` plugin.

###### Testing performed

- [ ] ci/build.sh


frankreno:
Closing in favor of #664

PR #647, opened 2020-05-19
Upgrade script grep sed bash fixes

### LINKED ISSUE #644: Multiple problems with upgrade script
* upgrade script doesn't work under GNU sed >= 4.4
* upgrade needs recent GNU Bash but doesn't check for that
* grep has to be quite modern but we don't check for that
* tests don't work inside the vagrant env (ubuntu 18.04)

perk-sumo:
###### Description

Fixes #644

* fix for upgrade script doesn't work under GNU sed >= 4.4
* upgrade needs recent GNU Bash but doesn't check for that
* grep does not have to be modern
* allow for newer yq

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
Could you add `vagrant/provision.sh` update from my PR?

PR #649, opened 2020-05-19
Add --force flag for pv

vsinghal13:
Add --force flag for pv

###### Description

###### Testing performed

- [ ] ci/build.sh


vsinghal13:
> LGTM. If it was true in 0.17 and they upgrade to 1.0 is it also needed?

Actually, PV support is not available in 0.17

PR #652, opened 2020-05-20
Fix upgrade script for vagrant env

sumo-drosiek:
###### Description

* Unify sed flags for unix (mac) and linux (ubuntu)
* yq can be in version higher than 3.2.1, the reason of previous condition was that yq installed via snap doesn't have access to filesystem, so all read/write operations were failing:
   - https://mikefarah.gitbook.io/yq/#snap-notes
   - https://snapcraft.io/docs/snap-confinement

###### Testing performed

- [ ] ci/build.sh


perk:
👍 fixed upstream: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/647

PR #664, opened 2020-05-20
Set `Ignore_Older` to `24h` by default to prevent older logs from being ingested on install.

### LINKED ISSUE #596: Set Default IgnoreOlder in FluentBit

frankreno:
###### Description

Fixes #596.

Set `Ignore_Older` to `24h` by default to prevent older logs from being ingested when collection is first installed.  Relative times are supported.  Tested using shorter time range (1m) and validated that the older logs are not ingested.

https://github.com/fluent/fluent-bit/issues/221
https://docs.fluentbit.io/manual/pipeline/inputs/tail#config

Note, this is not supported on `systemd` plugin so only updating `tail` plugin.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -60,6 +60,7 @@ rawConfig: |-
     Skip_Long_Lines  On
     DB               /tail-db/tail-containers-state-sumo.db
     DB.Sync          Normal
+    Ignore_Older     24h
   [INPUT]
     Name            systemd
     Tag             host.*
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -469,6 +469,7 @@ fluent-bit:
       Skip_Long_Lines  On
       DB               /tail-db/tail-containers-state-sumo.db
       DB.Sync          Normal
+      Ignore_Older     24h
     [INPUT]
       Name            systemd
       Tag             host.*


avdhoot:
@frankreno I doubt it will work(not working for me). To `Ignore_Older` work input plugin need parser. more https://github.com/fluent/fluent-bit/issues/716

frankreno:
@avdhoot - we use a parser on our input plugin for container logs. Local testing showed this working but definitely worth another check. Can you share your configuration? Are you seeing older logs ingested for container logs or systemd/kubelet?

avdhoot:
Non working config:
```
    [INPUT]
        Name             tail
        Path             /var/log/containers/*.log
        Multiline        On
        Parser_Firstline multi_line
        Tag              containers.*
        Refresh_Interval 1
        Rotate_Wait      60
        Mem_Buf_Limit    5MB
        Skip_Long_Lines  On
        DB               /tail-db/tail-containers-state-sumo.db
        DB.Sync          Normal
        Ignore_Older 1m
```
Working config. To it work parser should extract time.
```
[INPUT]
    Name             tail
    Path             /var/log/containers/*.log
    Multiline        On
    Parser_Firstline multi_line
    Tag              containers.*
    Refresh_Interval 1
    Rotate_Wait      60
    Mem_Buf_Limit    5MB
    Skip_Long_Lines  On
    DB               /tail-db/tail-containers-state-sumo.db
    DB.Sync          Normal
    Ignore_Older 1m
    Parser  Docker
```

avdhoot:
@frankreno in case you missed it. Result of above config. There is still some leak but it is manageable.  
 
image

frankreno:
Thank you @avdhoot - we will investigate and report back.  In that search are you looking at all logs sent to the collector? You should exclude any logs from systemd because there is a known issue that we ingest old logs there. The `systemd` does not support a way to ignore older log files, so I just want to make sure the leak you see above is not from those logs.

avdhoot:
To make things more clear. I have used config mentioned in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/664#issuecomment-660851105. Currently we are not sending `systemd` logs.

Adhira-Deogade:
@avdhoot With your configs [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/664#issuecomment-660851105), I am still collecting historical logs. Am I doing something wrong? Thanks.

PR #672, opened 2020-05-21
Expose http sources for metrics in values.yaml

sumo-drosiek:
###### Description

Expose http sources in values.yaml

###### ToDO

 - [x] Move out configuration from configmap to separate files (like fluentd conf)
 - [x] Discuss if logs/events/default metrics should be exposed
        I decided to not expose logs and events so far
 - [x] Add enable/disable flag
 - [x] Dynamic generation of fluentd configuration(?)

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -401,13 +401,13 @@ prometheus:
         sourceLabels: [__name__]
     - # control plane metrics
       # coredns
-      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane
+      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane.coredns
       writeRelabelConfigs:
       - action: keep
         regex: coredns
         sourceLabels: [job]
     - # etcd server
-      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane
+      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane.kube-etcd
       writeRelabelConfigs:
       - action: keep
         regex: kube-etcd
--- a/deploy/helm/sumologic/conf/events/events.conf
+++ b/deploy/helm/sumologic/conf/events/events.conf
@@ -35,7 +35,7 @@
   @type sumologic
   @id sumologic.endpoint.events
   sumo_client {{ include "sumologic.sumo_client" . | quote }}
-  endpoint "#{ENV['SUMO_ENDPOINT_EVENTS']}"
+  endpoint "#{ENV['SUMO_ENDPOINT_EVENTS_SOURCE']}"
   data_type logs
   disable_cookies true
   verify_ssl {{ .Values.fluentd.verifySsl | quote }}
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -453,11 +453,11 @@ Generate line for data terraform section
 
 Example usage:
 
-{{ include "terraform.sources.data" $source }}
+{{ include "terraform.sources.data" (dict "Endpoint" "enpoint-default-metrics" "Name" "default") }}
 
 */}}
 {{- define "terraform.sources.data" -}}
-{{ printf "%-41s = \"${sumologic_http_source.%s.url}\"" (include "terraform.sources.endpoint_name" .) . }}
+{{ printf "%-41s = \"${sumologic_http_source.%s.url}\"" .Endpoint .Name }}
 {{- end -}}
 
 {{/*
@@ -483,4 +483,27 @@ resource "sumologic_http_source" "{{ .Name }}" {
     {{- end -}}
     {{ end }}
 }
+{{- end -}}
+
+{{/*
+get configuration variable name for sources confg map
+
+Example usage:
+
+{{ include "terraform.sources.config-map-variable" (dict "Context" .Values "Name" $name "Endpoint" $endpoint) }}
+
+*/}}
+{{- define "terraform.sources.config-map-variable" -}}
+{{- $name := .Name -}}
+{{- $ctx := .Context -}}
+{{- $endpoint := .Endpoint -}}
+{{- if not $endpoint -}}
+{{- $source := (index $ctx.sumologic.sources "default") -}}
+{{- if (index $ctx.sumologic.sources .Name "config-name") -}}
+{{- $endpoint = index $ctx.sumologic.sources .Name "config-name" -}}
+{{- else -}}
+{{- $endpoint = printf "endpoint-%s" (include "terraform.sources.name_metrics" $name) -}}
+{{- end -}}
+{{- end -}}
+{{ $endpoint }}
 {{- end -}}
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -377,3 +377,17 @@ Example usage (as one line):
   </buffer>
 </match>
 {{- end -}}
+
+{{/*
+Convert source name to terraform name:
+ * converts all `-` to `_`
+ * adds `_source` suffix
+
+Example usage:
+
+{{ include "sources.terraform_name" $source }}
+
+*/}}
+{{- define "sources.terraform_name" -}}
+{{ replace "-" "_" .name }}_source
+{{- end -}}
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -354,14 +354,14 @@ Example usage (as one line):
 
 {{ include "utils.metrics.match" (dict 
   "Values" . 
-  "Match" "prometheus.metrics.kubelet" 
+  "Tag" "prometheus.metrics.kubelet" 
   "Endpoint" "SUMO_ENDPOINT_METRICS" 
   "Storage" .Values.fluentd.buffer.filePaths.metrics.default
   "Id" sumologic.endpoint.metrics
 )}}
 */}}
 {{- define "utils.metrics.match" -}}
-<match {{ .Match }}>
+<match {{ .Tag }}>
 {{- if .Drop }}
   @type null
 {{- else }}
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -353,7 +353,7 @@ Generate metrics match configuration
 Example usage (as one line):
 
 {{ include "utils.metrics.match" (dict 
-  "Values" .Values 
+  "Values" . 
   "Match" "prometheus.metrics.kubelet" 
   "Endpoint" "SUMO_ENDPOINT_METRICS" 
   "Storage" .Values.fluentd.buffer.filePaths.metrics.default
@@ -364,10 +364,11 @@ Example usage (as one line):
 <match {{ .Match }}>
   @type sumologic
   @id {{ .Id }}
-  endpoint "#{ENV['{{ .Endpoint }}']}"
-{{- .Values.fluentd.metrics.outputConf | nindent 2 }}
+  sumo_client {{ include "sumologic.sumo_client" .Context | quote }}
+  endpoint "#{ENV['SUMO_ENDPOINT_{{ replace "-" "_" .Endpoint | upper }}']}"
+{{- .Context.Values.fluentd.metrics.outputConf | nindent 2 }}
   <buffer>
-    {{- if or .Values.fluentd.persistence.enabled (eq .Values.fluentd.buffer.type "file") }}
+    {{- if or .Context.Values.fluentd.persistence.enabled (eq .Context.Values.fluentd.buffer.type "file") }}
     @type file
     path {{ .Storage }}
     {{- else }}
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -385,10 +385,10 @@ Convert source name to terraform name:
 
 Example usage:
 
-{{ include "sources.terraform_name" $source }}
+{{ include "terraform.sources.name" $source }}
 
 */}}
-{{- define "sources.terraform_name" -}}
+{{- define "terraform.sources.name" -}}
 {{ replace "-" "_" .name }}_source
 {{- end -}}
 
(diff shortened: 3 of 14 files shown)


deploy/helm/sumologic/templates/setup/setup-configmap.yaml
```diff
@@ -81,52 +69,18 @@ data:
...
+    {{ range $source := .Values.sumologic.sources }}
+    resource "sumologic_http_source" "{{ template "sources.terraform_name" $source }}" {
+        name         = local.{{ template "sources.terraform_source_name" $source }}
         collector_id = "${sumologic_collector.collector.id}"
     }
+    {{ end -}}
 
     resource "sumologic_http_source" "events_source" {
         name         = local.events-source-name
         category     = {{ if .Values.fluentd.events.sourceCategory }}{{ .Values.fluentd.events.sourceCategory | quote }}{{- else}}{{ "\"${var.cluster_name}/${local.events-source-name}\"" }}{{- end}}
```
sumo-drosiek, line 81: move category inside values.yaml

deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -146,15 +155,16 @@ data:
...
-        endpoint-events                           = "${sumologic_http_source.events_source.url}"
-        endpoint-logs                             = "${sumologic_http_source.logs_source.url}"
         endpoint-metrics                          = "${sumologic_http_source.default_metrics_source.url}"
         endpoint-metrics-apiserver                = "${sumologic_http_source.apiserver_metrics_source.url}"
         endpoint-metrics-kube-controller-manager  = "${sumologic_http_source.kube_controller_manager_metrics_source.url}"
         endpoint-metrics-kube-scheduler           = "${sumologic_http_source.kube_scheduler_metrics_source.url}"
         endpoint-metrics-kube-state               = "${sumologic_http_source.kube_state_metrics_source.url}"
         endpoint-metrics-kubelet                  = "${sumologic_http_source.kubelet_metrics_source.url}"
         endpoint-metrics-node-exporter            = "${sumologic_http_source.node_exporter_metrics_source.url}"
+        endpoint-control-plane                    = "${sumologic_http_source.control_plane_metrics_source.url}"
```
perk-sumo, line 165: It's back, good :)

deploy/helm/sumologic/values.yaml
```diff
@@ -390,6 +420,58 @@ fluentd:
...
+      control-plane:
+        match: prometheus.metrics.control-plane**
+        id: sumologic.endpoint.metrics.control.plane
+        endpoint: metrics-control-plane
+        weight: 90
+      default:
+        match: prometheus.metrics**
+        id: sumologic.endpoint.metrics
+        endpoint: metrics
+        weight: 100
```
perk-sumo, line 473: I like this 👍

deploy/helm/sumologic/values.yaml
```diff
@@ -82,6 +82,36 @@ sumologic:
         helm.sh/hook-weight: "0"
         helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
 
+  ## Configuration of http sources
+  ## name: source name
+  ## value: source values - it's visible in sumologic platform
+  ## endpoint: binds source with fluentd configuration: (fluentd.metrics.output[])
+  sources:
```
perk-sumo, line 89: This would better be a default sources + additional sources. Or something like that.

sumo-drosiek:
I updated the PR with more user-friendly configuration, but this is highly breaking change for non-helm users

sumo-drosiek:
Here is example of adding new source to sumologic and fluentd:

```
    --set sumologic.sources.new_source.value="my new source" \ # create new source named my new source and available in helm as new_source
    --set fluentd.metrics.output.new_source.match="test_match" \ # fluentd <match test_match> clausule
    --set fluentd.metrics.output.new_source.id="test_id" \ # fluentd @id parameter
    --set fluentd.metrics.output.new_source.weight=0 \ # order of fliuentd match (lower, earlier)
    --set fluentd.metrics.output.new_source.drop=true # if true, drop traffic instead of passing it to the sumologic \
    --set fluentd.buffer.filePaths.metrics.new_source=/fluentd/buffer/metrics.apiserver # set persistence
```

sumo-drosiek:
Unfortunately if source value is changed, the new source is created via terraform and old one is not cleaned up

deploy/helm/prometheus-overrides.yaml
```diff
@@ -228,13 +228,13 @@ prometheus:
...
       # coredns
-      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane
+      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane.coredns
       writeRelabelConfigs:
       - action: keep
         regex: coredns
         sourceLabels: [job]
     - # etcd server
-      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane
+      url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane.kube-etcd
```
sumo-drosiek, line 237: Fixes prometheus issue. It has been saying that configuration for `http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane` is duplicated

sumo-drosiek:
@frankreno Do we consider fluentd env changes as breaking changes?

deploy/helm/prometheus-overrides.yaml
frankreno: The only thing i find misleading with this setup is that we collect from other control plane components and do not prefix them with control-plane.  Any reason to just remove the control-plane part?

deploy/helm/sumologic/conf/events/events.conf
```diff
@@ -35,7 +35,7 @@
   @type sumologic
   @id sumologic.endpoint.events
   sumo_client {{ include "sumologic.sumo_client" . | quote }}
-  endpoint "#{ENV['SUMO_ENDPOINT_EVENTS']}"
+  endpoint "#{ENV['SUMO_ENDPOINT_EVENTS_SOURCE']}"
```
frankreno, line 38: Would this be considered a breaking change for non-helm users? Helm should handle this, but what abotu non-helm.

deploy/helm/sumologic/values.yaml
```diff
@@ -390,6 +412,52 @@ fluentd:
     ## To use additional output plugins
     extraOutputPluginConf: |-
 
+    ## Output configuration
+    ## match: fluentd match
```
frankreno, line 416: nit - we should call this tag imo because that is what it really is, the tag for the match block.

deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -79,51 +83,56 @@ data:
...
         collector_id = "${sumologic_collector.collector.id}"
     }
-  
-    resource "sumologic_http_source" "node_exporter_metrics_source" {
-        name         = local.node-exporter-metrics-source-name
+    
+    resource "sumologic_http_source" "events_source" {
+        name         = local.events_source
         collector_id = "${sumologic_collector.collector.id}"
+        category     = "${var.cluster_name}/${local.events_source}"
```
frankreno, line 135: this seems to revert the current behavior of being able to change the sourceCategory for events?

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/conf/setup/sumo-k8s.tf#L56

deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -79,51 +83,56 @@ data:
...
-    resource "sumologic_http_source" "kubelet_metrics_source" {
-        name         = local.kubelet-metrics-source-name
+    
+    resource "sumologic_http_source" "scheduler_metrics_source" {
+        name         = local.scheduler_metrics_source
         collector_id = "${sumologic_collector.collector.id}"
     }
-  
+    
+    resource "sumologic_http_source" "state_metrics_source" {
```
frankreno, line 122: Perhaps we do this in another PR, but I believe we need to expose the sourceCategory for all metrics sources.  Right now, we do not set it which results in all metrics getting the default sourceCageory of "Http Input". this makes data volume reporting tricky.  We should follow the same model as events for the metrics sources so that the data volume reporting in Sumo is managemble.

frankreno: Logs are a bit different, because logs sourceCategory is set via the HTTP header. This is not true for metrics and events which is why we need to be able to provide a sensible default.

deploy/kubernetes/setup-sumologic.yaml.tmpl
samjsong: This is actually consistent with our current behavior since this file is for non-helm users

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/kubernetes/setup-sumologic.yaml.tmpl#L95

deploy/helm/sumologic/conf/events/events.conf
samjsong: actually since non-helm users just run `kubectl apply -f fluentd-sumologic.yaml` to upgrade, the new yaml file should have the renamed environment variables so it should be fine.

However there IS a breaking change here, and that is a result of the keys in the sumologic secret being changed (added suffix `-source`), which means that non-helm users will have to re-run the setup job in order to get an updated secret (though the collector and sources will remain untouched).

samjsong:
Overall LGTM, one thing that might be cool to add (probably in separate PR) is the ability to append more sources. In fact the Zaidan team has the need for additional http sources and in my migration work I don't think I will be able to move everything until we have this functionality

deploy/helm/prometheus-overrides.yaml
sumo-drosiek: Yes, the fluentd configuration will be more complex: either more complex regex for tags or output section per endpoint

sumo-drosiek:
@samjsong What do you mean by additional http sources?

I think the `sumologic.sources` section should meet you expectations :)

deploy/helm/prometheus-overrides.yaml
frankreno: How many metrics is each of these producing? We need to be conscious of blacklisting and if we are close to the limit, we should distribute across multiiple HTTP sources to prevent that.

deploy/helm/sumologic/conf/events/events.conf
frankreno: Good catch, yeah we dont what the breaking change from the secret.

deploy/helm/sumologic/values.yaml
sumo-drosiek: fixed

sumo-drosiek:
@frankreno We are planning to reduce the metrics list anyway. Waiting for dashboards and exact list of metrics

vsinghal13:
Also, one more suggestion, it is best practice to have separate `tf` files for variables, resources, and providers. Would be great if we can split `sumo-k8s.tf` similarly.

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -353,27 +353,157 @@ Generate metrics match configuration
...
+{{- if .Drop }}
+  @type null
+{{- else }}
   @type sumologic
   @id {{ .Id }}
-  endpoint "#{ENV['{{ .Endpoint }}']}"
-{{- .Values.fluentd.metrics.outputConf | nindent 2 }}
+  sumo_client {{ include "sumologic.sumo_client" .Context | quote }}
+  endpoint "#{ENV['{{ include "terraform.sources.endpoint" .Endpoint}}']}"
+{{- .Context.Values.fluentd.metrics.outputConf | nindent 2 }}
```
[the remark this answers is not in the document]
sumo-drosiek: I need to pass k8s context to have access to .Values, `.Context` in this template is simply `.` from kubernetes. 
`.` is override in `range` functions and templates, so there is need to pass it as a variable (or a dict value for templates, because they get only one argument). Here is flow (btw it covers both cases):
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/8e58e3f3f1da3a704e7b2a70bd166c1cd9aa1012/deploy/helm/sumologic/conf/metrics/metrics.conf#L50
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/8e58e3f3f1da3a704e7b2a70bd166c1cd9aa1012/deploy/helm/sumologic/conf/metrics/metrics.conf#L63

deploy/kubernetes/fluentd-sumologic.yaml.tmpl
```diff
@@ -389,107 +389,114 @@ data:
...
-    
-      <match prometheus.metrics.kubelet**>
+      
+      <match prometheus.metrics.container**>
         @type sumologic
-        @id sumologic.endpoint.metrics.kubelet
-        endpoint "#{ENV['SUMO_ENDPOINT_METRICS_KUBELET']}"
+        @id sumologic.endpoint.metrics.container
+        sumo_client "k8s_1.0.0"
+        endpoint "#{ENV['SUMO_ENDPOINT_KUBELET_METRICS_SOURCE']}"
```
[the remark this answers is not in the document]
sumo-drosiek: I suppose we have an exception here. For 1.0 release we are using kubelet endpoint for container.
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/bd42eaeaa7da7b5d2dd64c9e81a5bef69214c097/deploy/helm/sumologic/conf/metrics/metrics.conf#L73-L88

sumo-drosiek:
@vsinghal13 I will address split of the tf file in the another PR. This PR is oversized already 😞

deploy/helm/sumologic/values.yaml
```diff
@@ -429,6 +458,52 @@ fluentd:
     #     mountPath: /certs
     #     readOnly: true
   
+    ## Output configuration
+    ## tag: tag for fluentd match
+    ## id: sumologic output @id
+    ## endpoint: [optional] [output key by default] It should match the sumologic.sources[].endpoint
+    ## drop: [optional] [false by default] Change it to true to drop traffic for specific output
+    ## weight: [optional [0 by default] Order of fluentd output (lower means higher priority)
+    ## The weight has meaning in case of fluentd matches
```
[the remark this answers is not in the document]
sumo-drosiek: ok, I will address it in documentation rather than comment in the values.yaml, but here we go:

Let's say we have two two sources (`source1`, `source2`)

```
      source1:
        tag: prometheus.metrics.custom**
        weight: 90
      source2:
        tag: **
        weight: 80
```
is going to generate in fluentd 
```
# weight 80
<match **>
  # takes all records
</match>
# weight 90
<match prometheus.metrics.custom**>
  # it won't work
</match>
```

but if we switch the weights:

```
      source1:
        tag: **
        weight: 80
      source2:
        tag: prometheus.metrics**
        weight: 90
```
then flunetd config will change to
```
# weight 80
<match prometheus.metrics.custom**>
  # takes custom records
</match>
# weight 90
<match **>
  # takes all other records
</match>
```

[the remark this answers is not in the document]
sumo-drosiek: I'm going to address it in https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/680#issuecomment-634607171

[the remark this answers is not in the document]
vsinghal13:
```# weight 80
<match **>
  # takes all records
</match>
# weight 90
<match prometheus.metrics.custom**>
  # it won't work
</match>```

Even if the user specified this, the second match won't work because everything was already matched by the catch all block, right?

Issue #627, opened 2020-05-14
expose extra ENVS


vsinghal13:
We need to add support for customers to allow adding new environment variables.

Will have to expose config for sts, secret, etc

Use Case: Adding a new custom endpoint

vsinghal13:
It would be great if it could be done automatically: If the user specifies name of an additional source, we could generate the necessary configs across all the files.  cc @samjsong

PR #681, opened 2020-05-26
expose extraEnvVars, extraVolumes, extraVolumeMounts

vsinghal13:
###### Description

Solves #627 

###### Testing performed

- [ ] ci/build.sh


vsinghal13:
> LGTM - However, I think we should also expose the EnvVars in the metrics statefulset because customers may need to add endpoints for metrics as well.

Should we split this section into two, one for logs and one for metrics? Or have the same set of env available in both the sts?

frankreno:
> > LGTM - However, I think we should also expose the EnvVars in the metrics statefulset because customers may need to add endpoints for metrics as well.
> 
> Should we split this section into two, one for logs and one for metrics? Or have the same set of env available in both the sts?

I think we need seperate sections as we would have different endpoints for logs and metrics currently.

PR #682, opened 2020-05-26
add collectionMonitoring logic

### LINKED ISSUE #211: Can we default the collection process to exclude Sumologic Logs
The suggestion that we should default to excluding the sumologic namespace as those are primarily sumo specific logs.

vsinghal13:
###### Description

Fixes #211. 

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
@@ -28,7 +28,7 @@
       pattern /\[error\]|\[fatal\]|drop_oldest_chunk|retry succeeded/
     </regexp>
   </filter>
-{{- end}}
+{{- end }}
   # third-party kubernetes metadata  filter plugin
   <filter containers.**>
     @type kubernetes_metadata
@@ -50,11 +50,11 @@
     exclude_namespace_regex {{ printf "%s,%s" .Release.Namespace .Values.fluentd.logs.containers.excludeNamespaceRegex | quote }}
     {{- else}}
     exclude_namespace_regex {{ .Values.fluentd.logs.containers.excludeNamespaceRegex | quote }}
-    {{- end}}
+    {{- end }}
   </filter>
   {{ if .Values.fluentd.logs.containers.overrideOutputConf }}
   {{ .Values.fluentd.logs.containers.overrideOutputConf | nindent 4 }}
-  {{- else}}
+  {{- else }}
   <match containers.**>
     @type sumologic
     @id sumologic.endpoint.logs
@@ -71,6 +71,6 @@
       @include buffer.output.conf
     </buffer>
   </match>
-  {{- end}}
+  {{- end }}
 </label>
-{{- end}}
+{{- end }}
--- a/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
+++ b/deploy/helm/sumologic/conf/logs/logs.source.containers.conf
@@ -46,7 +46,7 @@
   <filter containers.**>
     @type kubernetes_sumologic
     @include logs.kubernetes.sumologic.filter.conf
-    {{ if eq .Values.sumologic.collectionMonitoring false}}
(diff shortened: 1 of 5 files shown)


deploy/helm/sumologic/conf/metrics/metrics.conf
```diff
@@ -19,6 +19,15 @@
     </exclude>
   </filter>
 {{- end }}
+{{- if eq .Values.sumologic.collectionMonitoring "false" }}
+  <filter prometheus.metrics**>
```
samjsong, line 23: CI seems to be failing due to the trailing whitespace on this line

deploy/helm/sumologic/conf/metrics/metrics.conf
```diff
@@ -19,6 +19,15 @@
     </exclude>
   </filter>
 {{- end }}
+{{- if eq .Values.sumologic.collectionMonitoring "false" }}
+  <filter prometheus.metrics**> 
+    @type grep
+    <exclude>
+      key @metric
+      pattern /^up|^prometheus_remote_storage_.*|^fluentd_.*|^fluentbit.*|^otelcol.*$/
```
samjsong, line 27: Ideally we would be able to remove from the Prometheus side so we're not sending unnecessary metrics to Fluentd, though I understand we can't template in the `values.yaml` file so this would be difficult to achieve...

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -46,6 +46,11 @@
   <filter containers.**>
     @type kubernetes_sumologic
     @include logs.kubernetes.sumologic.filter.conf
+    {{ if eq .Values.sumologic.collectionMonitoring false}}
+    exclude_namespace_regex {{ .Release.Namespace | quote }}
+    {{- else}}
+    exclude_namespace_regex {{ .Values.fluentd.logs.containers.excludeNamespaceRegex | quote }}
```
samjsong, line 52: we'll want to do a combination of the two, we don't want to ignore the customer's `excludeNamespaceRegex` just because they want to stop collecting the `sumologic` data

vsinghal13: done

deploy/helm/sumologic/conf/metrics/metrics.conf
vsinghal13: yes, that's why I added the filter here

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
samjsong: since this is regex should it be `"%s|%s"` instead of comma separated?

deploy/helm/sumologic/conf/logs/logs.source.containers.conf
```diff
@@ -46,6 +46,11 @@
   <filter containers.**>
     @type kubernetes_sumologic
     @include logs.kubernetes.sumologic.filter.conf
+    {{ if eq .Values.sumologic.collectionMonitoring false }}
+    exclude_namespace_regex {{ printf "%s,%s" .Release.Namespace .Values.fluentd.logs.containers.excludeNamespaceRegex | quote }}
+    {{- else}}
+    exclude_namespace_regex {{ .Values.fluentd.logs.containers.excludeNamespaceRegex | quote }}
+    {{- end}}
```
samjsong, line 53: nit: can we add space `{{- end }}` here and line 51 as well?

sumo-drosiek:
Is there way to exclude by pod labels? It would be more kubernetes way

I don't feel confident with excluding by namespace name in regex, because it can provide a lot of edge cases

vsinghal13:
> Is there way to exclude by pod labels? It would be more kubernetes way
> 
> I don't feel confident with excluding by namespace name in regex, because it can provide a lot of edge cases
> 
> Except for that question, LGTM :)

I could not find a way to exclude via pod labels.

PR #683, opened 2020-05-26
optional metrics resources

### LINKED ISSUE #678: Make it optional to collect kubernetes metrics
Is it possible to make it optional to collect kubernetes metrics in the HELM chart? Similar to kubernetes events https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L395? 

Background for this: we are using other tool to collect kubernetes metrics, and only use Sumo for log aggregation. Because metrics is not optional, the helm chart will always try to create metric service, statefulset, etc.

vsinghal13:
###### Description

Fixes #678

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -34,6 +34,10 @@ nodeExporter:
     interval:
 # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
 kubeTargetVersionOverride: 1.13.0-0
+## Set the enabled flag to false for either of the below two purposes:
+## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
+## 2. Disable metrics collection altogether
+enabled: true
 alertmanager:
   enabled: false
 grafana:
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -34,10 +34,6 @@ nodeExporter:
     interval:
 # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
 kubeTargetVersionOverride: 1.13.0-0
-## Set the enabled flag to false for either of the below two purposes:
-## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-## 2. Disable metrics collection altogether
-enabled: true
 alertmanager:
   enabled: false
 grafana:
--- a/deploy/helm/sumologic/requirements.yaml
+++ b/deploy/helm/sumologic/requirements.yaml
@@ -6,7 +6,7 @@ dependencies:
   - name: prometheus-operator
     version: 8.13.8
     repository: https://kubernetes-charts.storage.googleapis.com/
-    condition: prometheus-operator.enabled
+    condition: fluentd.metrics.enabled
   - name: falco
     version: 1.1.7
     repository: https://kubernetes-charts.storage.googleapis.com/
--- a/deploy/helm/sumologic/requirements.yaml
+++ b/deploy/helm/sumologic/requirements.yaml
@@ -2,7 +2,7 @@ dependencies:
   - name: fluent-bit
     version: 2.8.14
     repository: https://kubernetes-charts.storage.googleapis.com/
-    condition: fluent-bit.enabled
+    condition: sumologic.logs.enabled,fluent-bit.enabled
   - name: prometheus-operator
     version: 8.13.8
     repository: https://kubernetes-charts.storage.googleapis.com/
--- a/deploy/helm/sumologic/requirements.yaml
+++ b/deploy/helm/sumologic/requirements.yaml
@@ -6,7 +6,7 @@ dependencies:
   - name: prometheus-operator
     version: 8.13.8
     repository: https://kubernetes-charts.storage.googleapis.com/
-    condition: fluentd.metrics.enabled
+    condition: sumologic.metrics.enabled,prometheus-operator.enabled
   - name: falco
     version: 1.1.7
     repository: https://kubernetes-charts.storage.googleapis.com/
--- a/deploy/helm/sumologic/templates/metrics-configmap.yaml
+++ b/deploy/helm/sumologic/templates/metrics-configmap.yaml
@@ -1,4 +1,4 @@
-{{- if eq .Values.prometheus-operator.enabled true }}
+{{- if index .Values "prometheus-operator" "enabled" }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/deploy/helm/sumologic/templates/metrics-configmap.yaml
+++ b/deploy/helm/sumologic/templates/metrics-configmap.yaml
@@ -1,4 +1,4 @@
-{{- if eq .Values.fluentd.metrics.enabled true }}
+{{- if and (eq .Values.sumologic.metrics.enabled true) (eq .Values.fluentd.metrics.enabled true) }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/deploy/helm/sumologic/templates/metrics-configmap.yaml
+++ b/deploy/helm/sumologic/templates/metrics-configmap.yaml
@@ -1,3 +1,4 @@
+{{- if eq .Values.prometheus-operator.enabled true }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -10,3 +11,4 @@ data:
 {{- .Values.fluentd.metrics.rawConfig | nindent 4 }}
 {{- (tpl (.Files.Glob "conf/*.conf").AsConfig .) | nindent 2 }}
 {{- (tpl (.Files.Glob "conf/metrics/*.conf").AsConfig .) | nindent 2 }}
+{{- end }}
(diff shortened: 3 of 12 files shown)


frankreno:
One thing that was brought up on the issue that we missed, we should do this same conditional rendering for all metrics components.  The service, configMap, HPA, etc.

vsinghal13:
Have added the checks for `service`, `configmap` and `headless-service` as well. Skipped it for `hpa` as it already has the check for `fluentd.metrics.autoscaling.enabled` and is `false` by default.

sumo-drosiek:
I'm afraid that we cannot disable metrics using prometheus flag, because customers can have their own prometheus and still wants to send metrics. We should go with two flags unfortunately

frankreno:
Good point on existing Prometheus case. It will require two seperate flags.

perk-sumo:
Please check this out, we should be able to use one flag to rule them all: https://github.com/helm/helm/issues/2111

deploy/helm/sumologic/requirements.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
   - name: prometheus-operator
     version: 8.13.8
     repository: https://kubernetes-charts.storage.googleapis.com/
-    condition: prometheus-operator.enabled
+    condition: fluentd.metrics.enabled
```
sumo-drosiek, line 9: I believe we should use both conditions here, because now there is no way to disable operator without disabling metrics. Am I right?

perk-sumo: This should check for `sumologic.metrics.enabled` and `prometheus-operator.enabled`

deploy/helm/prometheus-overrides.yaml
```diff
@@ -34,10 +34,6 @@ nodeExporter:
     interval:
 # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
 kubeTargetVersionOverride: 1.13.0-0
-## Set the enabled flag to false for either of the below two purposes:
-## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-## 2. Disable metrics collection altogether
-enabled: true
```
perk-sumo, line 40: I think this should stay if someone would like to disable `prometheus-operator` only (in case they use their own prometheus but still want to use fluentd for tagging)

deploy/helm/sumologic/templates/metrics-configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.fluentd.metrics.enabled true }}
```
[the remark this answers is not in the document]
vsinghal13: I used this flag as this is part of `fluentd` itself. and we already have `fluentd.events.enabled` as compared to traces which is not part of the fluentd section.

Also I feel having both `sumologic.metrics.enabled` and `fluentd.metrics.enabled` will not add any functionality as it can be covered by only one.

Thoughts? cc @frankreno @samjsong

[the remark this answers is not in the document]
samjsong: hmm, I'm not sure I understand the usecase for having both `sumologic.metrics.enabled` and `fluentd.metrics.enabled`. I can't think of a case where you would want the fluentd pipeline enabled but not the metrics resources, or vice versa. @perk-sumo can you share your reasoning please?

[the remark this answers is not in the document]
sumo-drosiek: I feel we should have `sumologic.metrics.enabled` instead of `fluentd.metrics.enabled`, especially if we gonna replace fluentd in the future, but it's meaningful change, so I would stick to the current flag

deploy/helm/sumologic/values.yaml
```diff
@@ -592,10 +598,7 @@ prometheus-operator:
       interval:
   # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
   kubeTargetVersionOverride: 1.13.0-0
-  ## Set the enabled flag to false for either of the below two purposes:
-  ## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-  ## 2. Disable metrics collection altogether
-  enabled: true
```
sumo-drosiek, line 598: Could we revert this removal?

deploy/helm/sumologic/templates/metrics-configmap.yaml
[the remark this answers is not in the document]
perk-sumo: The idea is to have global flag that is not attached to the underlying implementation so that we (or the user) can switch the tool easily.
There are couple of usecases:
* disable all the metrics, all the fluentd's, prometheus, configmaps etc - set `sumologic.metrics.enabled=false`
* enable all the metrics in the default mode: `sumologic.metrics.enabled=true`
* enable metrics, but use own prometheus: `sumologic.metrics.enabled=true` and `prometheus-operator.enabled=false`
* enable metrics, but use own fluentd for tagging: `sumologic.metrics.enabled=true` and `fluentd.metrics.enabled=false`
* enable metrics, but use experimental OTC for that: `sumologic.metrics.enabled` and `fluentd.metrics.enabled=false`, `prometheus-operator.enabled=false`, `otelcol.metrics.enabled=true`

deploy/helm/sumologic/values.yaml
perk-sumo: It has been moved up which I actually like.

sumo-drosiek: Could you link me the line?
I see only addition to logs and metrics

deploy/helm/sumologic/templates/metrics-configmap.yaml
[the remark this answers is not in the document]
frankreno: What @perk-sumo suggestion seems to be the cleanest approach.

deploy/helm/sumologic/values.yaml
vsinghal13: added it back

deploy/helm/sumologic/templates/metrics-statefulset.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.sumologic.metrics.enabled true }}
```
perk-sumo, line 1:
```suggestion
{{- if and (eq .Values.sumologic.metrics.enabled true) (eq .Values.fluentd.metrics.enabled true) }}
```

deploy/helm/sumologic/templates/statefulset.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.sumologic.logs.enabled true }}
```
perk-sumo, line 1:
```suggestion
{{- if and (eq .Values.sumologic.logs.enabled true) (eq .Values.fluentd.logs.enabled true) }}
```

deploy/helm/sumologic/templates/metrics-service-headless.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.sumologic.metrics.enabled true }}
```
perk-sumo, line 1:
```suggestion
{{- if and (eq .Values.sumologic.metrics.enabled true) (eq .Values.fluentd.metrics.enabled true) }}
```

deploy/helm/sumologic/templates/metrics-service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.sumologic.metrics.enabled true }}
```
perk-sumo, line 1:
```suggestion
{{- if and (eq .Values.sumologic.metrics.enabled true) (eq .Values.fluentd.metrics.enabled true) }}
```

deploy/helm/sumologic/templates/service-headless.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.sumologic.logs.enabled true }}
```
perk-sumo, line 1:
```suggestion
{{- if and (eq .Values.sumologic.logs.enabled true) (eq .Values.fluentd.logs.enabled true) }}
```

deploy/helm/sumologic/templates/service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.sumologic.logs.enabled true }}
```
perk-sumo, line 1:
```suggestion
{{- if and (eq .Values.sumologic.logs.enabled true) (eq .Values.fluentd.logs.enabled true) }}
```

deploy/helm/sumologic/requirements.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
   - name: prometheus-operator
     version: 8.13.8
     repository: https://kubernetes-charts.storage.googleapis.com/
-    condition: prometheus-operator.enabled
+    condition: sumologic.metrics.enabled,prometheus-operator.enabled
```
perk-sumo, line 9: 👍 Please add this into the fluent-bit as well:
```condition: sumologic.logs.enabled,fluent-bit.enabled```

PR #697, opened 2020-06-01
Filter list of control-plane app metrics

sumo-drosiek:
###### Description

Restrict coredns and etcd  metrics

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -1097,12 +1097,26 @@ prometheus-operator:
               regex: (?:up|prometheus_remote_storage_.*|fluentd_.*|fluentbit.*|otelcol.*)
               sourceLabels: [__name__]
         # control plane metrics
-        # coredns
+        # coredns:
+        # coredns_cache_size
+        # coredns_cache_hits_total
+        # coredns_cache_misses_total
+        # coredns_dns_request_duration_seconds_count
+        # coredns_dns_request_duration_seconds_sum
+        # coredns_dns_request_count_total
+        # coredns_dns_response_rcode_count_total
+        # coredns_forward_request_count_total
+        # process_cpu_seconds_total
+        # process_open_fds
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -998,18 +998,53 @@ prometheus-operator:
...
+        # coredns_forward_request_count_total
+        # process_cpu_seconds_total
+        # process_open_fds
+        # process_resident_memory_bytes
+        - url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane.coredns
           writeRelabelConfigs:
             - action: keep
               regex: coredns
               sourceLabels: [job]
+            - action: keep
```
frankreno, line 1018: Just curious on the reason to have 2 keep blocks vs something like this:

```
        - url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.control-plane.coredns
           writeRelabelConfigs:
             - action: keep
               regex: coredns;coredns_cache_(size|(hits|misses)_total)|coredns_dns_request_duration_seconds_(count|sum)|coredns_(dns_request|dns_response_rcode|forward_request)_count_total|process_(cpu_seconds_total|open_fds|resident_memory_bytes)
               sourceLabels: [job,__name__]
```

sumo-drosiek: It's more clear in my personal opinion :)
I also wasn't sure how exactly works the writeRelabelConfig for such construct, but I read the docs and I don't see any risk related to that.
I can change it to be more compressed and to keep convention

frankreno: The only reason I brought it up was just consistency like we do the other rules. I have no problem with doing it this way either.

sumo-drosiek: Convention is important :D
I updated PR, need to re-test it :)

PR #699, opened 2020-06-01
Update documentation for additional sources

sumo-drosiek:
###### Description

Update documentation for additional sources. Waiting for adding logs and events, because it can change configuration structure

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/docs/additional_prometheus_configuration.md
+++ b/deploy/docs/additional_prometheus_configuration.md
@@ -162,10 +162,10 @@ sumologic:
         name: my-source-name
 ```
 
-This will create a new http source with `my-source-name` name in you sumologic account.
+This will create a new HTTP source with the name `my-source-name` in your Sumo Logic account.
 
 **This configuration doesn't modify or remove HTTP sources from your account.
-If you rename a source in `values.yaml`, new source will be added to sumologic platform**
+If you rename a source in `values.yaml`, the new source will be added to your Sumo Logic account**
 
 ### Update Fluentd configuration
 
@@ -193,9 +193,9 @@ fluentd:
         drop: true  # Drop records which match the tag (false by default)
 ```
 
-The weight concept helps to keep fluentd outputs in correct order.
-For above example, the `alternative_source` covers all tags for the `my_source`.
-It means that if `alternative_source` would have less weight than `my_source`
+The weight concept helps to keep Fluentd outputs in the correct order.
+In the above example, the `alternative_source` covers all tags for the `my_source`.
+This means if `alternative_source` has less weight than `my_source`
 it would process all records and none of them would be taken by `my_source`.
 
 ### Update the prometheus-overrides.yaml file to forward the metrics to Fluentd.
--- a/deploy/docs/additional_prometheus_configuration.md
+++ b/deploy/docs/additional_prometheus_configuration.md
@@ -163,6 +163,7 @@ sumologic:
 ```
 
 This will create a new HTTP source with the name `my-source-name` in your Sumo Logic account.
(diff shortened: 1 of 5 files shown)


deploy/docs/additional_prometheus_configuration.md
```diff
@@ -149,70 +149,52 @@ prometheus-operator:  # For values.yaml
...
+  sources:
+    # ...
+    my_source:
+      name: my-source-name
 ```
 
-  * Next you need to edit the Fluentd Deployment and add a new environment variable, pointing to the new secret.  Assuming you installed the collector in the  `sumologic` namespace, you can run `kubectl -n sumologic edit deployment collection-sumologic` or edit the YAML you deployed when you set up collection. Note, if you installed using helm, the name of the deployment may be different depending on how you installed the helm chart.
-  * Locate the `SUMO_ENDPOINT_LOGS` environment variable in the YAML and add a new environment variable that points to the secret key you created. The following is an example.
-  
+This will create a new http source in you sumologic account.
```
sumo-drosiek, line 163:
```suggestion
This will create a new http source with `my-source-name` in you sumologic account.
```

deploy/docs/additional_prometheus_configuration.md
```diff
@@ -149,70 +149,54 @@ prometheus-operator:  # For values.yaml
...
+        name: my-source-name
 ```
 
-  * Next you need to edit the Fluentd Deployment and add a new environment variable, pointing to the new secret.  Assuming you installed the collector in the  `sumologic` namespace, you can run `kubectl -n sumologic edit deployment collection-sumologic` or edit the YAML you deployed when you set up collection. Note, if you installed using helm, the name of the deployment may be different depending on how you installed the helm chart.
-  * Locate the `SUMO_ENDPOINT_LOGS` environment variable in the YAML and add a new environment variable that points to the secret key you created. The following is an example.
-  
+This will create a new HTTP source with the name `my-source-name` in your Sumo Logic account.
+
+**This configuration doesn't modify or remove HTTP sources from your account.
+If you rename a source in `values.yaml`, the new source will be added to your Sumo Logic account**
```
samjsong, line 168: we might want to specify that this HTTP source will be added to the same collector, though maybe it's implied? not too sure about this

deploy/docs/additional_prometheus_configuration.md
```diff
@@ -149,70 +149,54 @@ prometheus-operator:  # For values.yaml
...
-             @type sumologic
-             @id sumologic.endpoint.metrics
-             endpoint "#{ENV['SUMO_ENDPOINT_METRICS']}"
-             @include metrics.output.conf
-           </match>
-```
+The weight concept helps to keep Fluentd outputs in the correct order.
+In the above example, the `alternative_source` covers all tags for the `my_source`.
+This means if `alternative_source` has less weight than `my_source`
+it would process all records and none of them would be taken by `my_source`.
```
samjsong, line 199: This could be a difficult concept for those unfamiliar with Fluentd pipelines, maybe a link to some Fluentd documentation could be helpful here

deploy/docs/additional_prometheus_configuration.md
```diff
@@ -149,70 +149,57 @@ prometheus-operator:  # For values.yaml
...
+
+NOTE: [Explanation of fluentd match order](https://docs.fluentd.org/configuration/config-file#note-on-match-order)
```
samjsong, line 202:
```suggestion
NOTE: [Explanation of Fluentd match order](https://docs.fluentd.org/configuration/config-file#note-on-match-order)
```

PR #702, opened 2020-06-02
Fix Travis build due to default shallow git clone

perk-sumo:
###### Description

Fix Travis build due to default shallow git clone

https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

###### Testing performed

- [ ] ci/build.sh
- [ ] Redeploy fluentd and fluentd-events pods
- [ ] Confirm events, logs, and metrics are coming in

PR #703, opened 2020-06-02
Fix Travis build due to default shallow git clone

perk-sumo:
[fix] Travis build failed to create proper build tags due to default shallow git clone

`git describe` won't work correctly if there is more than 50 (Travis default) commits from the last description (tag)
https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth
###### Description

###### Testing performed

- [ ] ci/build.sh


perk-sumo:
Fixed by #702

PR #712, opened 2020-06-05
add secret cleanup job with pre-delete hook

vsinghal13:
###### Description

###### Testing performed

- [ ] ci/build.sh


deploy/helm/sumologic/templates/setup/setup-cleanup.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      serviceAccountName: {{ template "sumologic.metadata.name.setup.roles.serviceaccount" . }}
+      containers:
+        - name: kubectl
+          image: docker.io/bitnami/kubectl:1.14.1
+          imagePullPolicy: IfNotPresent
+          command:
+            - /bin/sh
+            - -c
+            - {{printf "kubectl delete secret sumologic -n %s;" .Release.Namespace }}
```
sumo-drosiek, line 22:
```suggestion
            - {{ printf "kubectl delete secret sumologic -n %s;" .Release.Namespace }}
```

deploy/helm/sumologic/templates/setup/setup-cleanup.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: collection-sumologic-secret-cleanup
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "3"
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ template "sumologic.metadata.name.setup.roles.serviceaccount" . }}
```
sumo-drosiek, line 14: I checked if setup-service-account is available and seems that it's removed after the setup job is done

```
Error creating: pods "collection-sumologic-secret-cleanup-" is forbidden: error looking up service account default/collection-sumologic-setup: serviceaccount "collection-sumologic-setup" not found
```

deploy/helm/sumologic/templates/setup/setup-cleanup.yaml
```diff
@@ -0,0 +1,24 @@
...
+      containers:
+        - name: kubectl
+          image: docker.io/bitnami/kubectl:1.14.1
```
perk-sumo, line 17: I'd just go with `latest`. Also `docker.io` can go away.
```suggestion
          image: bitnami/kubectl
```

deploy/helm/sumologic/templates/setup/setup-cleanup.yaml
sumo-drosiek: Also the `helm delete` failed at all (I don't see way to remove collection now). We need to fix that and check what happen if the container fails itself for some reason.

frankreno: The service account is surprising.  This is a pre-delete hook that runs before any resources are deleted in k8s. at least that is how it should behave per the docs.

sumo-drosiek: I suppose they are deleted together with setup-job when it's done

vsinghal13:
Looks like we would need new `service account`, `clusterrole` and `clusterrolebinding` specific to cleanup. The ones created for the setup job are deleted after the hook completion. The default `serviceaccount` does not have the required permissions to access and modify secrets. Therefore, I want with this approach where these are only related to the cleanup activity.

sumo-drosiek:
@vsinghal13 Please do not forget to add and use templates for cleanup roles and labels 😉

vsinghal13:
closing this as it is a candidate for operator

PR #722, opened 2020-06-09
Fix potential tf race confition

### LINKED ISSUE #720: Potential Race Condition in Setup
While I was going through the TF code used for setup, I noticed a potential race condition: The Kubernetes `Secret` that is created by Terraform does not depend on the `Namespace` that Terraform will also create in the same run. In most cases this is _probably_ fine, given that the default parallelism for Terraform is 10 and the Collectors take longer to create than a `Namespace`, but it should be fixed regardless:

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/d68ccf787d142cb3f5255728ef8fceef2b3862eb/deploy/helm/sumologic/conf/setup/sumo-k8s.tf#L98 should instead be something like this:

```terraform
namespace = kubernetes_namespace.sumologic_collection_namespace.metadata[0].name
```

sumo-drosiek:
###### Description

Fix potential tf race confition
Fixes #720 

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/setup/resources.tf
+++ b/deploy/helm/sumologic/conf/setup/resources.tf
@@ -27,7 +27,7 @@ resource "kubernetes_namespace" "sumologic_collection_namespace" {
 resource "kubernetes_secret" "sumologic_collection_secret" {
   metadata {
     name = "sumologic"
-    namespace = var.namespace_name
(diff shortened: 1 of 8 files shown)


sumo-drosiek:
This PR is made on the top of the #721

Issue #724, opened 2020-06-09
Logs from Systemd do not preserve originating host in _sourceHost


frankreno:
The _sourceHost metadata field in Sumo Logic should be set to the originating host for systemd logs.

PR #725, opened 2020-06-10
set correct source host for systemd logs

samjsong:
###### Description

After much investigation I found that the `sumologic` output plugin makes the http request for systemd logs with the `X-Sumo-Host` set to empty string. So even after my investigation I have no idea where that ip was coming from 😅 

but this change will set the correct one, as Frank said before, by reading `_HOSTNAME` from the record

###### Testing performed

- [x] ci/build.sh


samjsong:
solves https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/724

PR #731, opened 2020-06-15
Fix helm chart requirements conditions

sumo-drosiek:
###### Description

Multiple conditions works as below:
1. If first value is specified, it decides if the requirement is installed or not
2. If it's not, the second is taken under consideration

due to point 1 we should change order of conditions and comment the `prometheus-operator.enabled`

so for prometheus:

If `prometheus-operator.enabled` is not set, the `metrics.enabled` decides if we enable prometheus or not, otherwise `prometheus-operator.enabled` is decisive

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -9,7 +9,7 @@ resources: {}
 #   memory: 8Mi
 
 ## Set the enabled flag to false to not install fluent-bit helm chart as a dependency along with this helm chart
-# enabled: true
+# enabled: false
 service:
   flush: 5
 metrics:
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -8,6 +8,8 @@ resources: {}
 #   cpu: 10m
 #   memory: 8Mi
 
+## Set the enabled flag to false to not install fluent-bit helm chart as a dependency along with this helm chart
+# enabled: true
 service:
   flush: 5
 metrics:
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -8,7 +8,9 @@ resources: {}
 #   cpu: 10m
 #   memory: 8Mi
 
-## Set the enabled flag to false to not install fluent-bit helm chart as a dependency along with this helm chart
+## Uncomment the flag below to not install fluent-bit helm chart as a dependency along with this helm chart.
+## Comment it out if you wish to have the fluent-bit dependency installed.
+## Setting the flag to true instead of commenting out will break other functionality. 
 # enabled: false
 service:
   flush: 5
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -34,10 +34,8 @@ nodeExporter:
     interval:
 # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
 kubeTargetVersionOverride: 1.13.0-0
-## Set the enabled flag to false for either of the below two purposes:
-## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-## 2. Disable metrics collection altogether
-enabled: true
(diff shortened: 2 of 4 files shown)


deploy/helm/fluent-bit-overrides.yaml
```diff
@@ -8,6 +8,8 @@ resources: {}
 #   cpu: 10m
 #   memory: 8Mi
 
+## Set the enabled flag to false to not install fluent-bit helm chart as a dependency along with this helm chart
+# enabled: true
```
perk-sumo, line 12:
```suggestion
# enabled: false
```

deploy/helm/prometheus-overrides.yaml
```diff
@@ -34,10 +34,8 @@ nodeExporter:
     interval:
 # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
 kubeTargetVersionOverride: 1.13.0-0
-## Set the enabled flag to false for either of the below two purposes:
-## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-## 2. Disable metrics collection altogether
-enabled: true
+## Set the enabled flag to false to not install prometheus operator helm chart as a dependency along with this helm chart
+# enabled: true
```
perk-sumo, line 38:
```suggestion
# enabled: false
```

deploy/helm/sumologic/values.yaml
```diff
@@ -608,7 +608,8 @@ fluent-bit:
     #   cpu: 10m
     #   memory: 8Mi
 
-  enabled: true
+  ## Set the enabled flag to false to not install fluent-bit helm chart as a dependency along with this helm chart
+  # enabled: true
```
perk-sumo, line 612:
```suggestion
  # enabled: false
```

deploy/helm/sumologic/values.yaml
```diff
@@ -765,10 +766,8 @@ prometheus-operator:
       interval:
   # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
   kubeTargetVersionOverride: 1.13.0-0
-  ## Set the enabled flag to false for either of the below two purposes:
-  ## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-  ## 2. Disable metrics collection altogether
-  enabled: true
+  ## Set the enabled flag to false to not install prometheus operator helm chart as a dependency along with this helm chart
+  # enabled: true
```
perk-sumo, line 770:
```suggestion
  # enabled: false
```

deploy/helm/sumologic/values.yaml
```diff
@@ -608,7 +608,8 @@ fluent-bit:
     #   cpu: 10m
     #   memory: 8Mi
 
-  enabled: true
+  ## Set the enabled flag to false to not install fluent-bit helm chart as a dependency along with this helm chart
```
perk-sumo, line 611:
```suggestion
## Uncomment the flag below to not install fluent-bit helm chart as a dependency along with this helm chart.
## Comment it out if you wish to have the fluent-bit dependency installed.
## Setting the flag to true instead of commenting out will break other functionality. 
```

deploy/helm/sumologic/values.yaml
```diff
@@ -765,10 +766,8 @@ prometheus-operator:
       interval:
   # Ensure we use pre 1.14 recording rules consistently as current content depends on them.
   kubeTargetVersionOverride: 1.13.0-0
-  ## Set the enabled flag to false for either of the below two purposes:
-  ## 1. Not install prometheus operator helm chart as a dependency along with this helm chart
-  ## 2. Disable metrics collection altogether
-  enabled: true
+  ## Set the enabled flag to false to not install prometheus operator helm chart as a dependency along with this helm chart
```
perk-sumo, line 769:
```suggestion
## Uncomment the flag below to not install fluent-bit helm chart as a dependency along with this helm chart.
## Comment it out if you wish to have the fluent-bit dependency installed.
## Setting the flag to true instead of commenting out will break other functionality. 
```

sumo-drosiek:
https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies
```
Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.
```

PR #734, opened 2020-06-16
Use sumologic.com as prefix for internal labels

sumo-drosiek:
###### Description

I observed during review that we already use 'sumologic.com/' prefix in annotations so it would be better to stick that convention for internal labels

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
I want to put it into `1.1` so shouldn't be any

PR #754, opened 2020-06-30
Fill cluster name property for traces

pmm-sumo:
###### Description

`cluster` property was not filled for traces previously. This fixes it, using the new tracing taxonomy (so it goes to `k8s.cluster.name`)

###### Testing performed

- [ ] ci/build.sh


deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -12,5 +12,6 @@
 {{- $excludePodRegex := .Values.fluentd.logs.containers.excludePodRegex | quote }}
 {{- $excludeContainerRegex := .Values.fluentd.logs.containers.excludeContainerRegex | quote }}
 {{- $excludeHostRegex := .Values.fluentd.logs.containers.excludeHostRegex | quote }}
+{{- $clusterName := .Values.sumologic.clusterName | replace " " "-" }}
```
frankreno, line 15: To have consistency with logs and metrics, you should use this one as logs and metrics use this one - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/_helpers.tpl#L317

pmm-sumo: Fair point! I attempted doing that originally but chart had some errors while rendering, though now I see it was a simple mistake and the helper attribute works here. Thank you!

deploy/helm/sumologic/conf/traces/traces.otelcol.conf.yaml
```diff
@@ -12,5 +12,6 @@
...
\ No newline at end of file
+{{- $yamlFile | replace "processors.source.collector.replace" $_collector | replace "exporters.zipkin.url_replace" $endpoint | replace "processors.source.name.replace" $sourceName | replace "processors.source.category.replace" $sourceCategory | replace "processors.source.category_prefix.replace" $sourceCategoryPrefix | replace "processors.source.category_replace_dash.replace" $sourceCategoryReplaceDash | replace "processors.source.exclude_namespace_regex.replace" $excludeNamespaceRegex | replace "processors.source.exclude_pod_regex.replace" $excludePodRegex | replace "processors.source.exclude_container_regex.replace" $excludeContainerRegex | replace "processors.source.exclude_host_regex.replace" $excludeHostRegex | replace "processors.resource.cluster.replace" $clusterName | nindent 2 }}
```
sumo-drosiek, line 17: nit: split it on several lines to make reviews easier in the future

pmm-sumo: @sumo-drosiek totally agree, tried to split into several lines but must have been doing it wrong and the chart never rendered; do you know the right way to do it?

sumo-drosiek: I believe something like that:
```
{{- $yamlFile = replace "processors.source.collector.replace" $_collector $yamlFIle }}
```

PR #768, opened 2020-07-09
Use retry_max_interval instead of retry_timeout for fluentd buffer

sumo-drosiek:
###### Description

Use retry_max_interval instead of retry_timeout for fluentd buffer

https://docs.fluentd.org/configuration/buffer-section#retries-parameters

```
retry_timeout [time]
Default: 72h
The maximum seconds to retry to flush while failing, until plugin discards buffer chunks

retry_forever [bool]
Default: false
If true, plugin will ignore retry_timeout and retry_max_times options and retry flushing forever

retry_max_interval [time]
Default: none
The maximum interval seconds for exponential backoff between retries while failing
```

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -208,6 +208,7 @@ fluentd:
     chunkLimitSize: "1m"
     queueChunkLimitSize: 128
     totalLimitSize: "128m"
+    retryTimeout: "72h"
     retryMaxInterval: "1h"
     retryForever: true
 
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -208,7 +208,8 @@ fluentd:
     chunkLimitSize: "1m"
     queueChunkLimitSize: 128
     totalLimitSize: "128m"
-    retryTimeout: "1h"
+    retryMaxInterval: "1h"
+    retryForever: true
 
     ## File paths to buffer to, if Fluentd buffer type is specified as file above.
     ## Each sumologic output plugin buffers to its own unique file.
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -208,7 +208,8 @@ fluentd:
     chunkLimitSize: "1m"
     queueChunkLimitSize: 128
     totalLimitSize: "128m"
-    retryTimeout: "1h"
+    retryMaxInterval: "1h"
```
perk-sumo, line 211: I have a feeling this should be lower than 1h.
Does it hurt to retry sending every 15 or 10 min?

sumo-drosiek: shouldn't too much

samjsong:
I like `retry_max_interval`, we should use it 😄 
But is there a reason we cannot use `retry_max_interval` together with `retry_timeout` ? Maybe we can expose all three as options, maybe with `retry_forever` set to true by default, but give user the option to set a `retry_timeout` if they choose

perk-sumo:
@samjsong yeah, we could do that - let's keep the `retry_timeout` and change it from 1h the original default 72h

deploy/helm/sumologic/values.yaml
```diff
@@ -208,7 +208,9 @@ fluentd:
     chunkLimitSize: "1m"
     queueChunkLimitSize: 128
     totalLimitSize: "128m"
-    retryTimeout: "1h"
+    retryTimeout: "72h"
+    retryMaxInterval: "1h"
```
perk-sumo, line 212:
```suggestion
    retryMaxInterval: "10m"
```

deploy/helm/sumologic/values.yaml
frankreno: I think lower than one hour is good so that once things get resolved downstream, we are not waiting the full hour to recover.

perk-sumo:
@samjsong the `retry_timeout` can be changed using the extraConf so maybe let's not expose it (thanks @frankreno for pointing that out :))

PR #775, opened 2020-07-13
Upgrade fluentd base image, refactor docker file to pull steps into builder image, remove dependencies that are not needed.

frankreno:
Upgrade fluentd base image, refactor docker file to pull steps into builder image, remove dependencies that are not needed.


deploy/docker/Dockerfile
```diff
@@ -67,10 +56,19 @@ RUN mkdir /tmp/terraform && \
     chmod +x /bin/terraform && \
     rm -rf /tmp/terraform
 
+# Start with fresh image
+FROM fluent/fluentd:v1.11.1-debian-1.0
+
+USER root
```
perk-sumo, line 62: I wonder if we could drop the `root` user afer all of the apt-gets, mkdirs and chowns.

frankreno:
Interesting. The build script passes locally. I'll take a look.

deploy/docker/Dockerfile
frankreno: We don't run as root, we run the image as the fluent user.

frankreno:
CI Is passing now after removing tracing components.  @sumo-drosiek @perk-sumo please confirm I did not miss anything in the removal.

tests/tracing/static/collection-monitoring-false.input.yaml
```diff
@@ -1,6 +0,0 @@
-sumologic:
```
sumo-drosiek, line 1: This is test for otelcol, we shouldn't remove it

tests/tracing/static/simple.output.yaml
```diff
@@ -1,114 +0,0 @@
...
-kind: ConfigMap
-metadata:
-  name: RELEASE-NAME-sumologic-otelcol
-  labels:
-    app: RELEASE-NAME-sumologic-fluentd-metrics
-    chart: "sumologic-1.1.0-beta.2"
-    release: "RELEASE-NAME"
-    heritage: "Helm"
-data:
-  traces.otelcol.conf.yaml: |-
```
sumo-drosiek, line 13: This is test for otelcol, we shouldn't remove it

frankreno: But these tests reference the configmap, which is now removed?

sumo-drosiek: it references otelcol configmap

frankreno: Ahh, got it.  Ok fixed this.

frankreno:
Closing in favor of #817

deploy/docker/Dockerfile
perk-sumo: Ah, I see :) 
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/775/files#diff-aa72c1ddeec59a31f30834d956165473R78

Issue #749, opened 2020-06-26
Add Source Category customization options for default logs source


njparisi:
Similar to the sourceCategory and sourceCategoryPrefix option for the other sources we should add the same options for the default/catch all source.

PR #780, opened 2020-07-16
add source category for default logs pipeline

vsinghal13:
###### Description

Solves #749 

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek:
@vsinghal13 
Shouldn't we add it to the `logs.source.containers.conf`? Maybe create helm function for this plugin

vsinghal13:
> @vsinghal13
> Shouldn't we add it to the `logs.source.containers.conf`? Maybe create helm function for this plugin

Dominik, the `logs.source.containers.conf` file already uses the `kubernetes_sumologic` filter plugin. This si primarily added for the default(catch-all) section.

Issue #740, opened 2020-06-17
Set limits for job pod used during install


alexbklein91:
Some clusters/namespaces have restrictions on only allowing pods to be deployed if they have cpu/mem limits. It would be great to set limits on the job pod that is created as part of the deploy.

PR #782, opened 2020-07-17
add resource limits to setup job pod

vsinghal13:
###### Description

Solves #740 

###### Testing performed

- [ ] ci/build.sh


deploy/helm/sumologic/values.yaml
```diff
@@ -79,6 +79,13 @@ sumologic:
         helm.sh/hook-weight: "2"
         helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
     job:
+      resources:
```
frankreno, line 82: Have we tested that these are reasonable limits? The setup job really doesn't do a lot, so these seem like they might be high?

pmalek-sumo: I've tried to look at setup job's utilization and I've noticed values of used memory around 90-120Mi.

vsinghal13: Thanks @pmalek-sumo for verifying that. I think then the request limit of `128 Mi` is apt.

pmalek-sumo (approved):

perk-sumo (approved):
👍

sumo-drosiek (approved):

PR #816, opened 2020-08-03
Enable file persistence by default

vsinghal13:
###### Description

Enabled fluentd file persistence by default along with a warning message if the property `fluentd.persistence.enabled` is set to `false`

###### Testing performed

- [ ] ci/build.sh


vsinghal13:
> LGTM, but related question
> 
> 
> 
> Do we want to enable persistence for prometheus?

As discussed, we won't be enabling persistence on prometheus.

PR #819, opened 2020-08-04
add custom labels to fluentd sts and setup job pods

vsinghal13:
###### Description

- Users will be able to add labels to all fluentd pods and setup job.
- There is a common property to apply to all fluentd pods together, as well as the ability to add to each individually.
- expose the ability to add labels on fluent-bit and Prometheus

Configs:
- `sumologic.podLabels` -> add labels to fluentd sts, setup job, otelcol deployment
- `fluentd.podLabels` -> add labels to all fluentd sts.
- `fluentd.logs.statefulset.podLabels` -> add labels only to logs fluentd sts.
- `fluentd.metrics.statefulset.podLabels` -> add labels only to metrics fluentd sts.
- `fluentd.events.statefulsetpodLabels` -> add labels only to events fluentd sts.
- `otelcol.deployment.podLabels` -> add labels only to otelcol deployment.
- `sumologic.setup.job.podLabels` -> add labels only to setup job.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-job.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-job.yaml
@@ -9,7 +9,9 @@ metadata:
   labels:
     app: {{ template "sumologic.labels.app.setup.job" . }}
     {{- include "sumologic.labels.common" . | nindent 4 }}
+{{- if .Values.sumologic.setup.job.podLabels }}
 {{ toYaml .Values.sumologic.setup.job.podLabels | indent 4 }}
+{{- end }}
 spec:
   template:
     spec:
--- a/deploy/helm/sumologic/templates/setup/setup-job.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-job.yaml
@@ -9,6 +9,9 @@ metadata:
   labels:
     app: {{ template "sumologic.labels.app.setup.job" . }}
     {{- include "sumologic.labels.common" . | nindent 4 }}
+{{- if .Values.sumologic.podLabels }}
+{{ toYaml .Values.sumologic.podLabels | indent 4 }}
+{{- end }}
 {{- if .Values.sumologic.setup.job.podLabels }}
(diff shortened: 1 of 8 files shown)


deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -9,6 +9,9 @@ metadata:
   labels:
     app: {{ template "sumologic.labels.app.setup.job" . }}
     {{- include "sumologic.labels.common" . | nindent 4 }}
+{{- if .Values.sumologic.setup.job.podLabels }}
```
frankreno, line 12: I would add the common set here as well i think

sumo-drosiek:
We should add this to the otelcol as well 😅

frankreno:
Good call on OTELCOL, we should add them there too as it will be needed.

deploy/helm/sumologic/values.yaml
```diff
@@ -742,6 +759,8 @@ fluent-bit:
 ## Configure prometheus-operator
 ## ref: https://github.com/helm/charts/blob/master/stable/prometheus-operator/values.yaml
 prometheus-operator:
+  ## Labels to apply to all pormetheus operator resources
```
samjsong, line 762:
```suggestion
  ## Labels to apply to all prometheus operator resources
```

sumo-drosiek:
Customers should be aware that changing the labels for existing installation need to be done with `--force` flag

cc: @frankreno

perk-sumo:
@sumo-drosiek I'm trying to reproduce the error with annotations and can't do it, I don't need to provide the `--force` to change them.

sumo-drosiek:
@perk-sumo even if you use change value of the label during upgrade?

perk-sumo:
Yes, I've tested this with latest 1.2.3 and see no issue with fresh install nor upgrade.

sumo-drosiek:
@perk-sumo, confirming 👍

PR #820, opened 2020-08-05
Add docs on creating processing rules with setup job.

frankreno:
Add docs on creating processing rules with setup job.


frankreno:
@sumo-drosiek removing the rule from values.yaml and upgrading does remove the rule because we do run the setup job on the upgrade as well.

@vsinghal13 - good call, added

PR #835, opened 2020-08-19
Fixing the indentation of the processing rule example

justrelax19:
###### Description
The indentation was little off and therefore on uncommenting the code and doing helm upgrade, it was giving the YAML file error.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -122,12 +122,12 @@ sumologic:
       default:
         name: logs
         config-name: endpoint-logs
-        # properties can be used to extend default settings, such as processing rules, fields etc
-        #properties:
-          #filters:
-            #- name: "Test Exclude Debug"
-            #filter_type: "Exclude"
-            #regexp: ".*DEBUG.*"
+       # properties can be used to extend default settings, such as processing rules, fields etc
+       # properties:
+         # filters:
+           # - name: "Test Exclude Debug"
+           #   filter_type: "Exclude"
+           #   regexp: ".*DEBUG.*"
     events:
       default:
         name: events


deploy/helm/sumologic/values.yaml
```diff
@@ -122,12 +122,12 @@ sumologic:
...
-          #filters:
-            #- name: "Test Exclude Debug"
-            #filter_type: "Exclude"
-            #regexp: ".*DEBUG.*"
+       # properties can be used to extend default settings, such as processing rules, fields etc
+       # properties:
+         # filters:
+           # - name: "Test Exclude Debug"
+           #   filter_type: "Exclude"
+           #   regexp: ".*DEBUG.*"
```
perk-sumo, line 130: Good call!
But I like the idea of keeping this in line better, can we do it like the above suggestion?
```suggestion
        
        ## Properties can be used to extend default settings, such as processing rules, fields etc
        #properties:
        #  filters:
        #    - name: "Test Exclude Debug"
        #      filter_type: "Exclude"
        #      regexp: ".*DEBUG.*"

```

perk-sumo (changes requested):
Good find! But let's keep the comment hashes at the same level as this needs to be uncommented as a whole block.

perk-sumo:
I think there should not be any spaces between the comment hash and the commented key (`# key:` vs `#key:`) as this might be misleading.

pmalek-sumo:
> I think there should not be any spaces between the comment hash and the commented key (`# key:` vs `#key:`) as this might be misleading.

As far as I can see (et least `values.yaml`) we're using `# key:` approach (for example it's the same with telegraf's configs if that's relevant here) and I'd say we should stick with it.

Nevertheless I'm all ears to hear pros and cons of both approaches.

deploy/helm/sumologic/values.yaml
sumo-drosiek: what about `#properties:` ?

perk-sumo:
> we're using `# key:` approach

Actually we're using both:
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L20
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L44

I like the `#key:` better because for me there is no doubt what needs to be deleted (only `#`? or maybe `# `?)
But it doesn't look as good as `# key:`.
Anyway having both in one file does not help, let's fix that in `1.3`.

However we chose we should make it clear at the top, like I propose in this PR for comments starting with `##` or more characters:
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/836

deploy/helm/sumologic/values.yaml
perk-sumo: `#properties:` is a root key for this whole block

perk-sumo:
I've created new issue for `#key:` vs `# key:` and the whole file unification: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/837

PR #843, opened 2020-08-21
Fix incorrectly indented podAnnotations

pmalek-sumo:
###### Description

Previously add `podAnnotations` weren't properly indented everywhere producing things like:

```
# Source: sumologic/templates/setup/setup-job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: RELEASE-NAME-sumologic-setup
  namespace: default
  annotations:
    helm.sh/hook: pre-install,pre-upgrade
    helm.sh/hook-weight: "3"
  annotationforall: value1 ## <--- This should be indented with 2 more spaces
  labels:
    app: RELEASE-NAME-sumologic
    chart: "sumologic-1.2.0-beta.1"
    release: "RELEASE-NAME"
    heritage: "Helm"
spec:
  template:
    metadata:
      labels:
    spec:
...
```

This PR addresses that.

Additionally it fixes annotations in `deploy/helm/sumologic/templates/setup/setup-job.yaml` by moving it from `metadata.annotations` to `metadata.spec.template.metadata.annotations`.

To test the solution, I've perfomed not `helm template` as before which doesn't validate the yaml against the cluster but with `helm template --validate` to check the templates and yaml against the running k8s cluster.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/setup/setup-job.yaml
+++ b/deploy/helm/sumologic/templates/setup/setup-job.yaml
@@ -4,17 +4,20 @@ kind: Job
 metadata:
   name: {{ template "sumologic.metadata.name.setup.job" . }}
   namespace: {{ .Release.Namespace }}
-  annotations:
-{{ include "sumologic.annotations.app.setup.helmsh" "3" | indent 4 }}
-{{- if .Values.sumologic.podAnnotations }}
-{{ toYaml .Values.sumologic.podAnnotations | indent 2 }}
-{{- end }}
   labels:
     app: {{ template "sumologic.labels.app.setup.job" . }}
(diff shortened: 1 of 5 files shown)


deploy/helm/sumologic/templates/setup/setup-job.yaml
```diff
@@ -4,17 +4,20 @@ kind: Job
 metadata:
   name: {{ template "sumologic.metadata.name.setup.job" . }}
   namespace: {{ .Release.Namespace }}
-  annotations:
-{{ include "sumologic.annotations.app.setup.helmsh" "3" | indent 4 }}
```
sumo-drosiek, line 8: The helmsh annotations shouldn't be moved to spec

PR #850, opened 2020-08-24
Set text compress as default

sumo-drosiek:
###### Description

Set text compress as default due to https://github.com/fluent/fluentd/issues/3110

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -211,6 +211,9 @@ fluentd:
     totalLimitSize: "128m"
     retryMaxInterval: "10m"
     retryForever: true
+    ## usage of the gzip compression is highly not recommended due to fluentd issue:
+    ## rel: https://github.com/fluent/fluentd/issues/3056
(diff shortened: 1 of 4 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -211,6 +211,7 @@ fluentd:
     totalLimitSize: "128m"
     retryMaxInterval: "10m"
     retryForever: true
+    compress: text
```
frankreno, line 214: Should we add a comment that GZIP is not supported so it is clear why we set it this way?

sumo-drosiek: @frankreno I would say not recommended and point to the issue. Is it ok?

sumo-drosiek:
@vsinghal13 it was 10 times for my test data. It will be compression rate of the gzip for the given log files

perk-sumo (commented):
LGTM, thanks!

PR #854, opened 2020-08-26
attach node metadata to metrics when not present

samjsong:
###### Description

Attach node metadata to metrics when it is not already present.

###### Testing performed

- [x] ci/build.sh


sumo-drosiek (approved):
Looks fine

samjsong:
I've added a unit test for this case, as well as some basic tests that we never had for the "end to end" situation (for the `filter` method in this plugin). Please re-review when you get the chance :)

vsinghal13 (approved):

sumo-drosiek (approved):

Issue #862, opened 2020-08-28
Thanos sidecar has no limits set


perk-sumo:
On of containers is not limited by default - and it should be as sumo warns about not-limited containers.

Container name: thanos-sidecar
StatefulSet name: prometheus-collection-prometheus-oper-prometheus
Helm chart version affected: all

perk-sumo:
There are no limits for `thanos` in the prometheus-operator values: https://github.com/helm/charts/blob/master/stable/prometheus-operator/values.yaml#L1961

vsinghal13:
User can set it as follows:
https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#thanosspec

perk:
Yeah, the thing is:
1) sumo k8s app warns about not-limited containers
2) thanos sidecar is not limited by default in the upstream chart like others

So you install the collection and instantly get the warning about collection, that's not a good experience.

PR #874, opened 2020-08-31
define resources for thanos sidecar

vsinghal13:
###### Description

Solves #862 
Ref: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#thanosspec
###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -165,6 +165,8 @@ prometheus:
     thanos:
       baseImage: quay.io/thanos/thanos
       version: v0.10.0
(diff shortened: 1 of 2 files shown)


perk (changes requested):
We should add the limits by ourselves as the upstream chart is not limited.

perk-sumo (changes requested):
We should add the limits by ourselves as the upstream chart is not limited.

vsinghal13:
Thanks @perk-sumo for adding the limits.

perk-sumo (approved):
👍

PR #890, opened 2020-09-08
move metrics server chart repo

vsinghal13:
###### Description
- changed the repo for metrics-server chart
- add `4.3.1` version as a dependency

Ref: https://hub.helm.sh/charts/bitnami/metrics-server
###### Testing performed

- [ ] ci/build.sh
- [x] Tested on personal cluster


samjsong (approved):

deploy/helm/metrics-server-overrides.yaml
```diff
@@ -1,4 +1,6 @@
 # This file is auto-generated.
-args:
-- --kubelet-insecure-tls
-- --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
+apiService:
+  create: true
+extraArgs:
+  kubelet-insecure-tls: true
+  kubelet-preferred-address-types: InternalIP,ExternalIP,Hostname
```
sumo-drosiek, line 6: Looks like breaking change

vsinghal13: the `args` are the same, just the key has changed from `args` to `extraArgs` and the `apiServer.create` was set to `true` by default in the earlier chart, so we are explicitly setting it true here.

Also, since we do not enable `metrics-server` by default, I think it won't be considered a breaking change. cc @frankreno

frankreno: So because of the key change, if customers have enabled and created a new values.yaml, would they need to make a change? The key to breaking change is if customers old values.yaml works w/o changes.

frankreno (approved):
Let's sort out the breaking change comment first, then we can approve

deploy/helm/metrics-server-overrides.yaml
vsinghal13: yes, since the key to providing `args` is changed, if the customer has enabled it, then they will have to modify the `values.yaml`

vsinghal13:
Changed the milestone to `2.0` since this will be a breaking change for some customers.

Issue #776, opened 2020-07-14
Custom Metrics From Fluentd are scraped 2x in Prometheus


frankreno:
The additional ServiceMonitors we add to the values.yaml to scrape the custom fluentd metrics use the label `sumologic.com/app` to discover the pods to scrape.

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/values.yaml#L825

The problem is we use this label on both the core service and the headless service:

Main Service - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/events-service.yaml#L8

Headless Service - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/events-service-headless.yaml#L8

They both point to the same function int he TPL file and get the same value.

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/_helpers.tpl#L292

As a result, Prometheus sees this as 2 distinct targets, which means we are scraping the data 2x more than we expect.

PR #895, opened 2020-09-09
Don't scrape from headless services

pmalek-sumo:
###### Description

This prevents prometheus double scraping fluentd pods by removing the annotations from headless services.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/events-service-headless.yaml
+++ b/deploy/helm/sumologic/templates/events-service-headless.yaml
@@ -5,7 +5,6 @@ metadata:
   name: {{ template "sumologic.metadata.name.events.service-headless" . }}
   labels:
     app: {{ template "sumologic.labels.app.events.service-headless" . }}
-    {{- include "sumologic.labels.events" . | nindent 4 }}
     {{- include "sumologic.labels.common" . | nindent 4 }}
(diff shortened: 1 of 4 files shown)


pmalek-sumo:
Fixes https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/776

sumo-drosiek (commented):
I think thats the correct solution. We could eventually add additional label for scraping metrics
@perk thoughts?

perk-sumo:
Yeah, I think that works for now but let's fix it properly in `2.0`: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/896

perk-sumo (approved):

vsinghal13 (approved):

vsinghal13:
If we change the labels now, will it be considered a breaking change?

frankreno (approved):
LGTM

sumo-drosiek (changes requested):
Valid point from @vsinghal13 
@pmalek-sumo could you verify that?

sumo-drosiek (approved):
Verified, no need for `--force` and change is not considered as breaking change

PR #900, opened 2020-09-11
add performance doc

samjsong:
###### Description

Based on Vaibhav's performance tests.

I wasn't sure where to link to this doc, just adding it for now. Anybody have any suggestions? :)

### DIFF
--- a/deploy/docs/Performance.md
+++ b/deploy/docs/Performance.md
@@ -3,28 +3,26 @@
 For larger or more volatile loads, we recommend [enabling Fluentd autoscaling](./Best_Practices.md#Fluentd-Autoscaling), as this will allow Fluentd to automatically scale to support your data volume. However, the following examples will help you get an idea of the resources required to run collection on your cluster.
 
 ## Up to 500 application pods
-Our test cluster had 70 nodes running 500 application pods, each generating 128KB/s logs and 1200 DPM metrics.
+Our test cluster had 70 nodes running an average of 500 application pods, each generating either 128KB/s logs or 2400 DPM metrics. The application pods had about 20% churn rate.
 
-Data type | Rate per pod | # pods   | Total rate
---------- | ------------ | -------- | ----------
-Logs      | 128 KB/s     | 500 pods | 64 MB/s = **5.5 TB/day**
-Metrics   | 1200 DPM     | 500 pods | **600K DPM**
+Data type | Rate per pod | Min # pods | Max # pods | Max Total rate
+--------- | ------------ | ---------- | ---------- | --------------
+Logs      | 128 KB/s     | 50 pods    | 400 pods   | **4.3 TB/day**
+Metrics   | 2400 DPM     | 100 pods   | 450 pods   | **1.3M DPM** (including non-application metrics)
 
 We observed **35 Fluentd-logs** pods and **25 Fluentd-metrics** pods were sufficient for handling this load with the default resource limits.
 
-Prometheus memory consumption reached a maximum of **28GiB**.
+Prometheus memory consumption reached a maximum of **28GiB**, with an average of **16GiB**.
 
 ## Up to 2000 application pods
-Our test cluster had 210 nodes running 200 application pods, each generating 128KB/s logs and 1200 DPM metrics.
+Our test cluster had 210 nodes running an average of 2000 application pods, each generating either 128KB/s logs or 2400 DPM metrics. The application pods had about 10% churn rate.
 
-Data type | Rate per pod | # pods   | Total rate
---------- | ------------ | -------- | ----------
-Logs      | 128 KB/s     | 2000 pods | 256 MB/s = **22 TB/day**
-Metrics   | 1200 DPM     | 2000 pods | **2.4M DPM**
+Data type | Rate per pod | Min # pods | Max # pods | Max Total rate
+--------- | ------------ | ---------- | ---------- | --------------
+Logs      | 128 KB/s     | 900 pods   | 1875 pods  | **20 TB/day**
+Metrics   | 2400 DPM     | 125 pods   | 1100 pods  | **3.2M DPM** (including non-application metrics)
+
 
 We observed **135 Fluentd-logs** pods and **100 Fluentd-metrics** pods were sufficient for handling this load with the default resource limits. Additionally the **Fluentd-events** pod had to be given a 1 GiB memory limit to accommodate the increased events load.
 
-Prometheus memory consumption reached a maximum of **60GiB**.
-
-
-
+Prometheus memory consumption reached a maximum of **60GiB**, with an average of **45GiB**.
--- a/deploy/docs/Performance.md
+++ b/deploy/docs/Performance.md
@@ -19,11 +19,15 @@ For larger or more volatile loads, we recommend [enabling Fluentd autoscaling](.
       remoteTimeout: 1s
     ```
 5. For clusters with 2000 application pods, we found that the **Fluentd-events** pod had to be given a 1 GiB memory limit to accommodate the increased events load. If you find that the **Fluentd-events** pod is being OOMKilled, please increase the memory limits and requests accordingly.
-6. For our log generating test application pods, we found that increasing the IOPS to 375 minimum improved stability.
+6. For our log generating test application pods, we found that increasing the IOPS to 300 minimum improved stability.
 
 ### Up to 500 application pods
 
-Our test cluster had 70 nodes running an average of 500 application pods, each generating either 128KB/s logs or 2400 DPM metrics. The application pods had about 20% churn rate.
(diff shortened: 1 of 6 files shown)


vsinghal13 (commented):
Thanks, Sam! One suggestion would be to add the per fluentd replica capacity in terms of gb/day or DPM or provide some sort of formula for the same. This will help the customers get a better idea for the number of fluentd's required according to their load.

frankreno (approved):
LGTM! Same feedback on including the per replica load so users can easily apply to their loads. Also I wanna make this document more discoverable.  I am thinking on each of the various installation, helm, non-helm side-by-side, etc, it would be good to have some reference here.  Even a simple note. thoughts?

samjsong:
Thanks Frank! That's exactly what I was looking for when I asked about where to link this doc 🙂 I've gone ahead and addressed comments, thanks team

sumo-drosiek (approved):
I like it 👍

deploy/docs/Performance.md
```diff
@@ -0,0 +1,47 @@
...
+    ```
+    - url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.node
+      writeRelabelConfigs:
+      - action: keep
+        regex: node-exporter;(?:node_load1|node_load5|node_load15|node_cpu_seconds_total)
+        sourceLabels: [job, __name__]
+      remoteTimeout: 1s
+    ```
+5. For clusters with 2000 application pods, we found that the **Fluentd-events** pod had to be given a 1 GiB memory limit to accommodate the increased events load. If you find that the **Fluentd-events** pod is being OOMKilled, please increase the memory limits and requests accordingly.
+6. For our log generating test application pods, we found that changing the IOPS to 375 or 125GiB improved stability.
```
pmalek-sumo, line 22: This is not super clear to me:

> ... we found that changing the **IOPS to 375 or 125GiB** improved ... 

changing IOPS to 375 makes sense but to 125GiB? Isn't there a word missing by any chance?

deploy/docs/Performance.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Up to 500 application pods
```
pmalek-sumo, line 24: nitpicks:

* ideally there should be an empty line directly below a heading, otherwise a [MD022](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md022) is triggered when using `markdown` lint :)
* `Recommendations` section has `##` - the same as this here. IMHO `Up to 500 application pods` should be nested underneath `Recommendations` with `###` (same for `Up to 2000 application pods`)

deploy/docs/Performance.md
perk-sumo: Hm that refers to the EBS volume size on AWS I believe.

deploy/docs/Performance.md
perk-sumo:
```suggestion
## Up to 500 application pods

```

perk-sumo (commented):
I think we should add information about the cluster configuration as well (which EC2 instances were used, disk size (125GiB), disks type (GP2), IOPS (125) etc.

If that changed between the cluster sizes we should add this info as well.

deploy/docs/Performance.md
pmalek-sumo: Then IMHO we should rephrase it to e.g. 

```suggestion
6. For our log generating test application pods, we found that changing the IOPS to 375 and the volume size to 125GiB improved stability.
```

pmalek-sumo (approved):

deploy/docs/Performance.md
```diff
@@ -0,0 +1,57 @@
+# Performance
+
+For larger or more volatile loads, we recommend [enabling Fluentd autoscaling](./Best_Practices.md#Fluentd-Autoscaling), as this will allow Fluentd to automatically scale to support your data volume. However, the following recommendations and corresponding examples will help you get an idea of the resources required to run collection on your cluster.
+
+## Recommendations
+
+1. At least **8 Fluentd-logs** pods per **1 TB/day** of logs.
+2. At least **4 Fluentd-metrics** pods per **120k DPM** of metrics.
```
justrelax19, line 8: Hey @samjsong , have we written 4 Fluentd-metrics by considering the scrape interval as well? Because I want to understand the numbers 120k and 4 fluentd metrics.
AFAIR, one fluentd-metrics is able to support around 60K DPM, so if we are considering the scrape interval of 30sec, then this 4 fluentd is fine but if now, can you tell me the number calculation of 4 fluentd metrics and 120K DPM

deploy/docs/Performance.md
```diff
@@ -0,0 +1,57 @@
...
+      remoteTimeout: 1s
+    ```
+5. For clusters with 2000 application pods, we found that the **Fluentd-events** pod had to be given a 1 GiB memory limit to accommodate the increased events load. If you find that the **Fluentd-events** pod is being OOMKilled, please increase the memory limits and requests accordingly.
+6. For our log generating test application pods, we found that increasing the IOPS to 300 minimum improved stability.
+
+### Up to 500 application pods
+
+Our test cluster had 70 nodes (AWS m5a.2xlarge instances).
+
+We used 125 GiB GP2 volumes, which allowed for IOPS of 375. In addition we attached a 300GiB volume to the Prometheus pod.
```
justrelax19, line 28: We can mention for the point `In addition we attached a 300GiB volume to the Prometheus pod` that, since we were not forwarding the metrics to sumo and we were analyzing the metrics in grafana, so we need to store the metrics somewhere and therefore we used 300GiB volume

deploy/docs/Performance.md
```diff
@@ -0,0 +1,57 @@
...
+        regex: node-exporter;(?:node_load1|node_load5|node_load15|node_cpu_seconds_total)
+        sourceLabels: [job, __name__]
+      remoteTimeout: 1s
+    ```
+5. For clusters with 2000 application pods, we found that the **Fluentd-events** pod had to be given a 1 GiB memory limit to accommodate the increased events load. If you find that the **Fluentd-events** pod is being OOMKilled, please increase the memory limits and requests accordingly.
+6. For our log generating test application pods, we found that increasing the IOPS to 300 minimum improved stability.
+
+### Up to 500 application pods
+
+Our test cluster had 70 nodes (AWS m5a.2xlarge instances).
```
justrelax19, line 26: So, out of the 70 nodes, the fluentds were running on AWS m4.large nodes.
m5a.2xlarge instances was for the prometheus and for the application pods .

IIn 500 and 2000 pods, the fluentds and k8s collection pods were m4.large instances only.

deploy/docs/Performance.md
```diff
@@ -0,0 +1,57 @@
+# Performance
+
+For larger or more volatile loads, we recommend [enabling Fluentd autoscaling](./Best_Practices.md#Fluentd-Autoscaling), as this will allow Fluentd to automatically scale to support your data volume. However, the following recommendations and corresponding examples will help you get an idea of the resources required to run collection on your cluster.
+
+## Recommendations
+
+1. At least **8 Fluentd-logs** pods per **1 TB/day** of logs.
+2. At least **4 Fluentd-metrics** pods per **120k DPM** of metrics.
+3. The Prometheus pod will use on average **2GiB** memory per **120k DPM**; however in our experience this has gone up to **5GiB**, so we recommend allocating ample memory resources for the Prometheus pod if you wish to collect a high volume of metrics for a larger cluster.
+4. For clusters with 500 application pods or greater, we found the following configuration changes in the [`values.yaml`](./../helm/sumologic/values.yaml) file to lead to a more stable experience:
```
justrelax19, line 10: This is applicable for 2000 application pods as well

deploy/docs/Performance.md
samjsong: Right, since from Sumo backend perspective the customer is billed for 120k DPM rather than for 60k DPM at 30s scrape interval, I thought it would make it easier to understand if we kept just one number for the DPM

deploy/docs/Performance.md
samjsong: Agreed, I think 500 application pods or greater" should be enough to cover both of the tested cases

deploy/docs/Performance.md
samjsong: That's true, my opinion was that we could stick with the larger node size for simplicity but I agree we can be more accurate - I will make the change

deploy/docs/Performance.md
samjsong: Ah, that's a good point. In this case I think we can actually just remove this point, because it will not be needed for the customer since they will be sending their metrics to Sumo.

deploy/docs/Performance.md
justrelax19: Correct.

vsinghal13 (approved):

PR #903, opened 2020-09-11
Don't mount and don't define pos volumes for fluentd

pmalek-sumo:
###### Description

We're not using fluentd's [`tail` input plugin](https://docs.fluentd.org/input/tail) so don't define pos files volumes.

###### Testing performed

- [ ] ci/build.sh


sumo-drosiek (approved):
🎈

samjsong (approved):

vsinghal13 (approved):

PR #912, opened 2020-09-15
Increase cache refresh interval to 1 hour

astencel-sumo:
The cache TTL was increased as well because it always needs
to be greater than the refresh interval, to prevent a situation where
there is no cached value for a resource.


astencel-sumo:
Thanks @pmalek-sumo for this remark, the PR description was probably not detailed enough. Hopefully the below will be more descriptive, please let me know.

This PR increases default values for two parameters of the `enhance_k8s_metadata` plugin for Fluentd:
- `cache_refresh` - increased from 30 minutes to 1 hour,
- `cache_ttl` - increased from 1 hour to 2 hours.

The `cache_refresh` value has been increased because we want to perform less calls to k8s API server (or make the calls less often). 

The `cache_ttl` value has been increased because we want this value to be always bigger than the `cache_refresh` value, to make sure the cache is refreshed sooner than it goes stale. This is to make sure the cache never goes stale and is always up-to-date.

pmalek-sumo (approved):
So this change will decrease the number of API calls by a half. The question is: will this be enough for the user or should we investigate further?

astencel-sumo:
We should definitely investigate further. This PR is just a start.

Issue #914, opened 2020-09-15
 Setup process does not work with external Proxy


frankreno:
When your proxy is running in the k8s cluster, say squid proxy, the setup process can work correctly. However, when using an external proxy it cannot.  The issue stems from the Kubernetes provider that creates the secret.

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/conf/setup/setup.sh#L9-L10

We use the above environment variables, which are leveraged by the underlying Go HTTP client that TF providers use.  The problem is when using an external proxy, the request to create the Kubernetes secret gets routed to the proxy and when it is an external proxy, it cannot resolve to the correct URL.

sumo-drosiek:
I believe using `NO_PROXY` should solve this issue: https://github.com/hashicorp/terraform/issues/17386#issuecomment-370276866
Going to perform tests and prepare PR for that

PR #920, opened 2020-09-16
Do not proxy kubernetes internal traffic

sumo-drosiek:
###### Description

Do not proxy kubernetes internal traffic

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -32,6 +32,8 @@ sumologic:
   ## proxy urls
   httpProxy: ""
   httpsProxy: ""
+  ## Exclude kubernetes internal traffic from proxy
+  noProxy: kubernetes.default.svc
 
   ## Collector name
   #collectorName: ""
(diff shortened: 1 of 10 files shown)


sumo-drosiek:
Fixes #914

deploy/helm/sumologic/values.yaml
```diff
@@ -32,6 +32,8 @@ sumologic:
   ## proxy urls
   httpProxy: ""
   httpsProxy: ""
+  ## Exclude kubernetes internal traffic from proxy
+  noProxy: kubernetes.default.svc
```
frankreno, line 36: Love this approach it is cleaner.  Con is that if a customer has a different url where they customize the host in the TF kubernetes provider, they have to also update this.  Any way to have one property to control both?

sumo-drosiek: Should the cluster endpoint be excluded from the proxy every time?
I can merge both of the config options in helm, but there will be no way to include the kubernetes url then

frankreno: Ahh ok fair point. In that case we can leave it

frankreno (approved):
Lgtm

pmalek-sumo (approved):

PR #922, opened 2020-09-17
Fix grafana installation in vagrant

sumo-drosiek:
Fix grafana installation
```
Error: UPGRADE FAILED: template: sumologic/charts/prometheus-operator/charts/grafana/templates/deployment.yaml:46:10: executing "sumologic/charts/prometheus-operator/charts/grafana/templates/deployment.yaml" at <include "grafana.pod" .>: error calling include: template: sumologic/charts/prometheus-operator/charts/grafana/templates/_pod.tpl:75:18: executing "grafana.pod" at <.Values.sidecar.image.sha>: can't evaluate field sha in type interface {}
```

### DIFF
--- a/vagrant/values.yaml
+++ b/vagrant/values.yaml
@@ -19,6 +19,9 @@ prometheus-operator:
       prometheusSpec:
         externalLabels:
           cluster: microk8s
+    sidecar:
+      image:
+        sha: ''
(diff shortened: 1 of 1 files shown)


perk-sumo (approved):
👍

PR #924, opened 2020-09-17
Add nginx to the vagrant environment

sumo-drosiek:
Add nginx to the vagrant environment


pmalek-sumo (approved):
Just an idea: since the Makefile is growing with the addition of applications I'd maybe split it into parts, e.g. main part and include others like nginx, redis etc

sumo-drosiek:
@pmalek-sumo I plan to do it together with moving commands to the scripts

perk-sumo (approved):
👍

samjsong (approved):

vagrant/Makefile
```diff
@@ -101,3 +101,38 @@ create-secrets:
 # patch context to use sumologic namespace as default
 patch-context:
 	kubectl config set-context --current --namespace "${namespace}"
+
```
justrelax19, line 104: Is there no need of `nginx-ingress-k8s` now as there were commands for this as well in the old Makefile

justrelax19 (approved):

vagrant/Makefile
sumo-drosiek: Good point. We are decided to support official nginx for now, so there is no need to install the kubernetes' one. Anyway, I will prepare separate PR for that

PR #936, opened 2020-09-22
Increase memory limits for thanos-sidecar to 32Mi to prevent restarts

pmalek-sumo:
###### Description

Using `values.yaml` from the repository without any scaling changes it seems that `thanos-sidecar` gets OOMKilled every ~~15min.
This PR addresses that by increasing the memory limit to `32Mi`.

With that it seems that memory usage stabilize at around `18Mi` for the "default" setup.

image

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -170,7 +170,7 @@ prometheus:
       resources:
         limits:
           cpu: 10m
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):

perk-sumo (approved):
👍

PR #945, opened 2020-09-24
Install shellcheck with apt in vagrant

pmalek-sumo:
It seems that it isn't as easy as we might think:
`shellcheck`'s snap doesn't have access to all filesystems 

```
snap connections shellcheck
Interface        Plug                        Slot   Notes
home             shellcheck:home             :home  -
removable-media  shellcheck:removable-media  -      -
```

and it doesn't seem that it's easy to allow it to access (in our case) `/sumologic`

https://askubuntu.com/questions/1033344/how-to-give-snaps-access-to-somedir

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/vagrant/provision.sh
+++ b/vagrant/provision.sh
@@ -79,8 +79,7 @@ done
 # install requirements for ci/build.sh
 snap install ruby --channel=2.6/stable --classic
 gem install bundler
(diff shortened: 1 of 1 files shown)


astencel-sumo (approved):

sumo-drosiek (approved):
😕

PR #950, opened 2020-09-24
add crio parser to fluent-bit

vsinghal13:
###### Description

The default `cri` parser defined in fluent-bit uses `message` as the key, instead of `log`. This PR defines a custom `crio` parser which uses the `log` key instead, so that the Sumo Logic apps do not break.

image (24)
Example of `CRI`format log with `message` key attached.
 
###### Testing performed

- [ ] ci/build.sh


sumo-drosiek (approved):
LGTM
Do we have an example log from the cri?

perk-sumo (approved):
👍

PR #952, opened 2020-09-25
Introduce cache refresh variation

astencel-sumo:
###### Description

This PR introduces a random variation in cache refresh interval for each pod, allowing the load on API server to be spread in time.

By default, there are multiple Fluentd pods running in a cluster, each of the pods maintaining its own version of pod metadata cache. The cache is being refreshed by each Fluentd pod independently in regular intervals specified by the `fluentd.metadata.cacheRefresh` configuration variable. Before this change, the cache refresh interval was equal for all pods, causing all of them to flood API server with queries in approximately the same time.

The introduction of `fluentd.metadata.cacheRefreshVariationRange` configuration variable makes it possible to spread the load on API server by changing the cache refresh interval by a different, randomly assigned value for each pod. For example, if the `cacheRefresh` value is 1 hour and `cacheRefreshVariationRange` is 30 minutes, the actual cache refresh interval for each pod will vary between 45 minutes and 1 hour 15 minutes.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/README.md
+++ b/deploy/helm/sumologic/README.md
@@ -60,6 +60,7 @@ Parameter | Description | Default
 `fluentd.metadata.cacheSize` | Option to control the enabling of metadata filter plugin cache_size. | `10000`
 `fluentd.metadata.cacheTtl` | Option to control the enabling of metadata filter plugin cache_ttl (in seconds). | `7200`
 `fluentd.metadata.cacheRefresh` | Option to control the interval at which metadata cache is asynchronously refreshed (in seconds). | `3600`
+`fluentd.metadata.cacheRefreshVariation` | Option to control the variation in seconds by which the cacheRefresh option is changed for each pod separately. For example, if cache refresh is 1 hour and variation is 15 minutes, then actual cache refresh interval will be a random value between 45 minutes and 1 hour 15 minutes, different for each pod. This helps spread the load on API server that the cache refresh induces. Setting this to 0 disables cache refresh variation. | `900`
 `fluentd.metadata.pluginLogLevel` | Option to give plugin specific log level. | `error`
 `fluentd.logs.enabled` | Flag to control deploying the Fluentd logs statefulsets. | `true`
 `fluentd.logs.statefulset.nodeSelector` | Node selector for Fluentd log statefulset. | `{}`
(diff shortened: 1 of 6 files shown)


sumo-drosiek (changes requested):
Could we change name to `cache refresh variation` (without `range`) ?

perk-sumo (approved):
I'm 👍  except for the `variationRange` -> `variation` rename.

perk-sumo (changes requested):
Let's change the `variationRange` into `variation` in the code as well co that 30min variation means +/- 30min and not +/- 15min

deploy/helm/sumologic/README.md
```diff
@@ -60,6 +60,7 @@ Parameter | Description | Default
 `fluentd.metadata.cacheSize` | Option to control the enabling of metadata filter plugin cache_size. | `10000`
 `fluentd.metadata.cacheTtl` | Option to control the enabling of metadata filter plugin cache_ttl (in seconds). | `7200`
 `fluentd.metadata.cacheRefresh` | Option to control the interval at which metadata cache is asynchronously refreshed (in seconds). | `3600`
+`fluentd.metadata.cacheRefreshVariation` | Option to control the variation in seconds by which the cacheRefresh option is changed for each pod separately. For example, if cache refresh is 1 hour and variation is 15 minutes, then actual cache refresh interval will be a random value between 45 minutes and 1 hour 15 minutes, different for each pod. This helps spread the load on API server that the cache refresh induces. Setting this to 0 disables cache refresh variation. | `1800`
```
perk-sumo, line 63:
```suggestion
`fluentd.metadata.cacheRefreshVariation` | Option to control the variation in seconds by which the cacheRefresh option is changed for each pod separately. For example, if cache refresh is 1 hour and variation is 15 minutes, then actual cache refresh interval will be a random value between 45 minutes and 1 hour 15 minutes, different for each pod. This helps spread the load on API server that the cache refresh induces. Setting this to 0 disables cache refresh variation. | `900`
```

perk-sumo (approved):
LGTM but please squash :)

vsinghal13 (approved):

sumo-drosiek (approved):

PR #953, opened 2020-09-25
Use local changes when building and running locally

astencel-sumo:
This is a poor man's try at adding the following improvements:
- When building locally, push the built Docker image to local registry and use it when running locally.
- When building locally, use Docker cache.

Suggestions for improvements are welcome.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -106,7 +106,9 @@ bundle_fluentd_plugins "${VERSION}" || (echo "Failed bundling fluentd plugins" &
 
 echo "Building docker image with $DOCKER_TAG:local in $(pwd)..."
 cd ./deploy/docker || exit 1
-docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" --no-cache
+readonly no_cache=$([ "$DOCKER_USE_CACHE" == "true" ] && echo "" || echo "--no-cache")
+# shellcheck disable=SC2086
+docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" ${no_cache}
 rm -f ./gems/*.gem
 cd ../.. || exit 1
 
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -106,9 +106,11 @@ bundle_fluentd_plugins "${VERSION}" || (echo "Failed bundling fluentd plugins" &
 
 echo "Building docker image with $DOCKER_TAG:local in $(pwd)..."
 cd ./deploy/docker || exit 1
-readonly no_cache=$([ "$DOCKER_USE_CACHE" == "true" ] && echo "" || echo "--no-cache")
-# shellcheck disable=SC2086
-docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" ${no_cache}
+no_cache="--no-cache"
+if [[ "$DOCKER_USE_CACHE" == "true" ]]; then
+  no_cache=""
+fi
+docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" ${no_cache:+"--no-cache"}
 rm -f ./gems/*.gem
 cd ../.. || exit 1
 
--- a/vagrant/Makefile
+++ b/vagrant/Makefile
@@ -17,6 +17,9 @@ namespace = sumologic
 build:
 	cd ${root_dir} && ${root_dir}ci/build.sh
 
+generate-local-config:
+	${root_dir}vagrant/scripts/generate-local-config.sh
+
 build-local:
 	DOCKER_USE_CACHE=true \
 	${mkfile_path} \
--- a/vagrant/Makefile
+++ b/vagrant/Makefile
@@ -15,15 +15,8 @@ name = collection
 namespace = sumologic
 
 build:
-	cd ${root_dir} && ${root_dir}ci/build.sh
-
-generate-local-config:
-	${root_dir}vagrant/scripts/generate-local-config.sh
-
-build-local:
-	DOCKER_USE_CACHE=true \
+	cd ${root_dir} && DOCKER_USE_CACHE=true ${root_dir}ci/build.sh
 	${mkfile_path} \
-		build \
 		tag-and-push-local
 
 tag-and-push-local:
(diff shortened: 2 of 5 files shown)


ci/build.sh
```diff
@@ -106,7 +106,9 @@ bundle_fluentd_plugins "${VERSION}" || (echo "Failed bundling fluentd plugins" &
 
 echo "Building docker image with $DOCKER_TAG:local in $(pwd)..."
 cd ./deploy/docker || exit 1
-docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" --no-cache
+readonly no_cache=$([ "$DOCKER_USE_CACHE" == "true" ] && echo "" || echo "--no-cache")
+# shellcheck disable=SC2086
```
sumo-drosiek, line 110: what was that?

sumo-drosiek: i think it should be `if [[ ...`

vagrant/scripts/generate-local-values-file.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cat > /sumologic/vagrant/values.local.yaml <<END
+fluentd:
+  metadata:
+    pluginLogLevel: "info"
```
sumo-drosiek, line 6: I doubt we should add this to the auto generated file

vagrant/scripts/generate-local-values-file.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cat > /sumologic/vagrant/values.local.yaml <<END
```
sumo-drosiek, line 3: We shouldn't override existing file

vagrant/Makefile
```diff
@@ -44,10 +57,11 @@ helm-dependencies:
 	helm dependency update "${helm_path}"
 
 helm-upgrade:
+	$(eval local_values_param := $(shell [ -f ${local_values_file} ] && echo "-f ${local_values_file}" || echo ""))
```
sumo-drosiek, line 60: `eval` and `shell` 🙀

vagrant/scripts/generate-local-values-file.sh
```diff
@@ -0,0 +1,11 @@
...
+fluentd:
+  metadata:
+    pluginLogLevel: "info"
+image:
+  repository: localhost:32000/sumologic/kubernetes-fluentd
+  tag: local
+  pullPolicy: Always
+END
```
sumo-drosiek, line 11: Add target for this script to Makefile

sumo-drosiek (changes requested):
See comments

vagrant/scripts/generate-local-values-file.sh
perk-sumo: Hm why not? When you build stuff you override the old build dir.

sumo-drosiek: I want to keep some configuration for test without need to change it with every build. There would be no difference with using `vagrant/values.yaml` if we keep overriding this file automatically. Anyway this is not happening right now as I see. You have to manually fire the script. Its's not automated anywhere

ci/build.sh
```diff
@@ -106,7 +106,9 @@ bundle_fluentd_plugins "${VERSION}" || (echo "Failed bundling fluentd plugins" &
...
+docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" ${no_cache}
```
perk-sumo, line 111: per https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions
```suggestion
docker build . -f ./Dockerfile -t "$DOCKER_TAG:local" ${no_cache:+"no_cache"}
```

vagrant/scripts/generate-local-values-file.sh
perk-sumo: I think it's nice to have developer settings for the development values yaml file.

sumo-drosiek: I do not agree. every feature development requires another set of the settings and they may conflict with each other

vagrant/Makefile
astencel-sumo: This seems to be the way to define variables evaluated at rule execution time: https://stackoverflow.com/questions/1909188/define-make-variable-at-rule-execution-time. The Makefile's eval function: https://www.gnu.org/software/make/manual/make.html#Eval-Function

ci/build.sh
astencel-sumo: Done

vagrant/scripts/generate-local-values-file.sh
astencel-sumo: Removed the `pluginLogLevel` config entry.

ci/build.sh
astencel-sumo: @perk-sumo please see the change

sumo-drosiek (approved):

sumo-drosiek (approved):
🐇

pmalek-sumo (approved):

vagrant/Makefile
```diff
@@ -16,6 +17,21 @@ namespace = sumologic
...
+build-local:
+	DOCKER_USE_CACHE=true \
+	${mkfile_path} \
+		build \
+		tag-and-push-local
+
+tag-and-push-local:
+	$(eval timestamp := $(shell date +%s))
+	docker tag sumologic/kubernetes-fluentd:local localhost:32000/sumologic/kubernetes-fluentd:local-${timestamp}
+	docker push localhost:32000/sumologic/kubernetes-fluentd:local-${timestamp}
```
perk-sumo, line 32: Can we puch the `local` tag as well?

pmalek-sumo: What would be the use case for it?

vagrant/Makefile
```diff
@@ -16,6 +17,21 @@ namespace = sumologic
...
+	sed -i 's/^\([[:space:]]\+\)tag:.*$$/\1tag: local-${timestamp}/' ${local_values_file}
```
perk-sumo, line 33: Please change to`yq`

PR #965, opened 2020-09-29
rename hpa metrics

vsinghal13:
###### Description

The `kube-state-metrics` helm chart  currently uses the `v1.9.7` image  which still uses `hpa` instead of `horizontalpodautoscaler`.

These will be renamed as part of `v2.0.0` of kube-state-metrics.

`v2.0.0 Alpha` release notes: https://github.com/kubernetes/kube-state-metrics/releases

Ref PR: https://github.com/kubernetes/kube-state-metrics/pull/1003

###### Testing performed

- [ ] ci/build.sh


frankreno (approved):
LGTM - when we switch over to 2.0 we can adjust.

Issue #885, opened 2020-09-04
Prometheus container has no limits set by default


perk-sumo:

perk-sumo:
@frankreno we need to figure out how much memory/CPU we want to give here. I'd say for the small cluster, WDYT?

frankreno:
I think @pmaciolek was doing some benchmarking on this? or maybe @vsinghal13 ? we need to benchmark what it uses. I would say the size of the cluster should have little impact on the setup.  Its doing the same setup no matter what cluster.

frankreno:
just make sure we set for max usage, not average.

perk-sumo:
But max usage would be 128GB or sth.

frankreno:
Hah, I misread this, this is for Prometheus (not setup). I would say based on a small cluster and documented that it may need adjustments.

perk-sumo:
It should be based on our findings from: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/900

sumo-drosiek:
#959

perk-sumo:
The fix has been reverted and will be reintroduced in `2.0`

perk-sumo:
This made into 1.3 eventually: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/978

PR #971, opened 2020-10-02
fix: remove adding new resources limits and requests from 1.3

perk-sumo:
I propose to remove resources from Prometheus in `1.3` release and reintroduce it in `2.0`.
It has been added as a fix for https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/885

I see adding `resources` for Prometheus were wrong values can break the solution as a breaking change.

The same story with `setup job` and for `Thanos`.

###### Testing performed

- [x] ci/build.sh

### DIFF
--- a/deploy/helm/prometheus-overrides.yaml
+++ b/deploy/helm/prometheus-overrides.yaml
@@ -155,13 +155,13 @@ prometheus:
       labels: {}
       annotations: {}
     ## Define resources requests and limits for single Pods.
-    resources:
-      limits:
-        cpu: 2000m
-        memory: 8Gi
-      requests:
-        cpu: 1000m
-        memory: 2Gi
+    resources: {}
+    # limits:
+    #   cpu: 2000m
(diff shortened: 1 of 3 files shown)


sumo-drosiek (approved):

deploy/helm/prometheus-overrides.yaml
```diff
@@ -171,9 +171,9 @@ prometheus:
         limits:
           cpu: 10m
           memory: 32Mi
-        requests:
-          cpu: 1m
-          memory: 8Mi
+          # requests:
+          #   cpu: 1m
+          #   memory: 8Mi
```
sumo-drosiek, line 176: ❓

astencel-sumo (approved):
We could just remove the data instead of commenting out.

perk-sumo:
> We could just remove the data instead of commenting out.

@astencel-sumo this is a configuration file so it serves as an example.

PR #974, opened 2020-10-02
Extract unit tests from build.sh and run as separate build stage

sumo-drosiek:
###### Description

Extract unit tests from build.sh

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,18 @@ before_install:
   - sudo curl https://github.com/mikefarah/yq/releases/download/3.2.1/yq_linux_amd64 -L -o /usr/local/bin/yq-3.2.1
   - sudo chmod +x /usr/local/bin/yq-3.2.1
   - sudo ln -s /usr/local/bin/yq-3.2.1 /usr/local/bin/yq
-  - sudo apt install shellcheck
-script: bash ci/build.sh
+  - curl -Lo- https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.linux.x86_64.tar.xz | tar -xJf -
+  - sudo cp shellcheck-v0.7.1/shellcheck /usr/local/bin
+  - rm -rf shellcheck-v0.7.1/
 branches:
   only:
   - master
   - /^release-v\d+\.\d+$/
   - /^v\d+\.\d+\.\d+$/
   - /^v\d+\.\d+\.\d+-(alpha|beta|rc)\.\d+$/
+jobs:
+  include:
+    - stage: tests
+      script: bash ci/tests.sh
+    - stage: build
+      script: bash ci/build.sh
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
+# shellcheck disable=SC2154
 VERSION="${TRAVIS_TAG:-0.0.0}"
+# shellcheck disable=SC2154
 VERSION="${VERSION#v}"
 : "${DOCKER_TAG:=sumologic/kubernetes-fluentd}"
 : "${DOCKER_USERNAME:=sumodocker}"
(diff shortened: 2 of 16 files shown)


sumo-drosiek:
image

ci/build.sh
```diff
@@ -18,6 +18,31 @@ function get_branch_to_checkout() {
...
 
+function bundle_fluentd_plugin() {
+  local plugin_name="${1}"
+  local version="${2}"
+  # Strip everything after "-" (longest match) to avoid gem prerelease behavior
+  local gem_version="${version%%-*}"
+
+  cd "${plugin_name}" || exit 1
+
+  if [[ "${version}" == "" ]] ; then
```
pmalek-sumo, line 29: `== ""` vs `-z` ? Not sure which one we prefer (a couple of other scripts use `-z`)

ci/build.sh
```diff
@@ -26,52 +51,60 @@ function bundle_fluentd_plugins() {
...
+  echo "Pushing docker image $DOCKER_TAG:$version..."
+  echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+  docker push "$DOCKER_TAG:$version"
+}
+
+function push_helm_chart() {
+  local version="$1"
+
+  echo "Pushing new Helm Chart release $version"
+  set -x
```
pmalek-sumo, line 89: Do we want to keep it?

ci/build.sh
```diff
@@ -82,43 +115,34 @@ if [ -n "$GITHUB_TOKEN" ] && [ "$TRAVIS_EVENT_TYPE" == "pull_request" ] && [[ !
...
-fi
+  echo "Building docker image with ${tag}:local in $(pwd)..."
+  cd ./deploy/docker || exit 1
+  no_cache="--no-cache"
+  if [[ "$DOCKER_USE_CACHE" == "true" ]]; then
+    no_cache=""
+  fi
+  docker build . -f ./Dockerfile -t "${tag}:local" ${no_cache:+"--no-cache"}
+  rm -f ./gems/*.gem
+  cd ../.. || exit 1
```
pmalek-sumo, line 133: Maybe use a subshell or even better `pushd`+`popd`?
I'm not a huge fan of relative paths like `../..` in scripts

ci/run-tests.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -e
+
+VERSION="${TRAVIS_TAG:-0.0.0}"
+VERSION="${VERSION#v}"
+
+readonly ROOT_DIR="$(dirname "$(dirname "${0}")")"
+cd "${ROOT_DIR}"
+
+function test_fluentd_plugin() {
```
pmalek-sumo, line 10: Can we add a comment to functions like this that they will change CWD so the caller has to account for that?

ci/run-tests.sh
```diff
@@ -0,0 +1,65 @@
...
+  find . -name 'fluent-plugin-*' -type 'd' -print |
+    while read -r line; do
+      # Run tests in their own context
+      (test_fluentd_plugin "$(basename "${line}")" "${version}") || exit 1
+    done
+}
+
+## check the build scripts with shellcheck
+echo "Checking the build script with shellcheck..."
+shellcheck ci/build.sh
```
pmalek-sumo, line 51: Add option to enable all checks, my suggestion:

```suggestion
# Enable all checks but exclude SC2154 which complains about undeclared but used variables
# that we get from our CI environment e.g. ${TRAVIS_EVENT_TYPE} and ${TRAVIS_PULL_REQUEST_BRANCH}
shellcheck --enable all --exclude SC2154 ci/build.sh
shellcheck --enable all --exclude SC2154 ci/run-tests.sh
```

ci/run-tests.sh
```diff
@@ -0,0 +1,65 @@
...
+shellcheck ci/run-tests.sh
```
pmalek-sumo, line 52: Add option to enable all checks, my suggestion:

```suggestion
# Enable all checks but exclude SC2154 which complains about undeclared but used variables
# that we get from our CI environment e.g. ${TRAVIS_EVENT_TYPE} and ${TRAVIS_PULL_REQUEST_BRANCH}
shellcheck --enable all --exclude SC2154 ci/build.sh
shellcheck --enable all --exclude SC2154 ci/run-tests.sh
```

ci/run-tests.sh
```diff
@@ -0,0 +1,65 @@
...
+  fi
+
+  find . -name 'fluent-plugin-*' -type 'd' -print |
+    while read -r line; do
+      # Run tests in their own context
+      (test_fluentd_plugin "$(basename "${line}")" "${version}") || exit 1
+    done
+}
+
+## check the build scripts with shellcheck
```
pmalek-sumo, line 49: This comment says pretty much the same thing as the `echo` below

ci/build.sh
perk-sumo: Yeah, that's very helpful when debugging helm release issues.

ci/run-tests.sh
perk-sumo: Rather than excluding we should just take care of that and set those variables to some default, eg.
```TRAVIS_EVENT_TYPE="${TRAVIS_EVENT_TYPE:-push}"```

File: vagrant/Makefile [hunk 8]
```diff
@@ -14,6 +14,9 @@ local_values_file = ${root_dir}vagrant/values.local.yaml
 name = collection
 namespace = sumologic
 
+test:
+	${root_dir}ci/run-tests.sh
```
perk-sumo, line 18: I'm not entirely sold on this name, I think this should be just `test.sh`

ci/run-tests.sh
pmalek-sumo: I disagree. `TRAVIS_EVENT_TYPE` doesn't make sense when not on travis and this script is not intended **ONLY** to be run on travis.

vagrant/Makefile
perk-sumo: ... or `tests.sh`, just not `run-*`.

ci/run-tests.sh
perk-sumo: The function should not change `CWD`.

ci/build.sh
perk-sumo: Hm `-z` is not a look-and-know whereas `== ""` is 🤔

perk-sumo (commented):
overall 👍 but some comments needs answer

ci/run-tests.sh
perk-sumo: If it's not intended then those variables should be extracted somewhere else.
Anyway disabling globally shellcheck rule for something that is easily fixed is not a good idea IMO.

pmalek-sumo: Let's disable it in lines before the "known issues"

ci/build.sh
sumo-drosiek: I go with the `-z` as it was designed for that usage 🤷

sumo-drosiek: look-and-know doesn't apply to the bash anyway 🤔

vagrant/Makefile
sumo-drosiek: I'll go with `tests.sh`

ci/run-tests.sh
sumo-drosiek: So if I must change `CWD` I should ensure that during going out from the function the `CWD` will be restored?

pmalek-sumo: I'd say yes

ci/run-tests.sh
sumo-drosiek:
```
Checking the build script with shellcheck...
unrecognized option `--enable'

Usage: shellcheck [OPTIONS...] FILES...
  -e CODE1,CODE2..  --exclude=CODE1,CODE2..  exclude types of warnings
  -f FORMAT         --format=FORMAT          output format
  -C[WHEN]          --color[=WHEN]           Use color (auto, always, never)
  -s SHELLNAME      --shell=SHELLNAME        Specify dialect (sh,bash,dash,ksh)
  -x                --external-sources       Allow 'source' outside of FILES.
  -V                --version                Print version information
```

pmalek-sumo: :/ Let's leave it then without `--enable`

ci/run-tests.sh
sumo-drosiek: so @perk what is better in your opinion
```
# Didn't test that but should work
{ bundle exec rake && popd; } || { popd && exit 1; }
```

vs

```
  bundle exec rake
  result=$?
  popd
  if [ "$result" -eq "0" ]; then
   exit 1
  fi
```

perk-sumo: I'd say the latter as it's more straight forward.

ci/run-tests.sh
sumo-drosiek: I will update shellcheck in vagrant

ci/build.sh
sumo-drosiek: Will go with `pushd` + `popd`

vagrant/provision.sh
```diff
@@ -79,7 +79,11 @@ done
 # install requirements for ci/build.sh
 snap install ruby --channel=2.6/stable --classic
 gem install bundler
-apt install -y gcc g++ libsnappy-dev shellcheck
+apt install -y gcc g++ libsnappy-dev
+
+wget -qO- https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.linux.x86_64.tar.xz | tar -xJf -
```
perk-sumo, line 84: Can we extract the version into VARIABLE like we do for `HELM_2_VERSION` and `HELM_3_VERSION`?

perk-sumo (approved):
👍  except for one comment pending (and commits history cleanup 😉 )

pmalek-sumo:
> 👍 except for one comment pending (and commits history cleanup 😉 )

No squash?

.travis.yml
```diff
@@ -9,11 +9,18 @@ before_install:
   - sudo curl https://github.com/mikefarah/yq/releases/download/3.2.1/yq_linux_amd64 -L -o /usr/local/bin/yq-3.2.1
   - sudo chmod +x /usr/local/bin/yq-3.2.1
   - sudo ln -s /usr/local/bin/yq-3.2.1 /usr/local/bin/yq
-  - sudo apt install shellcheck
-script: bash ci/build.sh
+  - wget -qO- https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.linux.x86_64.tar.xz | tar -xJf -
+  - sudo cp shellcheck-v0.7.1/shellcheck /usr/local/bin
+  - rm -rf shellcheck-v0.7.1/
```
pmalek-sumo, line 14: I believe the only thing I could suggest to rethink is the usage of `curl` in one place and `wget` elsewhere.

I'd personally stick to `curl` wherever possible.

PR #993, opened 2020-10-08
Add scc for setup job

sumo-drosiek:
###### Description

Add scc for setup job

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -163,6 +163,10 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- template "sumologic.labels.app.setup" . }}
 {{- end -}}
 
+{{- define "sumologic.labels.app.setup.securitycontextconstraints" -}}
+{{- template "sumologic.fullname" . }}-setup-scc
+{{- end -}}
+
 {{- define "sumologic.labels.app.podsecuritypolicy" -}}
 {{- template "sumologic.fullname" . }}-psp
 {{- end -}}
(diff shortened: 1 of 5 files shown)


frankreno (approved):
Goof find! LGTM

perk-sumo (approved):
Megusta.

vsinghal13 (commented):
@frankreno I wonder how we were able to deploy the collection on your openshift cluster without hitting this issue? 
Also, I think we don't need an additional `scc` as the exact same `scc` already exists
[here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/scc.yaml) and we can just add the security context to the setup job, unless I am missing something?

frankreno:
> @frankreno I wonder how we were able to deploy the collection on your openshift cluster without hitting this issue?
> Also, I think we don't need an additional `scc` as the exact same `scc` already exists
> [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/scc.yaml) and we can just add the security context to the setup job, unless I am missing something?

When I handed the cluster to you guys, I may have had an SCC in there as I was playing with straight-up YAML outside the helm chart.  I believe the SCC is not created until after the hook completes.  It doesn't have the annotation to install as part of the pre-install hook.

sumo-drosiek:
@frankreno @vsinghal13 
> When I handed the cluster to you guys, I may have had an SCC in there

That is my guess too. When I created scc by hand, the setup has been finished correctly

sumo-drosiek:
> Also, I think we don't need an additional `scc` as the exact same `scc` already exists

I decided to keep separate scc for setup job as we do with other templates

sumo-drosiek:
I tried to use only `setup-scc`, but then only telegraf-operator pod was created

PR #994, opened 2020-10-08
Add possibility to disable metrics coming from OTC

mat-rumian:
**###### Description**

New variable `otelcol.metrics_enabled` gives possibility to disable or enable (false/true) metrics coming from OpenTelemetry Collector. Changes are touching otc-service configuration (ports), otc-deployment (additional argument in command), values.yaml (default value).

**Expected log when metrics are enabled:**
`{"level":"info","ts":1602188423.6246598,"caller":"service/service.go:256","msg":"Setting up own telemetry..."}
{"level":"info","ts":1602188423.6251273,"caller":"service/telemetry.go:100","msg":"Serving Prometheus metrics","address":":8888","legacy_metrics":false,"new_metrics":true,"level":3,"service.instance.id":"c8020a8f-9ccb-4d98-8619-b0ddf1f24331"}
{"level":"info","ts":1602188423.6261094,"caller":"service/service.go:293","msg":"Loading configuration..."}`

**Expected log when metrics are disabled:**
`{"level":"info","ts":1602188357.0610797,"caller":"service/service.go:256","msg":"Setting up own telemetry..."}
{"level":"info","ts":1602188357.0619454,"caller":"service/service.go:293","msg":"Loading configuration..."}
{"level":"info","ts":1602188357.0664148,"caller":"service/service.go:304","msg":"Applying configuration..."}`

**###### Testing performed**

Tested on test cluster.

### DIFF
--- a/deploy/helm/sumologic/templates/otelcol-deployment.yaml
+++ b/deploy/helm/sumologic/templates/otelcol-deployment.yaml
@@ -57,6 +57,9 @@ spec:
         command:
           - "/otelcontribcol"
           - "--config=/conf/traces.otelcol.conf.yaml"
+          {{- if eq .Values.sumologic.traces.otc_metrics.enabled false }}
+          - "--metrics-level=none"
+          {{- end }}
           - {{ printf "--mem-ballast-size-mib=%s" .Values.otelcol.deployment.memBallastSizeMib }}
         env:
         - name: GOGC
--- a/deploy/helm/sumologic/templates/otelcol-deployment.yaml
+++ b/deploy/helm/sumologic/templates/otelcol-deployment.yaml
@@ -57,7 +57,7 @@ spec:
         command:
           - "/otelcontribcol"
           - "--config=/conf/traces.otelcol.conf.yaml"
-          {{- if eq .Values.sumologic.traces.otc_metrics.enabled false }}
+          {{- if eq .Values.otelcol.metrics_enabled false }}
           - "--metrics-level=none"
           {{- end }}
           - {{ printf "--mem-ballast-size-mib=%s" .Values.otelcol.deployment.memBallastSizeMib }}
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -168,6 +168,9 @@ sumologic:
     fluentd_stdout: false
     ## How many spans per request should be send to receiver
     spans_per_request: 100
+    ## Metrics from Traces Collector
+    otc_metrics:
+      enabled: false
 
 fluentd:
   ## Specifies whether a PodSecurityPolicy should be created
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -1588,7 +1588,7 @@ otelcol:
   # To enable collecting all logs, set to false
   logLevelFilter: true
   # Metrics from Collector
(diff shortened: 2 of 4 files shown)


sumo-drosiek:
I would rather put it as `otelcol.metrics_enabled` than `sumologic.traces.otc_metrics.enabled`, because its about otelcol not sumologic per se

pmm-sumo (changes requested):
+1 on Dominik's suggestion. Also, we need to update the [list of params](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/README.md) with this attribute. Otherwise, it's a nice and clean PR 👍

pmm-sumo (approved):
Looks good. I am only wondering if we should `enable` or `disable` it by default... hmm. @sumo-drosiek and @perk - what's your take?

deploy/helm/sumologic/templates/otelcol-deployment.yaml
```diff
@@ -57,6 +57,9 @@ spec:
         command:
           - "/otelcontribcol"
           - "--config=/conf/traces.otelcol.conf.yaml"
+          {{- if eq .Values.otelcol.metrics_enabled false }}
+          - "--metrics-level=none"
+          {{- end }}
```
sumo-drosiek, line 62: What are the other metrics levels?

sumo-drosiek:
@pmaciolek, @mat-rumian How big is expected metrics load from the opentelemetry collector?

deploy/helm/sumologic/templates/otelcol-deployment.yaml
mat-rumian: I didn't find anything about levels description but levels itself are described [here](https://github.com/open-telemetry/opentelemetry-collector/blob/e7e66939263ccbd99c9ac9352960c263d464263c/internal/collector/telemetry/telemetry.go#L48) Looks like to get information what's under which level you have to go thru the code.

pmm-sumo:
> @pmaciolek, @mat-rumian How big is expected metrics load from the opentelemetry collector?

Not sure on DPM, but depending on the setup I think that there are typically ~50 metrics (including histograms)

sumo-drosiek (approved):

deploy/helm/sumologic/values.yaml
```diff
@@ -1587,6 +1587,9 @@ otelcol:
       pullPolicy: IfNotPresent
   # To enable collecting all logs, set to false
   logLevelFilter: true
+  # Metrics from Collector
+  metrics_enabled: true
```
perk-sumo, line 1591: This should be camelCase like the rest of the file.

perk-sumo: Or actually I have better idea, it should be like we have for Fluent Bit:
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/6a0ededa19f44ddd9ea36c96ac0a8be42f2a69b7/deploy/helm/sumologic/values.yaml#L729-L730

perk-sumo: I've created another PR for that one: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1037

perk-sumo:
There is a follow-up PR: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1037

PR #1003, opened 2020-10-13
Add known k8s api callers to vagrant values.yaml and bump falco chart to 1.5.0

pmalek-sumo:
###### Description

* enable falco by default in vagrant
* add known k8s api callers to prevent unnecessary notice logs - https://github.com/falcosecurity/falco/blob/388de273985758afbfc2833885faa24ee5da5703/rules/falco_rules.yaml#L2422-L2436
* bump falco chart from `1.4.0` to `1.5.0` (this bumps falco from `0.25.0` to `0.26.1`)

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/vagrant/values.yaml
+++ b/vagrant/values.yaml
@@ -29,3 +29,31 @@ sumologic:
 telegraf-operator:
   enabled: true
   replicaCount: 1
+
+falco:
+  enabled: true
+  customRules:
+    # Mark the following as known k8s api callers:
+    # * fluentd and its plugins
+    # * grafana sidecar
+    # * terraform provider started in setup job
+    # * telegraf operator
+    # * kube state metrics server
+    # * prometheus
+    # * prometheus operator
+    rules_user_known_k8s_api_callers.yaml: |-
+      - macro: user_known_contact_k8s_api_server_activities
+        condition: >
+          (proc.pcmdline = "fluentd /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins") or
+          (proc.cmdline = "fluentd /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins") or
+          (proc.cmdline = "ruby -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
+          (proc.cmdline = "event_loop -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
+          (proc.cmdline = "watch_endpoints -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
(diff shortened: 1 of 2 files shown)


vagrant/values.yaml
```diff
@@ -29,3 +29,31 @@ sumologic:
...
+    rules_user_known_k8s_api_callers.yaml: |-
+      - macro: user_known_contact_k8s_api_server_activities
+        condition: >
+          (proc.pcmdline = "fluentd /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins") or
+          (proc.cmdline = "fluentd /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins") or
+          (proc.cmdline = "ruby -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
+          (proc.cmdline = "event_loop -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
+          (proc.cmdline = "watch_endpoints -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
+          (proc.cmdline = "watch_events -Eascii-8bit:ascii-8bit /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins --under-supervisor") or
+          (proc.cmdline = "filter_kuberne* /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins") or
```
sumo-drosiek, line 53: Does it make sense to use some regex here? Is it possible?

vagrant/values.yaml
```diff
@@ -29,3 +29,31 @@ sumologic:
...
+    # * grafana sidecar
+    # * terraform provider started in setup job
+    # * telegraf operator
+    # * kube state metrics server
+    # * prometheus
+    # * prometheus operator
+    rules_user_known_k8s_api_callers.yaml: |-
+      - macro: user_known_contact_k8s_api_server_activities
+        condition: >
+          (proc.pcmdline = "fluentd /usr/local/bundle/bin/fluentd -c /fluentd/etc/fluent.conf -p /fluentd/plugins") or
```
perk-sumo, line 47: Can we use both proc names and the namespace?

pmalek-sumo: We could but we'd need to template it because user can configure the destination template for our collection?

pmalek-sumo: For vagrant env for now hardcoding **should** be fine though

vagrant/values.yaml
pmalek-sumo: All I can find regarding regexes in falcos docs is that you can disable sets of rules matching regexes so I guess not

sumo-drosiek: ok, so `filter_kuberne*` is actually process name?

pmalek-sumo: This is exactly verbatim what was in the falco log, copy-pasted.

sumo-drosiek (approved):
Thanks 🚀

PR #1017, opened 2020-10-16
Override Fluent Bit image to fluent/fluent-bit:1.6.0

vsinghal13:
###### Description

Override Fluent Bit image to 1.6.0

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/fluent-bit-overrides.yaml
+++ b/deploy/helm/fluent-bit-overrides.yaml
@@ -1,4 +1,8 @@
 # This file is auto-generated.
+image:
+  fluent_bit:
+    repository: fluent/fluent-bit
+    tag: 1.6.0
 ## Resource limits for fluent-bit
 resources: {}
 # limits:
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -702,6 +702,10 @@ metrics-server:
 ## Configure fluent-bit
 ## ref: https://github.com/helm/charts/blob/master/stable/fluent-bit/values.yaml
 fluent-bit:
+  image:
+    fluent_bit:
+      repository: fluent/fluent-bit
+      tag: 1.6.0
   ## Resource limits for fluent-bit
   resources: {}
     # limits:


deploy/helm/fluent-bit-overrides.yaml
```diff
@@ -66,17 +70,17 @@ rawConfig: |-
...
-    Tag              containers.*
-    Refresh_Interval 1
-    Rotate_Wait      60
-    Mem_Buf_Limit    5MB
-    Skip_Long_Lines  On
-    DB               /tail-db/tail-containers-state-sumo.db
-    DB.Sync          Normal
+    Name               tail
+    Path               /var/log/containers/*.log
+    Docker_Mode        On
```
frankreno, line 75: Is there any risk that doing this will break someones collection or will it all be seamless?

sumo-drosiek: I would recommend to bump to 1.6.0 due to this issue

sumo-drosiek: https://github.com/fluent/fluent-bit/issues/2585

perk-sumo: Hm I don't think we should change the default config - we can do it in 2.0 but I'm very hesitant for 1.3 patch.

perk-sumo: Given that this is a PR against `master` (which is `2.0` by now) I think this is fine but should be documented.

sumo-drosiek: @perk-sumo, @frankreno, @vsinghal13 What is the conclusion for this PR?

vsinghal13: I think we can merge it for now, and when @frankreno  submits the PR for the fluent bit chart migration from helm stable we can revisit and see if we need to remove the override. Thoughts?

frankreno: For 1.3.1 we should only bump the docker image.  Lets bump to 1.6 as @sumo-drosiek suggested to avoid that issue.  Then as @perk-sumo suggests we can use 2.0 to make the config changes as needed and I use that as part of the fluent-bit update PR.

vsinghal13: @frankreno this PR is for `2.0`. The one related to `1.3.1` is already merged with just the docker image bump to `1.6`

frankreno: Ahh I misunderstood the comments now I see ok. I'll have the updated PR for migration done shortly. Tested against the latest chart and all worked. Just doing doc updates

perk-sumo: I've extracted the default Fluent Bit config change into separate PR: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1035

Let's keep this one about bumping to 1.6.

perk-sumo (approved):
I'm good with this bump.

I've extracted the config change into separate PR: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1035

sumo-drosiek (approved):

PR #1054, opened 2020-10-29
Use helm from kubernetes-tools

sumo-drosiek:


ci/build.sh
```diff
@@ -93,20 +101,16 @@ function push_helm_chart() {
   set -x
 
   git checkout -- .
-  sudo helm init --client-only
-  sudo helm repo add falcosecurity https://falcosecurity.github.io/charts
-  sudo helm repo add bitnami https://charts.bitnami.com/bitnami
-  sudo helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
```
pmalek-sumo, line 99: Why we don't need this here but we ned it below? 🤔  in `generate_overrides`

ci/build.sh
```diff
@@ -161,21 +165,19 @@ function generate_overrides() {
...
+
+  # Get list of templates available for default values.yaml
+  with_files=$(find deploy/helm/sumologic/templates -maxdepth 1 -iname "*.yaml" ! -path "*/otelcol-*" ! -path "*/scc.yaml" ! -path "*/*hpa.yaml" ! -path "*/*psp.yaml" | sed 's#deploy/helm/sumologic/templates#-s templates#g' | sed 's/yaml/yaml /g')
   # shellcheck disable=SC2086
   helm template deploy/helm/sumologic \
     ${with_files} \
     --namespace "\$NAMESPACE" \
-    --name collection \
-    --set dryRun=true >> deploy/kubernetes/fluentd-sumologic.yaml.tmpl \
+    --name-template collection \
```
pmalek-sumo, line 179: Why does this parameter change?

ci/build.sh
```diff
@@ -192,13 +194,14 @@ function generate_overrides() {
...
+  # Get list of templates available for default values.yaml
+  with_files=$(find deploy/helm/sumologic/templates/setup -maxdepth 1 -iname "*.yaml" ! -path "*/setup-scc.yaml" | sed 's#deploy/helm/sumologic/templates#-s templates#g' | sed 's/yaml/yaml /g')
   # shellcheck disable=SC2086
   helm template deploy/helm/sumologic \
     ${with_files} \
     --namespace "\$NAMESPACE" \
-    --name collection \
-    --set dryRun=true >> deploy/kubernetes/setup-sumologic.yaml.tmpl \
+    --name-template collection \
+    --set dryRun=true >> deploy/kubernetes/setup-sumologic.yaml.tmpl 2>/dev/null \
```
pmalek-sumo, line 204: not specifically related to this PR but how does `>>` work here when used in the middle of parameter list?

ci/build.sh
sumo-drosiek: we don't need them below too

ci/build.sh
sumo-drosiek: helm2 -> helm3

ci/build.sh
sumo-drosiek: same like on the end

pmalek-sumo (approved):

ci/build.sh
```diff
@@ -192,13 +191,14 @@ function generate_overrides() {
   echo "Generating setup job yaml from helm chart..."
   echo "# This file is auto-generated." > deploy/kubernetes/setup-sumologic.yaml.tmpl
 
-  with_files=$(find deploy/helm/sumologic/templates/setup/ -maxdepth 1 -iname "*.yaml" | sed 's#deploy/helm/sumologic/templates#-x templates#g' | sed 's/yaml/yaml \\/g')
+  # Get list of templates available for default values.yaml
+  with_files=$(find deploy/helm/sumologic/templates/setup -maxdepth 1 -iname "*.yaml" ! -path "*/setup-scc.yaml" | sed 's#deploy/helm/sumologic/templates#-s templates#g' | sed 's/yaml/yaml /g')
```
perk-sumo, line 195: `! -path "*/setup-scc.yaml"`?

perk-sumo (commented):
Overall 👍  but one thing seems fishy.

ci/build.sh
sumo-drosiek: do not take `*/setup-scc.yaml` during generating the `deploy/kubernetes/sumologic.yaml.tmpl`
It is disabled by default in the `values.yaml` so helm3 doesn't see it as template

ci/build.sh
```diff
@@ -14,6 +14,14 @@ err_report() {
 }
 trap 'err_report $LINENO' ERR
 
+function helm() {
+  docker run --rm \
+    -v "$(pwd):/chart" \
+    -w /chart \
+    sumologic/kubernetes-tools:master \
```
perk-sumo, line 21: This should not be `master` but rather `latest` or just fixed version.

perk-sumo (approved):
LGTM but let's make sure we are on fixed `tools` version when it's ready.

PR #1064, opened 2020-11-03
Create fields automatically in setup job

pmalek-sumo:
###### Description

Use https://github.com/SumoLogic/terraform-provider-sumologic/ (version `v2.3.0` introduced fields support) to create fields in setup job.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/conf/setup/setup.sh
+++ b/deploy/helm/sumologic/conf/setup/setup.sh
@@ -15,14 +15,14 @@ function remaining_fields() {
     echo "${RESPONSE}" | jq '.remaining'
 }
 
-# Check if we'd have at least 10 fields remaining after 8 additional fields
+# Check if we'd have at least 10 fields remaining after additional fields
 # would be created for the collection
 function should_create_fields() {
     local REMAINING=$(remaining_fields)
-    if [[ $(( REMAINING - 8 )) -ge 10 ]] ; then
-        echo 1
+    if [[ $(( REMAINING - {{ len .Values.sumologic.logs.fields }} )) -ge 10 ]] ; then
+        return 1
     else
-        echo 0
+        return 0
     fi
 }
 
@@ -34,12 +34,13 @@ COLLECTOR_NAME="{{- if .Values.sumologic.collectorName }}{{ .Values.sumologic.co
 terraform init
 
 # Sumo Logic fields
-readonly CREATE_FIELDS="$(should_create_fields)"
-if [[ "${CREATE_FIELDS}" -eq 0 ]]; then
+if should_create_fields ; then
+    readonly CREATE_FIELDS=0
     echo "Couldn't automatically create fields\n"
     echo "There's not enough fields which we could use for collection fields creation"
     echo "Please free some of them and rerun the setup job"
 else
+    readonly CREATE_FIELDS=1
     readonly FIELDS_RESPONSE="$(curl -XGET -s \
         -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
         "${SUMOLOGIC_BASE_URL}"v1/fields | jq '.data[]' )"
@@ -53,33 +54,27 @@ else
         fi
 
         terraform import \
-            -var="create_fields=${CREATE_FIELDS}" \
+            -var="create_fields=1" \
             sumologic_field."${FIELD}" "${FIELD_ID}"
     done
 fi
 
 # Sumo Logic Collector and HTTP sources
-terraform import \
-    -var="create_fields=${CREATE_FIELDS}" \
-    sumologic_collector.collector "$COLLECTOR_NAME"
+terraform import sumologic_collector.collector "$COLLECTOR_NAME"
 
 {{- $ctx := .Values -}}
 {{- range $type, $sources := .Values.sumologic.sources }}
 {{- if eq (include "terraform.sources.component_enabled" (dict "Context" $ctx "Type" $type)) "true" }}
 {{- range $key, $source := $sources }}
 {{- if eq (include "terraform.sources.to_create" (dict "Context" $ctx "Type" $type "Name" $key)) "true" }}
-terraform import \
-    -var="create_fields=${CREATE_FIELDS}" \
-    sumologic_http_source.{{ template "terraform.sources.name" (dict "Name" $key "Type" $type) }} "$COLLECTOR_NAME/{{ $source.name }}"
+terraform import sumologic_http_source.{{ template "terraform.sources.name" (dict "Name" $key "Type" $type) }} "$COLLECTOR_NAME/{{ $source.name }}"
 {{- end }}
 {{- end }}
 {{- end }}
 {{- end }}
 
 # Kubernetes Secret
-terraform import \
-    -var="create_fields=${CREATE_FIELDS}" \
-    kubernetes_secret.sumologic_collection_secret {{ .Release.Namespace }}/sumologic
+terraform import kubernetes_secret.sumologic_collection_secret {{ .Release.Namespace }}/sumologic
 
 # Apply planned changes
 terraform apply -auto-approve \
--- a/deploy/helm/sumologic/conf/setup/setup.sh
+++ b/deploy/helm/sumologic/conf/setup/setup.sh
@@ -44,7 +44,7 @@ else
         -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
         "${SUMOLOGIC_BASE_URL}"v1/fields | jq '.data[]' )"
 
-    declare -ra FIELDS=("cluster" "container" "deployment" "host" "namespace" "node" "pod" "service")
+    declare -ra FIELDS=({{ include "helm-toolkit.utils.joinListWithSpaces" .Values.sumologic.logs.fields }})
     for FIELD in "${FIELDS[@]}" ; do
         FIELD_ID=$( echo "${FIELDS_RESPONSE}" | jq -r "select(.fieldName == \"${FIELD}\") | .fieldId" )
         # Don't try to import non existing fields
--- a/deploy/helm/sumologic/conf/setup/setup.sh
+++ b/deploy/helm/sumologic/conf/setup/setup.sh
@@ -20,9 +20,9 @@ function remaining_fields() {
 function should_create_fields() {
     local REMAINING=$(remaining_fields)
     if [[ $(( REMAINING - {{ len .Values.sumologic.logs.fields }} )) -ge 10 ]] ; then
-        return 1
-    else
         return 0
+    else
+        return 1
     fi
 }
 
@@ -35,11 +35,6 @@ terraform init
 
 # Sumo Logic fields
 if should_create_fields ; then
-    readonly CREATE_FIELDS=0
-    echo "Couldn't automatically create fields"
-    echo "You do not have enough field capacity to create the required fields automatically."
-    echo "Please refer to https://help.sumologic.com/Manage/Fields to manually create the fields after you have removed unused fields to free up capacity."
-else
     readonly CREATE_FIELDS=1
     readonly FIELDS_RESPONSE="$(curl -XGET -s \
         -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
@@ -57,6 +52,11 @@ else
             -var="create_fields=1" \
             sumologic_field."${FIELD}" "${FIELD_ID}"
     done
+else
+    readonly CREATE_FIELDS=0
+    echo "Couldn't automatically create fields"
+    echo "You do not have enough field capacity to create the required fields automatically."
+    echo "Please refer to https://help.sumologic.com/Manage/Fields to manually create the fields after you have removed unused fields to free up capacity."
 fi
 
 # Sumo Logic Collector and HTTP sources
--- a/deploy/helm/sumologic/conf/setup/setup.sh
+++ b/deploy/helm/sumologic/conf/setup/setup.sh
@@ -1,14 +1,24 @@
 #!/bin/bash
-cp /etc/terraform/{locals,main,providers,resources,variables}.tf /terraform
-cd /terraform
 
 # Fix URL to remove "v1" or "v1/"
 export SUMOLOGIC_BASE_URL=${SUMOLOGIC_BASE_URL%v1*}
-
 # Support proxy for terraform
 export HTTP_PROXY=${HTTP_PROXY:=""}
 export HTTPS_PROXY=${HTTPS_PROXY:=""}
 export NO_PROXY=${NO_PROXY:=""}
+
+function remaining_fields() {
+    readonly RESPONSE="$(curl -XGET -s \
+        -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
+        "${SUMOLOGIC_BASE_URL}"v1/fields/quota)"
+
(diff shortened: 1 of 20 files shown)


deploy/helm/sumologic/conf/setup/main.tf
```diff
@@ -1,6 +1,5 @@
 terraform {
   required_providers {
     sumologic  = "~> 2.3.0"
-    kubernetes = "~> 1.11.3"
```
sumo-drosiek, line 4: why?

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,15 +1,23 @@
...
+function remaining_fields() {
+    readonly RESPONSE="$(curl -XGET -s \
+        -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
+        "${SUMOLOGIC_BASE_URL}"v1/fields/quota)"
+
+    echo "${RESPONSE}" | jq '.remaining'
+}
+
+cp /etc/terraform/{locals,main,providers,resources,variables}.tf /terraform
+cd /terraform
```
sumo-drosiek, line 19: what about `fields.tf`?

deploy/helm/sumologic/conf/setup/main.tf
pmalek-sumo: ouch the other way around, let me fix that

deploy/helm/sumologic/conf/setup/setup.sh
pmalek-sumo: Done, please check.
I'm not sure where `/etc/terraform` is mounted though so I don't know if copy from `/etc/terraform/fields/` is correct.

sumo-drosiek: `firlds.tf` it's not mounted as all as it is in subdirectory
`{{- (tpl (.Files.Glob "conf/setup/*").AsConfig .) | nindent 2 }}`
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/master/deploy/helm/sumologic/templates/setup/setup-configmap.yaml

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,15 +1,25 @@
...
+    readonly RESPONSE="$(curl -XGET -s \
+        -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
+        "${SUMOLOGIC_BASE_URL}"v1/fields/quota)"
+
+    echo "${RESPONSE}" | jq '.remaining'
+}
+
+cp /etc/terraform/{locals,main,providers,resources,variables}.tf /terraform/
+mkdir /terraform/fields/
+cp /etc/terraform/fields/{main,fields}.tf /terraform/fields/
```
sumo-drosiek, line 20: AFAIK `{main,fields}.tf` are in `/etc/terraform/`

sumo-drosiek, line 20:
```suggestion
cp /etc/terraform/{main_fields,fields}.tf /terraform/fields/
```

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -32,6 +42,19 @@ terraform import kubernetes_secret.sumologic_collection_secret {{ .Release.Names
...
 terraform apply -auto-approve
 
+# Fields
+#
+# Check if we have at least 10 fields remaining after we would
+# create 8 additional fields for the collection
+readonly REMAINING=$(remaining_fields)
+if [[ $(( REMAINING - 8 )) -ge -10 ]] ; then
+    echo "Creating fields..."
+    ( cd fields && terraform destroy -auto-approve )
```
sumo-drosiek, line 52:
```suggestion
    ( cd fields && terraform apply -auto-approve )
```
?

tests/terraform/static/collector_fields.output.yaml
```diff
@@ -163,46 +227,116 @@ data:
...
+        -var="create_fields=${CREATE_FIELDS}"
   
     # Cleanup env variables
     export SUMOLOGIC_BASE_URL=
     export SUMOLOGIC_ACCESSKEY=
     export SUMOLOGIC_ACCESSID=
   
     bash /etc/terraform/custom.sh
-  variables.tf: |-
+  variables.tf: |
```
pmalek-sumo, line 339: As far as I understand this I needed to change this because I've added a new line at the end of `variables.tf` file.

Since all other section (apart from `providers.tf` which I guess is a mistake?) have a new line at EOF I'm changing this as well to use `  variables.tf: |` instead of `  variables.tf: |-`

cc: @sumo-drosiek @perk-sumo

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,36 +1,89 @@
...
+if [[ "${CREATE_FIELDS}" -eq 0 ]]; then
+    echo "Couldn't automatically create fields\n"
+    echo "There's not enough fields which we could use for collection fields creation"
+    echo "Please free some of them and rerun the setup job"
+else
+    readonly FIELDS_RESPONSE="$(curl -XGET -s \
+        -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
+        "${SUMOLOGIC_BASE_URL}"v1/fields | jq '.data[]' )"
+
+    declare -ra FIELDS=("cluster" "container" "deployment" "host" "namespace" "node" "pod" "service")
```
sumo-drosiek, line 47: Could we move it to the `values.yaml`?

tests/terraform/static/collector_fields.output.yaml
sumo-drosiek: Thats correct, thats fine

deploy/helm/sumologic/conf/setup/setup.sh
pmalek-sumo: Yes we could. We could also rewrite `fields.tf` to be a template looping over this array (declared in `values.yaml`)

sumo-drosiek: if we move fields to `values.yaml` I don't see any solution other than looping over it in `fields.tf`

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,36 +1,89 @@
...
+        "${SUMOLOGIC_BASE_URL}"v1/fields/quota)"
+
+    echo "${RESPONSE}" | jq '.remaining'
+}
+
+# Check if we'd have at least 10 fields remaining after 8 additional fields
+# would be created for the collection
+function should_create_fields() {
+    local REMAINING=$(remaining_fields)
+    if [[ $(( REMAINING - 8 )) -ge 10 ]] ; then
```
sumo-drosiek, line 22: I think we have to update those numbers based on the `values.yaml`

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,36 +1,89 @@
...
+cd /terraform
+
 COLLECTOR_NAME="{{- if .Values.sumologic.collectorName }}{{ .Values.sumologic.collectorName }}{{- else}}{{ .Values.sumologic.clusterName }}{{- end}}"
 
 terraform init
 
-# Sumo Collector and HTTP sources
-terraform import sumologic_collector.collector "$COLLECTOR_NAME"
+# Sumo Logic fields
+readonly CREATE_FIELDS="$(should_create_fields)"
```
sumo-drosiek, line 37: I think better would be using `$?` and `exit`

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,36 +1,89 @@
...
+if [[ "${CREATE_FIELDS}" -eq 0 ]]; then
```
sumo-drosiek, line 38: I think better would be using `$?` and `exit`

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,36 +1,89 @@
...
+    declare -ra FIELDS=({{ include "helm-toolkit.utils.joinListWithSpaces" .Values.sumologic.logs.fields }})
+    for FIELD in "${FIELDS[@]}" ; do
+        FIELD_ID=$( echo "${FIELDS_RESPONSE}" | jq -r "select(.fieldName == \"${FIELD}\") | .fieldId" )
+        # Don't try to import non existing fields
+        if [[ -z "${FIELD_ID}" ]]; then
+            continue
+        fi
+
+        terraform import \
+            -var="create_fields=${CREATE_FIELDS}" \
```
sumo-drosiek, line 56:
```suggestion
            -var="create_fields=1" \
```

deploy/helm/sumologic/conf/setup/variables.tf
```diff
@@ -15,4 +15,9 @@ variable "collector_name" {
...
   type  = string
   default = "{{ .Release.Namespace }}"
-}
\ No newline at end of file
+}
+
+variable "create_fields" {
+  description = "If set, terraform will attempt to create fields at Sumo Logic"
+  type = bool
+}
```
sumo-drosiek, line 23: could we set default to `false/0` and do not change other imports?

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -778,3 +778,15 @@ Example:
...
+{{/*
+Generate a space separated list of quoted values:
+
+Example:
+
+{{ include "helm-toolkit.utils.joinListWithSpaces" .Values.sumologic.logs.fields }}
+*/}}
+{{- define "helm-toolkit.utils.joinListWithSpaces" -}}
+{{- $local := dict "first" true -}}
+{{- range $k, $v := . -}}{{- if not $local.first }} {{ end -}}{{ $v | quote }}{{- $_ := set $local "first" false -}}{{- end -}}
```
sumo-drosiek, line 791: could you introduce string variable and split this concatenation on multiple lines?

tests/terraform/static/all_fields.output.yaml
```diff
@@ -49,6 +49,63 @@ data:
...
+      data_type = "String"
+      state = "Enabled"
+    }
+    resource "sumologic_field" "service" {
+      count = var.create_fields ? 1 : 0
+  
+      field_name = "service"
+      data_type = "String"
+      state = "Enabled"
+    }
```
sumo-drosiek, line 108: I would rather set `fields` to empty list in `input` file in current tests than copying this into all of them

sumo-drosiek (approved):
🥁

deploy/helm/sumologic/templates/NOTES.txt
```diff
@@ -11,4 +11,8 @@ Check the release status by running:
 
 {{- if eq .Values.fluentd.persistence.enabled  false }}
 WARNING: File persistence for fluentd is disabled. This might lead to loss of data in case of memory buffer overflow. We recommend turning this property on for production environments by setting fluentd.persistence.enabled=true
-{{- end }}
\ No newline at end of file
+{{- end }}
+
+We've tried to automatically create fields. In an unlikely scenario that this
+fails please refer to the following to create them manually:
+https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/2b3ca63/deploy/docs/Installation_with_Helm.md#prerequisite
```
frankreno, line 18: There is no easy way to only show the text in the event of an actual failure I assume?

deploy/kubernetes/setup-sumologic.yaml.tmpl
```diff
@@ -247,6 +351,12 @@ data:
       type  = string
       default = "$NAMESPACE"
     }
+  
+    variable "create_fields" {
+      description = "If set, terraform will attempt to create fields at Sumo Logic"
+      type        = bool
+      default     = false
```
frankreno, line 358: Why default to false?

frankreno (approved):
LGTM, but left a few comments

deploy/helm/sumologic/templates/NOTES.txt
pmalek-sumo: That was discussed and we couldn't figure out a way of getting return feedback from the setup job to helm to conditionally render this message.

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,36 +1,89 @@
...
 terraform init
 
-# Sumo Collector and HTTP sources
-terraform import sumologic_collector.collector "$COLLECTOR_NAME"
+# Sumo Logic fields
+readonly CREATE_FIELDS="$(should_create_fields)"
+if [[ "${CREATE_FIELDS}" -eq 0 ]]; then
+    echo "Couldn't automatically create fields\n"
+    echo "There's not enough fields which we could use for collection fields creation"
+    echo "Please free some of them and rerun the setup job"
```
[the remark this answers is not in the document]
pmalek-sumo: Done

deploy/kubernetes/setup-sumologic.yaml.tmpl
pmalek-sumo: No specific reason. I've provided the default as suggested by @sumo-drosiek prevent the need for specifying terraform variable during imports.

We can also default to `true` if that's more appropriate for our use case. It will be specified on `terraform apply` anyway

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/bc18a3e2bfe0285484a777a561a38946325b8e34/deploy/helm/sumologic/conf/setup/setup.sh#L79-L81

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,21 +1,67 @@
...
+cp /etc/terraform/{locals,main,providers,resources,variables,fields}.tf /terraform/
+cd /terraform
+
 COLLECTOR_NAME="{{- if .Values.sumologic.collectorName }}{{ .Values.sumologic.collectorName }}{{- else}}{{ .Values.sumologic.clusterName }}{{- end}}"
 
 terraform init
 
-# Sumo Collector and HTTP sources
+# Sumo Logic fields
+if should_create_fields ; then
```
perk-sumo, line 37: The function name is misleading because now the flow is as follow:
if `should_create_fields` then don't create them
otherwise please do

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -1,21 +1,67 @@
...
+
+# Check if we'd have at least 10 fields remaining after additional fields
+# would be created for the collection
+function should_create_fields() {
+    local REMAINING=$(remaining_fields)
+    if [[ $(( REMAINING - {{ len .Values.sumologic.logs.fields }} )) -ge 10 ]] ; then
+        return 1
+    else
+        return 0
+    fi
```
perk-sumo, line 26: Given the function name those should be the other way round I suppose?
```suggestion
    if [[ $(( REMAINING - {{ len .Values.sumologic.logs.fields }} )) -ge 10 ]] ; then
        return 0
    else
        return 1
    fi
```

perk-sumo (commented):
One nit around function name and the result it produces.

deploy/kubernetes/setup-sumologic.yaml.tmpl
pmalek-sumo: After discussion with folks on PL sync I've changed this to `true` (which doesn't really change anything in our code but expresses our intent of creating fields by default)

perk-sumo (commented):
👍

PR #1071, opened 2020-11-06
Update autoscaling/v1 to autoscaling/v2beta2

pmalek-sumo:
###### Description

Update autoscaling/v1 to autoscaling/v2beta2 in order to enable custom metrics autoscaling and autoscaling based on memory utilization.

This PR adds `targetMemoryUtilizationPercentage` key to enable autoscaling based on memory utilization.

ref: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-multiple-metrics-and-custom-metrics

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -322,7 +322,13 @@ fluentd:
       enabled: false
       minReplicas: 3
       maxReplicas: 10
-      targetCPUUtilizationPercentage: 50
+      metrics:
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 50
 
     ## Option to specify PodDisrutionBudgets
     ## You can specify only one of maxUnavailable and minAvailable in a single PodDisruptionBudget
@@ -558,7 +564,13 @@ fluentd:
       enabled: false
       minReplicas: 3
       maxReplicas: 10
-      targetCPUUtilizationPercentage: 50
+      metrics:
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 50
 
     ## Option to specify PodDisrutionBudgets
     ## You can specify only one of maxUnavailable and minAvailable in a single PodDisruptionBudget
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -322,13 +322,8 @@ fluentd:
       enabled: false
       minReplicas: 3
       maxReplicas: 10
-      metrics:
-      - type: Resource
-        resource:
-          name: cpu
-          target:
-            type: Utilization
-            averageUtilization: 50
+      targetCPUUtilizationPercentage: 50
+      # targetMemoryUtilizationPercentage: 50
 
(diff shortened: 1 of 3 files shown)


sumo-drosiek:
@frankreno @perk-sumo 
Do we want to support more factors than CPU? IMHO, we should avoid to copy k8s templates to values.yaml if it's possible

frankreno:
I agree with @sumo-drosiek.  I feel like we should be able to handle this w/o a breaking change as well. In the values.yaml we can have the same key we do today (`targetCPUUtilizationPercentage`).  We can modify the HPA with new format in the template as opposed to the values.  If customers want to add more metrucs, we can add an `extraHPAMetrics` property that the template could render no?

pmalek-sumo:
> I agree with @sumo-drosiek. I feel like we should be able to handle this w/o a breaking change as well. In the values.yaml we can have the same key we do today (`targetCPUUtilizationPercentage`). We can modify the HPA with new format in the template as opposed to the values. If customers want to add more metrucs, we can add an `extraHPAMetrics` property that the template could render no?

@frankreno 

The reason I'm hesitant on this approach is that

* we would be "overriding" the "scope" (?) of `targetCPUUtilizationPercentage` (`v1` vs `v2beta`) - this in itself wouldn't be that big of a deal
* we already had a thread on slack about a customer who'd like to configure hpa based on memory, this will require either `extraHPAMetrics` as suggested or something like `targetMemoryUtilizationPercentage` which doesn't exist neither in `v1` nor  in `v2beta2` which might confuse users 

Not my call, just chippin' in :)

perk-sumo:
There is one problem with doing a breaking change and fully migrating - it's still a `beta` version.
I'd rather have some "scope" inconsistence than changing our configuration back and forth in case the hpa api changes.

pmalek-sumo:
> There is one problem with doing a breaking change and fully migrating - it's still a `beta` version.
> I'd rather have some "scope" inconsistence than changing our configuration back and forth in case the hpa api changes.

Ok, sure. Let's do that then.
One thing though: do we set the default and if so then what value do we want for `targetMemoryUtilizationPercentage`?

perk-sumo:
Let's set it at 50% but not use it by default.

deploy/helm/sumologic/values.yaml
```diff
@@ -322,7 +322,13 @@ fluentd:
       enabled: false
       minReplicas: 3
       maxReplicas: 10
-      targetCPUUtilizationPercentage: 50
```
perk-sumo, line 325: Let's not change this key just yet (as long as autoscaling `v2` is in beta)

perk-sumo (changes requested):
Let's not change the `values.yaml` keys around autoscaling just yet.

pmalek-sumo:
Removing `breaking-change` label as we decided to use `targetMemoryUtilizationPercentage` key for autoscaling based on memory utilization

pmalek-sumo:
@perk-sumo Please take a look one more time

sumo-drosiek (approved):

perk-sumo (approved):
👍

PR #1133, opened 2020-11-19
Switch back to docker_mode

frankreno:
When migrating to the new fluent-bit chart, I removed docker_mode support.  This adds it back in.


sumo-drosiek (approved):
Thanks :)

PR #1137, opened 2020-11-19
Only import sources when collector exists

pmalek-sumo:
###### Description

Only import sources when collector exists.
This should cut down the time needed for setup job to complete.

###### Testing performed

- [ ] ci/build.sh


perk-sumo (approved):
LGTM

PR #1181, opened 2020-11-26
Use Github Actions for CI

pmalek-sumo:
###### Description

Start using Github Actions for CI.
For now it works for PRs but further work is required to handle github environment variables for pushing docker images purposes.

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,8 +24,9 @@ jobs:
       - run: gem install bundler
       - name: yq
         run: |
-            curl --retry 10 --retry-max-time 120 --retry-delay 5 -L -o /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
-            chmod +x /usr/local/bin/yq
+            curl --retry 10 --retry-max-time 120 --retry-delay 5 -L -o /tmp/yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
+            chmod +x /tmp/yq
+            sudo mv /tmp/yq /usr/local/bin/yq
       - name: shellcheck
         run: |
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -0,0 +1,2 @@
+shellcheck:
+	./ci/shellcheck.sh
```
sumo-drosiek, line 2: add to vagrant as well

sumo-drosiek: also add it to `test` target or create new one to perform all required tests

.github/workflows/main.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          ruby-version: '2.6'
+      - run: gem install bundler
+      - name: yq
+        run: |
+            curl --retry 10 --retry-max-time 120 --retry-delay 5 -L -o /tmp/yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
+            chmod +x /tmp/yq
+            sudo mv /tmp/yq /usr/local/bin/yq
+      - name: tests
+        run: bash ci/tests.sh
```
sumo-drosiek, line 51: we could use

```suggestion
        run: make test
```

maybe 🤔

ci/build.sh
```diff
@@ -72,8 +72,8 @@ function set_up_github() {
   local token="${1}"
   local branch="${2}"
 
-  git config --global user.email "travis@travis-ci.org"
-  git config --global user.name "Travis CI"
+  git config --global user.email "bot@github.com"
+  git config --global user.name "Github Action"
```
sumo-drosiek, line 76: shouldn't we use platform independent values?
I would skip email and set `user.name` to `Continuous Integration`

pmalek-sumo: Done, it doesn't work now anyway because of not handling env variables that contain branch name etc. in GH env

sumo-drosiek (approved):

perk-sumo (approved):
👍

PR #1186, opened 2020-11-27
Add Github Actions build and push actions

pmalek-sumo:
###### Description

This introduces pushing helm chart and docker image for `main` branch, `release` branches and for tags.
There are 3 separate workflows:

- `PRs checks` - checks for PRs for `main` branch
- `Dev builds` - for `main` and release branches; building docker images, pushing it to docker hub and pushing helm chart to dev repo
- `Release builds` - for release tags; building docker images, pushing it to docker hub and pushing helm chart to main primary repo

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -1,166 +1,27 @@
 #!/usr/bin/env bash
 
-# shellcheck disable=SC2154
-VERSION="${TRAVIS_TAG:-0.0.0}"
-# shellcheck disable=SC2154
-VERSION="${VERSION#v}"
-: "${DOCKER_TAG:=sumologic/kubernetes-fluentd}"
-: "${DOCKER_USERNAME:=sumodocker}"
+readonly ROOT_DIR="$(dirname "$(dirname "${0}")")"
+# shellcheck disable=SC1090
+source "${ROOT_DIR}"/ci/_build_functions.sh
 
-echo "Starting build process in: $(pwd) with version tag: ${VERSION}"
-err_report() {
-    echo "Script error on line $1"
-    exit 1
-}
-trap 'err_report $LINENO' ERR
-
-function helm() {
-  docker run --rm \
-    -v "$(pwd):/chart" \
-    -w /chart \
-    sumologic/kubernetes-tools:2.0.0 \
-    helm "$@"
-}
-
-function get_branch_to_checkout() {
-  [[ "${TRAVIS_EVENT_TYPE}" == pull_request ]] \
-    && echo "${TRAVIS_PULL_REQUEST_BRANCH}" \
-    || echo "${TRAVIS_BRANCH}"
-}
-
-function bundle_fluentd_plugin() {
-  local plugin_name="${1}"
-  local version="${2}"
-  # Strip everything after "-" (longest match) to avoid gem prerelease behavior
-  local gem_version="${version%%-*}"
-
-  if [[ -z "${version}" ]] ; then
-    echo "Please provide the version when bundling fluentd plugins"
-    exit 1
-  fi
-  pushd "${plugin_name}" || exit 1
-
-  echo "Building gem ${plugin_name} version ${gem_version} in $(pwd) ..."
-  sed -i.bak "s/0.0.0/${gem_version}/g" ./"${plugin_name}".gemspec
-  rm -f ./"${plugin_name}".gemspec.bak
-
-  echo "Install bundler..."
-  bundle install
-
-  echo "Build gem ${plugin_name} ${gem_version}..."
-  gem build "${plugin_name}"
-  mv ./*.gem ../deploy/docker/gems
-  popd || exit 1
-}
-
-function bundle_fluentd_plugins() {
-  local version="${1}"
-
-  if [[ -z "${version}" ]] ; then
-    echo "Please provide the version when bundling fluentd plugins"
-    exit 1
-  fi
-
-  find . -maxdepth 1 -name 'fluent-plugin-*' -type 'd' -print |
-    while read -r line; do
-      # Run tests in their own context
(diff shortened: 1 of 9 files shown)


ci/build.sh
```diff
@@ -1,166 +1,24 @@
...
-  rm main.tf
-  rm -f ./gems/*.gem
-  popd || exit 1
+if is_checkout_on_tag; then
+  VERSION="$(git describe --tags --always)"
+  readonly VERSION="${VERSION#v}"
+else
+  # Don't use non-tag 'git describe' output as this will
+  # fail when building fluentd plugins.
+  readonly VERSION="0.0.0"
```
sumo-drosiek, line 14: Didn't we tested that the plugins are build correctly?

ci/build.sh
```diff
@@ -1,166 +1,24 @@
...
-  echo "Test docker image locally..."
-  ruby deploy/test/test_docker.rb || exit 1
-}
+: "${DOCKER_TAG:=sumologic/kubernetes-fluentd}"
 
-# Set up Github
-if [ -n "${GITHUB_TOKEN}" ]; then
-  set_up_github "${GITHUB_TOKEN}" "$(get_branch_to_checkout)"
-fi
+echo "Starting build in: $(pwd) with version tag: ${VERSION}"
```
sumo-drosiek, line 19: Please change `pwd` to the repo main directory. functions won't work otherwise

ci/push.sh
```diff
@@ -0,0 +1,30 @@
...
+
+: "${DOCKER_TAG:=sumologic/kubernetes-fluentd}"
+: "${DOCKER_USERNAME:=sumodocker}"
+
+if [[ -z "${DOCKER_PASSWORD}" ]]; then
+  echo "Provide DOCKER_PASSWORD in order to push built image to container registry"
+  exit 1
+fi
+
+echo "Starting push in: $(pwd) with version tag: ${VERSION}"
```
sumo-drosiek, line 19: Please change pwd to the repo main directory. functions won't work otherwise

ci/build.sh
pmalek-sumo: Argh, I've removed the old build from GH actions but `git describe ...` returned short SHA which `gem bundle ...` complained about.

ci/build.sh
pmalek-sumo: i.e. `cd` to repo root dir or change the `echo`?

sumo-drosiek: `cd`/`pushd` to repo root. `echo` is fine

sumo-drosiek: we are using in functions relative paths, so if you run the script outside of the repo root it's going to fail

ci/build.sh
sumo-drosiek: so git describe doesn;t return `1.3.2-1234567`?

ci/build.sh
pmalek-sumo: Added `pushd "$(dirname "${BASH_SOURCE[0]}")/.." || exit 1`

ci/push.sh
pmalek-sumo: Added `pushd "$(dirname "${BASH_SOURCE[0]}")/.." || exit 1`

ci/build.sh
pmalek-sumo: Ah, ok so it didn't but I've added the `git fetch --tags --prune --unshallow` and then I've figured out that we don't really need to fetch full history (via `git fetch --unshallow ...`) but it's enough to fetch current branch (and make sure we have `unshallow`ed repo history

This is handled in `fetch_current_branch` in here: 

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/5735894346043a74c008733dcb74da34566e6b44/ci/_build_functions.sh#L126-L143

pmalek-sumo: And besides bandwidth we also save time as `git fetch --tags --unshallow` takes roughly 10-20s https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/1464248561?check_suite_focus=true#step:5:1
while fetching only specified branch it takes about a second https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/1470602457?check_suite_focus=true#step:6:5 (till line 72)

ci/build.sh
```diff
@@ -1,166 +1,27 @@
...
+pushd "${ROOT_DIR}" || exit 1
```
sumo-drosiek, line 20:
```suggestion
pushd "${ROOT_DIR}" || exit 1
echo "Starting build in: $(pwd) with version tag: ${VERSION}"
```

ci/build.sh
sumo-drosiek: So we can simplify the if else statement now

ci/tests.sh
```diff
@@ -2,57 +2,22 @@
...
+source "${ROOT_DIR}"/ci/_build_functions.sh
+
+if is_checkout_on_tag; then
+  VERSION="$(git describe --tags --always)"
+  readonly VERSION="${VERSION#v}"
+else
+  # Don't use non-tag 'git describe' output as this will
+  # fail when building fluentd plugins.
+  readonly VERSION="0.0.0"
+fi
```
sumo-drosiek, line 18: Can we simplify this as well?

pmalek-sumo: Done. For the reference: changing this to assign what `git describe --tags` returns will yield in semver versions (e.g. `2.0.0`) for versions that are not a `2.0.0` tag, so `2.0.0-dev.0-136-g205f297` will yields gems with version `2.0.0`

sumo-drosiek (approved):
LGTM

PR #1183, opened 2020-11-26
Bundle kubernetes_metadata plugin with fix for Fluentd pod restarts

astencel-sumo:
###### Description

This PR bundles the `kubernetes_metadata` Fluentd plugin from https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/ into our repo, as the upstream is not active and our [pull request there](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/267) is not being merged despite being accepted.

There are two commits in this PR: the first one bundles the plugin's sources from `master` (at commit https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/commit/84f66a8f9e06ab5b5211053fcce4cd8ab4bd74ba from Oct 8), the second one incorporates changes from the pull request https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/267.

###### Testing performed

- [x] ci/build.sh
- [x] Redeploy fluentd and fluentd-events pods
- [x] Confirm events, logs, and metrics are coming in

PR #1193, opened 2020-11-30
Backport Bundle kubernetes_metadata plugin with fix for Fluentd pod restarts

astencel-sumo:
Original PR on main: #1183.


kkujawa-sumo (approved):

astencel-sumo:
This should be merged only after 1.3.3 is released.

PR #1232, opened 2020-12-11
Replace Docker Hub with ECR

astencel-sumo:


pmalek-sumo (approved):

deploy/helm/sumologic/values.yaml
```diff
@@ -1873,7 +1873,7 @@ otelagent:
     ## Memory Ballast size should be max 1/3 to 1/2 of memory.
     memBallastSizeMib: "250"
     image:
-      repository: "sumologic/opentelemetry-collector"
+      repository: "public.ecr.aws/u5z5f8z6/opentelemetry-collector"
```
pmalek-sumo, line 1876: BTW: in yaml, when you don't use special characters you don't need `"` around strings

PR #1243, opened 2020-12-15
Migration script v2 set remoteWrite remoteTimeout to 5s

pmalek-sumo:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/upgrade-2.0.0.sh
+++ b/deploy/helm/sumologic/upgrade-2.0.0.sh
@@ -386,6 +386,18 @@ function add_new_scrape_labels_to_prometheus_service_monitors(){
     'prometheus-operator.prometheus.additionalServiceMonitors.[*].selector.matchLabels."sumologic.com/scrape"' true
 }
 
+function kube_prometheus_stack_set_remote_write_timeout_to_5s() {
+  if [[ -z "$(yq r "${TEMP_FILE}" -- 'kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite')" ]]; then
+    return
+  fi
+
+  if [[ -n "$(yq r "${TEMP_FILE}" -- 'kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite.[*].remoteTimeout')" ]]; then
+    echo
+    info "kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite.[*].remoteTimeout is set"
+    info "Please note that we've set it by default to 5s in 2.0.0"
+  fi
+}
+
 function migrate_sumologic_sources() {
   # Nothing to migrate, return
   if [[ -z $(yq r "${TEMP_FILE}" sumologic.sources) ]] ; then
@@ -552,6 +564,7 @@ migrate_customer_keys
 migrate_prometheus_recording_rules
 add_new_scrape_labels_to_prometheus_service_monitors
 migrate_prometheus_operator_to_kube_prometheus_stack
+kube_prometheus_stack_set_remote_write_timeout_to_5s
 
 migrate_sumologic_sources
 migrate_sumologic_setup_fields
--- a/tests/upgrade_v2_script/static/prometheus_remote_write_timeout.input.yaml
+++ b/tests/upgrade_v2_script/static/prometheus_remote_write_timeout.input.yaml
@@ -1,0 +1,117 @@
+prometheus-operator:
+  prometheus:
+    prometheusSpec:
+      scrapeInterval: "30s"
+      resources:
+        limits:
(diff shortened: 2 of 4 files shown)


sumo-drosiek (commented):
It would be nice to check if value is already set, but LGTM

pmalek-sumo:
> It would be nice to check if value is already set, but LGTM

So I've tried a different approach with merging the temporary file that we work on with `remoteWrite` document (with specified `5s` `remoteTimeout`s) but the problem with this is that it might overwrite other values in that tree.

Maybe there's a way of doing this in `yq` but I'm not familiar with it. I'm also not sure if we can iterate over elements with `yq` and then take action in the script.

perk-sumo (commented):
I think we shouldn't change this value at all.
Just check the user config and in case it is set we should display the INFO that we've changed the default value to 5.

pmalek-sumo:
> I think we shouldn't change this value at all.
> Just check the user config and in case it is set we should display the INFO that we've changed the default value to 5.

Changed to just log if `remoteWrite` is set

sumo-drosiek:
> Maybe there's a way of doing this in `yq` but I'm not familiar with it. I'm also not sure if we can iterate over elements with `yq` and then take action in the script.

It's possible in `yq`. We did it in upgrade-script for v1

deploy/helm/sumologic/upgrade-2.0.0.sh
```diff
@@ -386,6 +386,18 @@ function add_new_scrape_labels_to_prometheus_service_monitors(){
...
+function kube_prometheus_stack_set_remote_write_timeout_to_5s() {
+  if [[ -z "$(yq r "${TEMP_FILE}" -- 'kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite')" ]]; then
+    return
+  fi
+
+  if [[ -n "$(yq r "${TEMP_FILE}" -- 'kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite.[*].remoteTimeout')" ]]; then
+    echo
+    info "kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite.[*].remoteTimeout is set"
+    info "Please note that we've set it by default to 5s in 2.0.0"
+  fi
```
sumo-drosiek, line 398: We should inform if `remoteTImeout` is not set as well

tests/upgrade_v2_script/static/prometheus_remote_write_timeout.input.yaml
```diff
@@ -0,0 +1,117 @@
...
+              sourceLabels: [__name__]
+
+        - url: http://$(CHART).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.applications.jmx
+          writeRelabelConfigs:
+            - action: keep
+              regex: (?:java_lang_(ClassLoading_(TotalL|Unl|L)oadedClassCount|Compilation_TotalCompilationTime|GarbageCollector_(Collection(Count|Time)|LastGcInfo_(GcThreadCount|duration|(memoryU|u)sage(After|Before)Gc_.*_used))|MemoryPool_(CollectionUsage(ThresholdSupported|_committed|_max|_used)|(Peak|)Usage_(committed|max|used)|UsageThresholdSupported)|Memory_((Non|)HeapMemoryUsage_(committed|max|used)|ObjectPendingFinalizationCount)|OperatingSystem_(AvailableProcessors|(CommittedVirtual|(Free|Total)(Physical|))MemorySize|(Free|Total)SwapSpaceSize|(Max|Open)FileDescriptorCount|ProcessCpu(Load|Time)|System(CpuLoad|LoadAverage))|Runtime_(BootClassPathSupported|Pid|Uptime|StartTime)|Threading_(CurrentThread(AllocatedBytes|(Cpu|User)Time)|(Daemon|Peak|TotalStarted|)ThreadCount|(ObjectMonitor|Synchronizer)UsageSupported|Thread(AllocatedMemory.*|ContentionMonitoring.*|CpuTime.*))))
+              sourceLabels: [__name__]
```
sumo-drosiek, line 117: newline at EOF

deploy/helm/sumologic/upgrade-2.0.0.sh
sumo-drosiek: cc @perk-sumo

sumo-drosiek: or set it

perk-sumo: It is set by default in values.yaml, no need to add this into migrated conf file.

perk-sumo (approved):
👍

PR #1259, opened 2020-12-21
Upgrade OTC to 0.16.0-sumo

mat-rumian:
###### Description

###### Testing performed

- [ ] ci/build.sh

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -1983,7 +1983,7 @@ otelcol:
     memBallastSizeMib: "683"
     image:
       repository: "public.ecr.aws/sumologic/opentelemetry-collector"
-      tag: "0.12.0"
+      tag: "0.16.0-sumo"
       pullPolicy: IfNotPresent
 
     ## Extra Environment Values - allows yaml definitions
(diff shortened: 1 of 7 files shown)


pmalek-sumo:
@mat-rumian Please fix tests :)

mat-rumian:
@pmm-sumo please merge when you decide it's a proper time for that :)

deploy/helm/sumologic/values.yaml
```diff
@@ -2102,15 +2102,36 @@ otelcol:
...
-        ## Number of workers that dequeue batches
-        num_workers: 16
-        ## Maximum number of batches kept in memory before data is dropped
-        queue_size: 10000
-        ## Whether to retry on failure or give up and drop
-        retry_on_failure: true
+      ## Smart cascading filtering rules with preset limits.
+      cascading_filter:
+        ## (default = 30s): Wait time since the first span of a trace before making
+        ## a filtering decision
```
pmalek-sumo, line 2108: Not sure if we want those comments about the defaults.

If we really want it then I'd put it at the end of the comment like so (and small suggestion regarding wording used):

```suggestion
        ## Wait time between obtaining first span of a trace and making a filtering decision
        ## Default: 30s
```

pmm-sumo:
@perk-sumo customer are asking for this capability, can we include it in 2.0.0?

pmm-sumo (approved):

deploy/helm/sumologic/values.yaml
sumo-drosiek: I would skip the defaults too
@mat-rumian

pmm-sumo: @sumo-drosiek I am going to update this section a little bit

pmm-sumo: I gave it a though and I think this is complex enough and almost always requires assigning values by user to better have too much rather too little description

pmm-sumo:
I have tested it and it now looks good

One caveat is that the default gRPC port has changed from `55680` to `4317`, though in the migration period OTC listens on both of them

perk-sumo (approved):

PR #1261, opened 2020-12-21
Prevent access to uninitialized variable

astencel-sumo:
The `@pods_to_services` variable is being initialized in another thread: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/035dc054d0811d13fb14aafcb85e048f3191d6fc/fluent-plugin-enhance-k8s-metadata/lib/sumologic/kubernetes/service_monitor.rb#L52 We need to make sure `@pods_to_services` is not accessed until it's properly initialized.

### DIFF
--- a/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
+++ b/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
@@ -83,7 +83,7 @@ module Fluent
             record.delete('service')
           end
           metadata = get_pod_metadata(namespace_name, pod_name)
-          service = @pods_to_services[pod_name]
+          service = @pods_to_services[pod_name] unless @pods_to_services.nil?
           metadata['service'] = {'service' => service.sort!.join('_')} if !(service.nil? || service.empty?)
 
           if @data_type == 'metrics' && (record['node'].nil? || record['node'] == "")
--- a/fluent-plugin-enhance-k8s-metadata/test/plugin/test_filter_enhance_k8s_metadata.rb
+++ b/fluent-plugin-enhance-k8s-metadata/test/plugin/test_filter_enhance_k8s_metadata.rb
@@ -67,7 +67,6 @@ class EnhanceK8sMetadataFilterTest < Test::Unit::TestCase
 
   def create_driver(conf)
     driver = Fluent::Test::Driver::Filter.new(Fluent::Plugin::EnhanceK8sMetadataFilter).configure(conf).instance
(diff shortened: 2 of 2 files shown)


fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
```diff
@@ -83,7 +83,7 @@ def decorate_record(record)
             record.delete('service')
           end
           metadata = get_pod_metadata(namespace_name, pod_name)
-          service = @pods_to_services[pod_name]
+          service = @pods_to_services[pod_name] unless @pods_to_services.nil?
```
sumo-drosiek, line 86: does it mean we are going to not tag service properly until some background job is done?

astencel-sumo: Yes I believe this is how it works (and used to work before as well, only with an error which is now prevented).

pmalek-sumo: Do other lines (below this one) make sense when `service` is not assigned here (it would be nil then, right?)

astencel-sumo: The next line checks if `service` is `nil`, so it seems OK to me:
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/1b6ce8eed201269b7a29f92b2218bbc2bcba3817/fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb#L87

sumo-drosiek: > Yes I believe this is how it works (and used to work before as well, only with an error which is now prevented).

Not sure, error prevented from ingesting data if `@pods_to_services` was not initialised, now we do not wait for initialisation and part of the records won't get metadata even if there is service for pod :confused:

sumo-drosiek (approved):

fluent-plugin-enhance-k8s-metadata/lib/fluent/plugin/filter_enhance_k8s_metadata.rb
astencel-sumo: We did some investigation and it occurs that when a Fluentd plugin crashes on a record, Fluentd does not drop o retry the record which caused the crash, it just happily passes the record (possibly half-modified by the plugin) further into the processing pipeline.

PR #1301, opened 2021-01-07
Migrate prometheus pre 1.14 rules only when prometheus-operator.kubeTargetVersionOverride is set to 1.13.0-0

pmalek-sumo:
###### Description

Migrate only when `prometheus-operator.kubeTargetVersionOverride` is set to `1.13.0-0`.

---

###### Testing performed


deploy/helm/sumologic/upgrade-2.0.0.sh
```diff
@@ -181,17 +181,14 @@ function migrate_prometheus_recording_rules() {
...
   local RECORDING_RULES_OVERRIDE
-  RECORDING_RULES_OVERRIDE=$(yq r "${TEMP_FILE}" -- 'prometheus-operator.kubeTargetVersionOverride')
+  readonly RECORDING_RULES_OVERRIDE=$(yq r "${TEMP_FILE}" -- 'prometheus-operator.kubeTargetVersionOverride')
 
   if [[ "${RECORDING_RULES_OVERRIDE}" == "1.13.0-0" ]]; then
+    add_prometheus_pre_1_14_recording_rules "${TEMP_FILE}"
     info "Removing prometheus kubeTargetVersionOverride='1.13.0-0'"
     yq d -i "${TEMP_FILE}" "prometheus-operator.kubeTargetVersionOverride"
-    add_prometheus_pre_1_14_recording_rules "${TEMP_FILE}"
-  elif [[ -z "${RECORDING_RULES_OVERRIDE}" ]]; then
```
sumo-drosiek, line 190: what about this condition? What if `kubeTargetVersionOverride` is unset?

sumo-drosiek:
Also do we have test for cover that?

deploy/helm/sumologic/upgrade-2.0.0.sh
pmalek-sumo: We don't want to add pre 1.14 rules then.
As per discussion with @frankreno

pmalek-sumo:
> Also do we have test for cover that?

All tests that specify `"prometheus-operator.kubeTargetVersionOverride"`as `1.13.0-0` should have new rules added.
Otherwise it shouldn't be added. This is reflected in tests.

sumo-drosiek (approved):

PR #1303, opened 2021-01-08
Add v2 migration logic for prometheus retention

pmalek-sumo:
###### Description

Add v2 migration logic for prometheus retention (`prometheus-operator.prometheus.prometheusSpec.retention`)

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/upgrade-2.0.0.sh
+++ b/deploy/helm/sumologic/upgrade-2.0.0.sh
@@ -177,6 +177,25 @@ function migrate_prometheus_operator_to_kube_prometheus_stack() {
   yq d -i "${TEMP_FILE}" "prometheus-operator"
 }
 
+function migrate_prometheus_retention_period() {
+  if [[ -z "$(yq r "${TEMP_FILE}" -- 'prometheus-operator')" ]]; then
+    return
+  fi
+  local RETENTION
+  readonly RETENTION="$(yq r "${TEMP_FILE}" -- 'prometheus-operator.prometheus.prometheusSpec.retention')"
+
+  if [[ -z "${RETENTION}" ]]; then
+    return
+  fi
+
+  if [[ "${RETENTION}" == "7d" ]]; then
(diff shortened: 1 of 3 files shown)


astencel-sumo (dismissed):

deploy/helm/sumologic/upgrade-2.0.0.sh
```diff
@@ -177,6 +177,25 @@ function migrate_prometheus_operator_to_kube_prometheus_stack() {
...
+  readonly RETENTION="$(yq r "${TEMP_FILE}" -- 'prometheus-operator.prometheus.prometheusSpec.retention')"
+
+  if [[ -z "${RETENTION}" ]]; then
+    return
+  fi
+
+  if [[ "${RETENTION}" == "7d" ]]; then
+    info "Changing prometheus retention period from 7d to 1d"
+  elif [[ "${RETENTION}" != "1d" ]]; then
+    warning "Changing prometheus retention period from (custom set) ${RETENTION} to 1d"
```
perk-sumo, line 194: I don't think we should change it in such case.

perk-sumo (changes requested):

deploy/helm/sumologic/upgrade-2.0.0.sh
perk-sumo: I mean - if someone changed that from the default 7d they probably knew what they were doing.
I think we should rather inform user about the change of the default value for the chart and explicitly say that we leave it as it is.

pmalek-sumo: Changed, please recheck.

astencel-sumo (approved):

perk-sumo (approved):
👍

deploy/helm/sumologic/upgrade-2.0.0.sh
perk-sumo: Thanks :)

PR #1337, opened 2021-01-14
Fix secret cleanup

pmalek-sumo:
###### Description

* Add separate `ServiceAccount`, `ClusterRole` and `ClusterRoleBinding` for cleanup job
* Change cleanup job helm hook weight from `0` to `3` so that the RBAC resources (listed above) are created before the job runs

---

###### Testing performed


sumo-drosiek (approved):

perk-sumo (approved):
🎸

Issue #1412, opened 2021-02-11
v2 upgrade script: remoteTimeout is not added to custom remoteWrite section


astencel-sumo:
**Describe the bug**

When upgrading from `v1` to `v2`, if you have your Prometheus `remoteWrite` section overridden (that is entry `prometheus-operator.prometheus.prometheusSpec.remoteWrite` in `values.yaml`), the upgrade script does not add `remoteTimeout` parameter to the Fluentd `remoteWrite` entries. We consider this to be a bug, as the `remoteTimeout` parameter is crucial to ensure good performance when sending data from Prometheus to Fluentd.

**To Reproduce**

1. Prepare a `values.yaml` file for `v1` collection that has `prometheus-operator.prometheus.prometheusSpec.remoteWrite` entry customized.
2. Run the upgrade script [upgrade-2.0.0.sh](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/ee6257fce3a5a7f4ff5614fdc12b481f6ef57cc2/deploy/helm/sumologic/upgrade-2.0.0.sh)

```shell
./upgrade-2.0.0.sh values.yaml
```

3. Observe the `prometheus-operator.prometheus.prometheusSpec.remoteWrite` in the output file `new_values.yaml`.

**Expected behavior**

The `prometheus-operator.prometheus.prometheusSpec.remoteWrite` section in the output file should contain `remoteTimeout` parameter for the Fluentd `remoteWrite` entries.

**Actual behaviour**

The `prometheus-operator.prometheus.prometheusSpec.remoteWrite` section in the output file does not contain `remoteTimeout` parameter for the Fluentd `remoteWrite` entries.

perk-sumo:
Backport to `v2.0`: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1475

PR #1467, opened 2021-03-02
Fix v2.0.0 migration script not migrating remoteTimeout

pmalek-sumo:
###### Description

Fixes https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/1412

### DIFF
--- a/deploy/helm/sumologic/upgrade-2.0.0.sh
+++ b/deploy/helm/sumologic/upgrade-2.0.0.sh
@@ -252,6 +252,7 @@ function kube_prometheus_stack_update_remote_write_regexes() {
 	  path: 'kube-prometheus-stack.prometheus.prometheusSpec.remoteWrite.[+]'
 	  value:
 	    url: ${URL_METRICS_OPERATOR_RULE}
+	    remoteTimeout: 5s
 	    writeRelabelConfigs:
 	      - action: keep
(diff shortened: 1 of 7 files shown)


tests/upgrade_v2_script/static/prometheus_remote_write_timeout.output.yaml
```diff
@@ -1,79 +1,58 @@
 kube-prometheus-stack:
   prometheus:
     prometheusSpec:
-      scrapeInterval: 30s
```
kkujawa-sumo, line 4: Why this file contains changes not related to remoteTimeout?

pmalek-sumo: Good catch! Should be fixed now.

kkujawa-sumo: It seems that now changes are related only to remoteTimeout but it different order. May we avoid reordering? I'll be easier to review.

pmalek-sumo: I guess the reason for this is how `yq` approaches yaml maps/hashmaps/? 🤷

perk-sumo (approved):

PR #1490, opened 2021-03-10
Bump OTC, add compression, use OTLPHTTP exporter

mat-rumian:
###### Description

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -1933,7 +1933,7 @@ otelagent:
     memBallastSizeMib: "250"
     image:
       repository: "public.ecr.aws/sumologic/opentelemetry-collector"
-      tag: "v0.21.1-sumo"
+      tag: "0.22.0-sumo"
       pullPolicy: IfNotPresent
 
     ## Extra Environment Values - allows yaml definitions
@@ -2039,7 +2039,7 @@ otelcol:
     memBallastSizeMib: "683"
     image:
       repository: "public.ecr.aws/sumologic/opentelemetry-collector"
-      tag: "v0.21.1-sumo"
+      tag: "0.22.0-sumo"
       pullPolicy: IfNotPresent
 
     ## Extra Environment Values - allows yaml definitions


deploy/helm/sumologic/values.yaml
```diff
@@ -2208,17 +2208,18 @@ otelcol:
       # logging:
       #   loglevel: debug
       otlphttp:
-        endpoint: ${SUMO_ENDPOINT_DEFAULT_TRACES_SOURCE}
+        traces_endpoint: ${SUMO_ENDPOINT_DEFAULT_TRACES_SOURCE}
```
pmalek-sumo, line 2211: This is a breaking change for end users. Is  tracing via otelcol considered a beta feature so that it is expected or do we need to explicitly warn users whenever we release next version?

pmalek-sumo (commented):
Please fix the tests

deploy/helm/sumologic/values.yaml
mat-rumian: @pmaciolek what do you think about it?

deploy/helm/sumologic/values.yaml
```diff
@@ -1933,7 +1933,7 @@ otelagent:
     memBallastSizeMib: "250"
     image:
       repository: "public.ecr.aws/sumologic/opentelemetry-collector"
-      tag: "0.19.2-sumo"
+      tag: "v0.21.1-sumo"
```
pmm-sumo, line 1936: Perhaps we could go with `v0.22.0-sumo`?

deploy/helm/sumologic/values.yaml
pmm-sumo: Why is this considered a breaking change @pmalek-sumo ?

pmalek-sumo: Yup. Someone could've specified a custom `endpoint` and now it won't work i.e. it will have to be changed to `traces_endpoint`, otherwise this custom value won't be used when installing the chart.

perk-sumo: We usually keep this kind of changes as PRs with `breaking change` label.
They are picked up during next breaking change release.

perk-sumo (commented):
👍

PR #1528, opened 2021-03-25
Add specific 2.0 version to helm command when installing the chart

### LINKED ISSUE #1502: Document how to use specific version of the helm chart
Backport it to `v1.3` and `v2.0`

pmalek-sumo:
Fixes #1502

### DIFF
--- a/deploy/docs/Installation_with_Helm.md
+++ b/deploy/docs/Installation_with_Helm.md
@@ -104,6 +104,7 @@ The following command will install the Sumo Logic chart with the release name
 
 ```bash
 helm upgrade --install my-release sumologic/sumologic \
+  --version=2.0 \
   --set sumologic.accessId="<SUMO_ACCESS_ID>" \
   --set sumologic.accessKey="<SUMO_ACCESS_KEY>" \
   --set sumologic.clusterName="<MY_CLUSTER_NAME>"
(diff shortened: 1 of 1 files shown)


deploy/docs/Installation_with_Helm.md
```diff
@@ -104,6 +104,7 @@ The following command will install the Sumo Logic chart with the release name
 
 ```bash
 helm upgrade --install my-release sumologic/sumologic \
+  --version=2.0 \
```
astencel-sumo, line 107: Should this change live in `main` branch or only  in `release-v2.0` branch?

Setting this on `main` means we need to update it when we release `v2.1` and so forth.

frankreno: Will this install the latest 2.0 every time? WE will need to keep this updated with every release right?

frankreno:
We will also need to do this in the non-helm install in the k8s tools command.

Rather than adding `--version` to every permutation, would it be better to simply add a new section after 

```If the namespace does not exist, you can add the `--create-namespace` flag.```

That shows a single use of the version flag? The intention of the issue was just to show users who dont know helm how to install specific versions.

pmalek-sumo:
This was intended to be targeted for `release-v2.0` branch so no updates would need to be done to those commands anymore, at this branch.

pmalek-sumo:
Superseeded by https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1529

PR #1529, opened 2021-03-25
Release 2.0 document installation of specific version

pmalek-sumo:
###### Description

Specify `--version` flag when installing with helm so that users don't get surprised by getting e.g. 2.0 (with breaking changes) installed.

Addresses https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/1502

### DIFF
--- a/deploy/docs/Installation_with_Helm.md
+++ b/deploy/docs/Installation_with_Helm.md
@@ -104,6 +104,7 @@ The following command will install the Sumo Logic chart with the release name
 
 ```bash
 helm upgrade --install my-release sumologic/sumologic \
+  --version=2.0 \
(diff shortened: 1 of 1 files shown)


pmalek-sumo:
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1528#issuecomment-806807595

@frankreno This section already exists in https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/release-v2.0/deploy/docs/Installation_with_Helm.md#upgrading-sumo-logic-collection

Do we want to leave it as is or merge this PR to be more explicit?

astencel-sumo (approved):

PR #1537, opened 2021-03-30
Prepare v2.0.6

perk-sumo:
###### Description

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: sumologic
-version: 2.0.5
-appVersion: 2.0.5
+version: 2.0.6
+appVersion: 2.0.6
 description: A Helm chart for collecting Kubernetes logs, metrics, traces and events into Sumo Logic.
 type: application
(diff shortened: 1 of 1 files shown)


kkujawa-sumo (approved):

deploy/helm/sumologic/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: sumologic
-version: 2.0.5
-appVersion: 2.0.5
+version: 2.0.6
```
astencel-sumo, line 3: Don't we need a `rc` version?

perk-sumo: There are mostly doc changes.
The only real fix is for setup job - and we run this with `v2.1` already so I don't feel like we do need `rc.1` for that.
I'm OFC open for discussion :)

pmalek-sumo (approved):

deploy/helm/sumologic/Chart.yaml
pmalek-sumo: Ok for taking into account the diff/log

PR #1550, opened 2021-04-12
enabling custom labes for event, metrics and logs services

mozz-lx:
###### Description

Enabled custom labels in fluentd logs, events and metrics services.

---

###### Testing performed


pmalek-sumo (commented):
Generally the idea seems nice, 1 minor comment though:

This shouldn't really piggy back on `fluentd.podLabels` as it's not pods that we attach these labels to here but rather services. Hence I suggest to add new value e.g. `fluentd.serviceLabels` similar to `fluentd.podLabels` and use that

pmalek-sumo:
One more thing actually: please make this a template so that we don't need the `if` everywhere just in the template 👍

mozz-lx:
> One more thing actually: please make this a template so that we don't need the `if` everywhere just in the template 👍

Hello @pmalek-sumo,

Changed fluentd.podLabels to fluentd.serviceLabels. I'm very familiar with the templates that you mentioned. I'm hopping this can work for now.

Marco.

pmalek-sumo (approved):

pmalek-sumo:
> > One more thing actually: please make this a template so that we don't need the `if` everywhere just in the template 👍
> 
> Hello @pmalek-sumo,
> 
> Changed fluentd.podLabels to fluentd.serviceLabels. I'm not familiar with the templates that you mentioned. I'm hopping this can work for now.
> 
> Marco.

Thanks for changing this. I'll merge it and introduce the aforementioned templates.

PR #1555, opened 2021-04-13
Add markdown-link-check

pmalek-sumo:
###### Description

Add markdown-link-check ([action](https://github.com/gaurav-nelson/github-action-markdown-link-check) & [tool](https://github.com/tcort/markdown-link-check)) in order to check for dead links.

---

###### Testing performed

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -48,6 +48,14 @@ jobs:
           make helm-dependency-update
           make helm-lint
 
+  markdown-link-check:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@master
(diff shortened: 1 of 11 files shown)


sumo-drosiek:
Ideally we link all files so we doesn't break the paths while moving them

.github/workflows/pull_requests.yml
```diff
@@ -48,6 +48,14 @@ jobs:
           make helm-dependency-update
           make helm-lint
 
+  markdown-link-check:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@master
```
astencel-sumo, line 54: Is there a reason for `actions/checkout@master` and not `@v2` as the other actions?

astencel-sumo (approved):
👏

PR #1576, opened 2021-04-16
Make markdown-links-lint work on Darwin

pmalek-sumo:
###### Description

Make `markdown-links-lint` work on Darwin by using `ggrep`.

---

###### Testing performed

### DIFF
--- a/ci/markdown_links_lint.sh
+++ b/ci/markdown_links_lint.sh
@@ -1,5 +1,15 @@
 #!/usr/bin/env bash
 
+GREP="grep"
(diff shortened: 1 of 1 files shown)


sumo-drosiek:
We should check if `grep`/`ggrep` exists on the system before we use it

pmalek-sumo:
> We should check if `grep`/`ggrep` exists on the system before we use it

Please check again :)

sumo-drosiek (approved):

sumo-drosiek:
LGTM :)

PR #1577, opened 2021-04-16
Enabling remote write for postgresql metrics

himanshu219:
###### Description

This PR enables metrics for PostgreSQL 

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -2043,7 +2043,7 @@ kube-prometheus-stack:
           remoteTimeout: 5s
           writeRelabelConfigs:
             - action: keep
-              regex: (?:postgresql_(blks_.*|buffers_.*|checkpoints_.*|db_size|deadlocks|flush_lag|heap_blks_.*|idx_blks_.*|idx_scan|idx_tup_.*|index_size|n_dead_tup|n_live_tup|n_tup_.*|num_locks|numbackends|replay_lag|replication_.*|seq_scan|seq_tup_read|stat_ssl_compression_count|table_size|tidx_blks_.*|toast_blks_.*|tup_.*|write_lag.*|xact_.*))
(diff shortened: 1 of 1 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -1990,6 +1990,62 @@ kube-prometheus-stack:
...
+        ## postgresql_heap_blks_hit
+        ## postgresql_idx_blks_read
+        ## postgresql_idx_blks_hit
+        ## postgresql_index_size
+        ## postgresql_table_size
+        - url: http://$(FLUENTD_METRICS_SVC).$(NAMESPACE).svc.cluster.local:9888/prometheus.metrics.postgresql
+          remoteTimeout: 5s
+          writeRelabelConfigs:
+            - action: keep
+              regex: (?:postgresql_(blks_hit.*|blks_read.*|buffers_backend.*|buffers_checkpoint.*|buffers_clean.*|checkpoints_req.*|checkpoints_timed.*|db_size.*|deadlocks.*|flush_lag.*|heap_blks_hit.*|heap_blks_read.*|idx_blks_hit.*|idx_blks_read.*|idx_scan.*|idx_tup_fetch.*|idx_tup_read.*|index_size.*|n_dead_tup.*|n_live_tup.*|n_tup_del.*|n_tup_hot_upd.*|n_tup_ins.*|n_tup_upd.*|num_locks.*|numbackends.*|replay_lag.*|replication_delay.*|replication_lag.*|seq_scan.*|seq_tup_read.*|stat_ssl_compression_count.*|table_size.*|tup_deleted.*|tup_fetched.*|tup_inserted.*|tup_returned.*|tup_updated.*|write_lag.*|xact_commit.*|xact_rollback.*)
```
pmalek-sumo, line 2046: * This regex is invalid: it doesn't have an matching closing parenthesis.
* Do we need to add a wildcard to every single pattern here? I'd try to avoid that. If we can't then please update the comment above to reflect the asterisks in regexes.

pmalek-sumo (changes requested):

deploy/helm/sumologic/values.yaml
himanshu219: strange it didn't failed when I redeployed with this command 
helm3 upgrade --install sumo-release sumologic/sumologic --namespace=sumo-collector --create-namespace -f sumologic-values.yaml --set telegraf-operator.enabled=true ...

himanshu219: done

sumo-drosiek: We should have as minimal regex (in terms of matches) as possible. We should avoid wildcard as @pmalek-sumo said.

pmalek-sumo (approved):
👍  @sumo-drosiek any comments? I believe this is alright now

sumo-drosiek (approved):
Everything is fine. Regex matches for 45 strings and all of the strings from list (45) are matched.
Thanks :rocket:

PR #1599, opened 2021-05-11
Fix additionalLabels in collection-fluent-bit

voidlily:
In https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/690 the prometheus additionalLabels were updated to use `sumologic.com/app` instead of `app`. The entry for `collection-fluent-bit` was missed, causing the helm chart to fail when using the built in prometheus stack due to duplicate labels.

This change should fix that so the built in prometheus stack works with collection-fluent-bit.

###### Description

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -1333,7 +1333,7 @@ kube-prometheus-stack:
             sumologic.com/scrape: "true"
       - name: collection-fluent-bit
         additionalLabels:
-          app: collection-fluent-bit
+          sumologic.com/app: collection-fluent-bit
         endpoints:
           - port: http
             path: /api/v1/metrics/prometheus


sumo-drosiek:
Hi @voidlily
Thank you for your PR!

This looks fine, but need to get more information. I would rather prefer to add new label instead of replacing current, unless I understand what exactly happens

> causing the helm chart to fail when using the built in prometheus stack due to duplicate labels

Could you elaborate what does it exactly mean? What do you mean by `built in prometheus stack`? Are you operating in openshift? 

I tested installation in both cases (before and after change) in our vagrant environment and didn't get any mentioned error.

gempesaw:
@sumo-drosiek 👋🏽 when I tried to install the helm chart, it generated an error message like 

```
  line 82: mapping key "app" already defined at line 77 for the following resource:
```

and then the lines in question are the `metadata.labels` of this `ServiceMonitor` thing: (arrows added manually for emphasis)

```yaml
- apiVersion: monitoring.coreos.com/v1
  kind: ServiceMonitor
  metadata:
    name: collection-fluent-bit
    namespace: sumologic
    labels:
      app: kube-prometheus-stack-prometheus # <--
      chart: kube-prometheus-stack-12.3.0
      release: "sumologic"
      heritage: "Helm"
      app: collection-fluent-bit            # <--
  spec:
    endpoints:
    - path: /api/v1/metrics/prometheus
      port: http
    namespaceSelector:
      matchNames:
      - $(NAMESPACE)
    selector:
      matchLabels:
        app.kubernetes.io/name: fluent-bit
        sumologic.com/scrape: "true"
```

sumo-drosiek (approved):

sumo-drosiek:
@gempesaw Thanks for the comment. It's clear for me now :+1:

PR #1629, opened 2021-05-27
docs: remove alpha releases docs (#1626) - backport release-v2.0

pmalek-sumo:
###### Description

Manual backport of #1626 

---

###### Testing performed


deploy/docs/Alpha_Release_Guide.md
```diff
@@ -1,42 +0,0 @@
...
-
-## Using an alpha release
-
-### Helm installation
-
-Follow the [Helm installation guide](./Installation_with_Helm.md), but override both the Helm chart version and the Docker image version.
-
-```
-helm repo update
-helm install collection sumologic/sumologic --namespace sumologic --create-namespace ... --version=0.11.3-alpha --set image.tag="0.11.3-alpha"
```
kkujawa-sumo, line 27: This instruction is still valid for alpha releases, they are available in main helm repository for Helm Chart, see: https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/gh-pages.

Dev versions of Helm Chart are built per merge to master branch and they are stored in different Helm repository (https://sumologic.github.io/sumologic-kubernetes-collection/dev)

pmalek-sumo: We don't really use the notion of `alpha` releases now. I believe removing that and leaving just the docs about dev versions (what used to be known as alpha) should be fine.

> Dev versions of Helm Chart are built per merge to master branch and they are stored in different Helm repository (https://sumologic.github.io/sumologic-kubernetes-collection/dev)

that's what `deploy/docs/Dev.md` describes right?

kkujawa-sumo: Yes, this is what `Dev.md` describes but it is also written `Dev.md` "This document contains information helpful for developers."
If we do not plan to make alpha releases I'm ok with this PR but adding this doc to the list [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1629/files#diff-0db16276644fc4f21eae05d761a230f6438feca8b7c0a92427c9d3b3ea85726aR18)  encourages to using dev releases (for me dev version is not alpha version). Do we want to make this?

I have mixed feelings because I'm used to software lifecycle which is shown here: https://en.wikipedia.org/wiki/Software_release_life_cycle#/media/File:Software_dev2.svg

pmalek-sumo: To me we used to use wrong notion of alpha releases which we understood by "every build" which shouldn't be the case and to what now we refer to as "dev" builds - what I believe is more accurate.

> ... but it is also written Dev.md "This document contains information helpful for developers."

That's true but it also says:

> We recommend testing dev releases on non-production clusters. <br/> These releases are generated continuously from contributions to this repo and may not be fully tested.

And by merging this PR we do not refer to any builds by `alpha` builds. We'll simply have (as we do currently) dev builds (builds for every commit) and release builds with all the stages we deem necessary (sometimes only with `rc`s, sometimes with `alpha` releases as well - just like in your wiki link).

kkujawa-sumo (approved):

deploy/docs/Alpha_Release_Guide.md
kkujawa-sumo: After re-thinking, it is ok for me to merge this PR. Nothing wrong is happening there is only exposed information about dev builds.

Issue #1612, opened 2021-05-25
Prefix all application metrics with `prometheus.metrics.applications` instead of `prometheus.metrics`


sumo-drosiek:
Prefix all application metrics with `prometheus.metrics.applications` instead of `prometheus.metrics`

eg. https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/values.yaml#L1993

vs https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/values.yaml#L1930

PR #1631, opened 2021-05-27
refactor(prometheus): unify application metrics urls

sumo-drosiek:
###### Description

solves #1612

BREAKING CHANGE: This can be a breaking change in case someone added own remoteWrite

---

###### Testing performed


pmalek-sumo (approved):

perk-sumo:
Moving to v2.2 and removing the `breaking change` label as this is not a breaking change.
It works the same before and after this patch.

PR #1650, opened 2021-06-11
feat: annotation_match for fluentd metadata plugin configurable in values.yaml

kkujawa-sumo:
###### Description

annotation_match for fluentd metadata plugin configurable in values.yaml

---

###### Testing performed

Example tested configuration:

values.yaml:
```yaml
sumologic:
  collectorName: "kubernetes-test" 
  accessId: ""
  accessKey: ""
  endpoint: ""

fluentd:
  logs:
    containers:
      perContainerAnnotationsEnabled: true
      perContainerAnnotationPrefixes:
        - "sumologic.com/"
        - "tailing-sidecar.sumologic.com/"
        
  metadata:
   annotation_match:
    - "\".*sumologic\\.com.*\""
```

pod1.yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: pod1
  namespace: sumologic
  annotations:
    sumologic.com/count1.sourceCategory: test-source-category5
    sumologic.com/count2.sourceCategory: test-source-category6
  labels:
    app: pod-with-annotations
spec:
  containers:
  - name: count1
    image: busybox
    args:
    - /bin/sh
    - -c
    - >
      i=0;
      while true;
      do
        echo "example0: $i $(date)"
        echo "example1: $i $(date)"
        echo "example2: $i $(date)"
        i=$((i+1));
        sleep 5;
      done
    volumeMounts:
    - name: varlog
      mountPath: /var/log
    - name: varlogconfig
      mountPath: /varconfig/log
  - name: count2
    image: busybox
    args:
    - /bin/sh
    - -c
    - >
      i=0;
      while true;
      do
        echo "example0: $i $(date)"
        echo "example1: $i $(date)"
        echo "example2: $i $(date)"
        i=$((i+1));
        sleep 5;
      done
    volumeMounts:
    - name: varlog
      mountPath: /var/log
    - name: varlogconfig
      mountPath: /varconfig/log
  volumes:
  - name: varlog
    emptyDir: {}
  - name: varlogconfig
    emptyDir: {}
```

pod2.yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: pod2
  namespace: sumologic
  annotations:
    tailing-sidecar.sumologic.com/count1.sourceCategory: test-source-category3
    tailing-sidecar.sumologic.com/count2.sourceCategory: test-source-category4
  labels:
    app: pod-with-annotations
spec:
  containers:
  - name: count1
    image: busybox
    args:
    - /bin/sh
    - -c
    - >
[...]

PR #1654, opened 2021-06-14
[manual backport] feat: annotation_match for fluentd metadata plugin configurable in values.yaml

kkujawa-sumo:
manual backport of  #1650

###### Description

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -311,6 +311,9 @@ fluentd:
     output: false
 
   metadata:
+    ## Option to control capturing of annotations by metadata filter plugin.
+    annotation_match:
+      - '"sumologic\\.com.*"'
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -311,6 +311,9 @@ fluentd:
     output: false
 
   metadata:
+    ## Option to control capturing of annotations by metadata filter plugin.
+    annotation_match:
+      - '"sumologic\\.com.*"'
```
sumo-drosiek, line 316: I didn't saw this before, but I don't think that adding `"` because helm magically converts list to fluentd format is a good idea. We should rather use some helm template function or write helper to mitigate this. Otherwise we have to explain in values.yaml how to properly build the `fluentd.metadata.annotation_match`

kkujawa-sumo:
There is no longer need to add this as part of 2.1 release.

PR #1668, opened 2021-06-17
refactor(ci): use forked label checker to allow empty valid label set

pmalek-sumo:
###### Description

Revert to using `verify-pr-label-action` action but forked and updated so that it doesn't require a "positive" valid label set.

ref: https://github.com/pmalek-sumo/verify-pr-label-action

---

###### Testing performed


perk-sumo (approved):
🚀

kkujawa-sumo (approved):

PR #1669, opened 2021-06-17
doc: add links for different documentation versions to doc starting page

perk-sumo:
###### Description

---

###### Testing performed

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,5 +1,14 @@
 # Deployment Guide
 
+Documentation versions:
+| version                                                                                                   | status                                     |
+|-----------------------------------------------------------------------------------------------------------|--------------------------------------------|
+| [v2.1](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v2.1/deploy/README.md)   | current / supported                        |
+| [v2.0](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v2.0/deploy/README.md)   | deprecated / unsupported                   |
+| [v1.3](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v1.3/deploy/README.md)   | deprecated / supported until 14th Jul 2021 |
+| [v0.17](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v0.17/deploy/README.md) | deprecated / unsupported                   |
+
+
 This page has instructions for collecting Kubernetes logs, metrics, and events;
 enriching them with deployment, pod, and service level metadata; and sending them to Sumo Logic.
 See our [documentation guide](https://help.sumologic.com/Solutions/Kubernetes_Solution)


deploy/README.md
```diff
@@ -1,5 +1,10 @@
 # Deployment Guide
 
+Documentation versions:
+- [v2.1](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v2.1/deploy/README.md)
+- [v1.3](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v1.3/deploy/README.md)
+- [v0.17](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v0.17/deploy/README.md)
```
pmalek-sumo, line 6: Why add this if 0.17 is unsupported?

perk-sumo: I think it should show the latest minor for all the majors, even the unsupported ones - just in case someone is still using that (I would not be surprised). 
But I'm open to discussion OFC.

pmalek-sumo: Ok fair enough. We explicitly state on the main readme page what is supported and what isn't.

pmalek-sumo (approved):

deploy/README.md
rplahn: Would it make sense to include the support status vs. just the versions and links? Maybe just use the same content from the main installation page. 
`Detailed instructions are available in our Installation Guides below.

Sumo Logic Helm Chart Version

| version | status |
|--|--|
|[v2.1](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v2.1/deploy/README.md) | current / supported  |
|[v2.0](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v2.0/deploy/README.md) | deprecated / unsupported  |
|[v1.3](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v1.3/deploy/README.md) | deprecated / supported until 14th Jul 2021 |
|[v0.17](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v0.17/deploy/README.md) | deprecated / unsupported |`

pmalek-sumo: This would be nice but it would duplicate the information IMHO and has a chance of going out of sync with the main README

rplahn: IMHO, the clarity that would be most helpful is an indication NOT to use the main branch but instead pull from the Releases for stable versions. But I don't want to beat this to death. I've learned the lesson for myself so perhaps it's clear enough for others.

pmalek-sumo: Ok, so I've checked if it's possible to include markdown files on in another and unfortunately it's not possible but I agree with @rplahn: if this makes is clearer for the users and saves us investigation time, let's do it.

perk-sumo: That makes sense; I've included whole table.

PR #1701, opened 2021-07-29
linux only containers should target linux workers

poblahblahblah:
###### Description

We have a few kubernetes on windows clusters that we want to pull some logs in from, but many (all?) of these containers are linux only.

This sets a nodeSelector to target linux workers via kubernetes.io/os

---

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/templates/cleanup/job.yaml
+++ b/deploy/helm/sumologic/templates/cleanup/job.yaml
@@ -38,6 +38,8 @@ spec:
             mountPath: /configmap
           - name: cleanup
             mountPath: /etc/terraform
+      nodeSelector:
+        kubernetes.io/os: linux
       containers:
         - name: cleanup
           image: {{ .Values.sumologic.setup.job.image.repository }}:{{ .Values.sumologic.setup.job.image.tag }}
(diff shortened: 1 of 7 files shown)


poblahblahblah:
This works... sort of.

The problem is that sub charts would also need to be updated to target linux if the container is linux only.

pmalek-sumo:
@poblahblahblah Thanks for your contribution!

This looks nice but what I believe will happen is that people will want to have that configurable at some point.

I believe we can start by changing what you have done to be configurable with `nodeSelector` property for:

* fluentd pods with one option under `fluentd.nodeSelector`
* otelcol pods with `otelcol.nodeSelector`
* otelagent pods with `otelagent.nodeSelector`
* setup and cleanup jobs with `sumologic.setup.job.nodeSelector` (this is already there it just needs to be added to be used in cleanup job the same way as https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e504b3acb6f227d8db3c973afe887dc4b75d459d/deploy/helm/sumologic/templates/setup/setup-job.yaml#L31-L34

The defaults for all of those can be set to `kubernetes.io/os: linux` just as in your version of this PR.

Sub charts are a different thing and we'd need to verify if `nodeSelector` can be configured for those that are used in our chart.

poblahblahblah:
@pmalek-sumo I can't believe I missed those!

I'll sweep through and check for any missing nodeSelectors and add the value as needed.

poblahblahblah:
We got this all working since all deployments/replicasets/daemonsets we needed were able to have nodeSelectors specified.

sumo-drosiek:
Related PR: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3570

PR #1747, opened 2021-09-14
Adding remote write configs for Cassandra Metrics

anphn-mtt:
###### Description

Added Remote/Write Configs for Cassandra metrics:
List of Metrics are on following page:

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cassandra

Metrics follow following fomart
``` bash 
        ## cassandra_CacheMetrics_ChunkCache_OneMinuteRate
        ## cassandra_ClientMetrics_connectedNativeClients_Value
        ## cassandra_ClientMetrics_RequestDiscarded_OneMinuteRate
        ## cassandra_CommitLogMetrics_CompletedTasks_Value
        ## cassandra_CommitLogMetrics_PendingTasks_Value
        ## cassandra_DroppedMessageMetrics_Dropped_OneMinuteRate
        ## cassandra_java_GarbageCollector_*_CollectionCount
        ## cassandra_java_GarbageCollector_*_CollectionTime
        ## cassandra_java_GarbageCollector_*_LastGcInfo_duration
        ## cassandra_java_GarbageCollector_*_LastGcInfo_GcThreadCount
        ## cassandra_java_GarbageCollector_*_LastGcInfo_memoryUsageAfterGc_*_used
        ## cassandra_java_GarbageCollector_*_LastGcInfo_memoryUsageBeforeGc_*_used
        ## cassandra_java_Memory_HeapMemoryUsage_used
        ## cassandra_java_OperatingSystem_AvailableProcessors
        ## cassandra_java_OperatingSystem_FreePhysicalMemorySize
        ## cassandra_java_OperatingSystem_SystemCpuLoad      
        ## cassandra_java_OperatingSystem_TotalPhysicalMemorySize
        ## cassandra_java_OperatingSystem_TotalSwapSpaceSize
        ## cassandra_Net_FailureDetector_DownEndpointCount
        ## cassandra_Net_FailureDetector_UpEndpointCount
        ## cassandra_TableMetrics_AllMemtablesHeapSize_Value
        ## cassandra_TableMetrics_AllMemtablesLiveDataSize_Value
        ## cassandra_TableMetrics_CompactionBytesWritten_Count
        ## cassandra_TableMetrics_EstimatedPartitionCount_Value
        ## cassandra_TableMetrics_KeyCacheHitRate_Value
        ## cassandra_TableMetrics_LiveSSTableCount_Value
        ## cassandra_TableMetrics_MemtableColumnsCount_Value
[...]

PR #1760, opened 2021-09-16
fix(ci): don't run PRs labels check on pull_request events to prevent failures on PR from forks

pmalek-sumo:
The gist of this is to avoid failures on PRs from forks like this: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1747/checks?check_run_id=3616995450


sumo-drosiek:
@pmalek-sumo This check is not required, so I'm not sure if we should remove it

pmalek-sumo:
> @pmalek-sumo This check is not required, so I'm not sure if we should remove it

I'm pretty sure it will still run as `pull_request_target`

sumo-drosiek (approved):

PR #1770, opened 2021-09-20
fix(otelcol): use source metdata set by sourceprocessor

pmalek-sumo:
Rely on source metadata set by source processor and don't construct it in the exporter.


sumo-drosiek (approved):
Thank you

PR #1772, opened 2021-09-21
Add kubelet, systemd filtration and refactor containers exporter

sumo-drosiek:
- unify exporting for systemd, collectd and containers logs
- add filtration for systemd like we already have in fluentd
- add kubelet pipeline

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -2938,12 +2938,19 @@ otelcol:
             keys:
               - _SYSTEMD_UNIT
               - _HOSTNAME
+              - SYSLOG_FACILITY
+              - PRIORITY
           source/systemd:
             collector: '{{ .Values.sumologic.collectorName | default .Values.sumologic.clusterName | quote }}'
             source_name: '{{ .Values.fluentd.logs.systemd.sourceName | quote }}'
             source_category: '{{ .Values.fluentd.logs.systemd.sourceCategory | quote }}'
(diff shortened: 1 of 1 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -2987,7 +2993,20 @@ otelcol:
...
-                - sumologic/systemd
+                - sumologic
+            logs/kubelet:
+              receivers:
+                - fluentforward
+              processors:
+                - memory_limiter
+                - filter/include_fluent_tag_host
+                - attributes/remove_fluent_tag
+                - filter/include_kubelet
```
pmalek-sumo, line 3004: Actually as I think of it now I believe we can swap those 2 to save processing of "non kubelet" entries, i.e. no need to remove the `fluent.tag` from a record that will be filtered out anyway.

The same can be done for systemd

sumo-drosiek: so we should move it as far as possible, we have additional filtering in sourceprocessor for example

pmalek-sumo: What do you mean "as far as possible"? Shouldn't we filter out as soon as possible in the pipeline?

sumo-drosiek: I mean moving `- attributes/remove_fluent_tag` just before `batch`

pmalek-sumo (approved):

PR #1776, opened 2021-09-22
fix(otelcol): fix match_type in filterprocessor's config for logs pipeline

pmalek-sumo:
It seems that `filterprocessor` doesn't validate the `match_type` used in its config.

I've created an issue for that: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5382

In the meantime let's just correct it in our `values.yaml`


perk-sumo (approved):
👀

PR #1777, opened 2021-09-23
fix(otelcol): fix systemd logs' source name

pmalek-sumo:
In oder to keep the metadata enrichment behavior from fluentd which is to extract the second element from `fluent.tag` https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/d87aab8d99d63c3ff364c72b85aa9cd2cba4855f/deploy/helm/sumologic/conf/logs/fluentd/logs.source.systemd.conf#L78 let's do the same in otelcol's config and set that as source name for systemd logs.

Additionally document the change in Changelog;.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - otelcol: add systemd logs pipeline (#1767)
 
   - This change introduces logs metadata enrichment with Sumo Open Telemetry
-    distro for systemd logs.
+    distro for systemd logs (when `sumologic.logs.metadata.provider` is set to
+    `otelcol`)
 
-    One notable change in behavior with respect to `fluentd` is that
-    `.Values.fluentd.logs.systemd.sourceName` is respected and
-    will set `_sourceName` of processed logs while fluentd would set this field
-    to the corresponding systemd serice's name e.g. `docker` for `docker.service`.
+    One notable change comparing the new behavior to `fluentd` metadat enrichment
+    is that setting source name in `sourceprocessor` configuration is respected
+    i.e.  whatever is set in
+    `otelcol.metadata.logs.config.processors.source/systemd:.source_name` will be
+    set as source name for systemd logs.
+
+    The old behavior is being retained i.e. extracting the source name from
+    `fluent.tag` using `attributes/extract_systemd_source_name_from_fluent_tag`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,12 +12,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-    distro for systemd logs.
-
-    One notable change in behavior with respect to `fluentd` is that
-    `.Values.fluentd.logs.systemd.sourceName` is respected and
-    will set `_sourceName` of processed logs while fluentd would set this field
-    to the corresponding systemd serice's name e.g. `docker` for `docker.service`.
+    distro for systemd logs (when `sumologic.logs.metadata.provider` is set to
+    `otelcol`)
+
+    One notable change comparing the new behavior to `fluentd` metadat enrichment
```
sumo-drosiek, line 18:
```suggestion
    One notable change comparing the new behavior to `fluentd` metadata enrichment
```

CHANGELOG.md
```diff
@@ -12,12 +12,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    is that setting source name in `sourceprocessor` configuration is respected
+    i.e.  whatever is set in
+    `otelcol.metadata.logs.config.processors.source/systemd:.source_name` will be
```
sumo-drosiek, line 21:
```suggestion
    `otelcol.metadata.logs.config.processors.source/systemd.source_name` will be
```

vagrant/values.yaml
```diff
@@ -170,6 +171,13 @@ otelcol:
                 record_attributes:
                   - key: _SYSTEMD_UNIT
                     value: snap.microk8s.daemon-kubelite.service
+          filter/include_systemd_snap_kubelite:
+            logs:
+              include:
+                match_type: strict
+                record_attributes:
+                  - key: _SYSTEMD_UNIT
+                    value: snap.microk8s.daemon-kubelite.service
```
sumo-drosiek, line 180: why do we need that?

sumo-drosiek: I see, `vagrant` :D

pmalek-sumo: To include kubelite lgos in `logs/kubelet` pipeline :) note: we don't have `kubelet` in vagrant env but `kubelite`

sumo-drosiek (approved):

PR #1792, opened 2021-09-29
chore(otelcol): bump ot distro version to 0.29.0-beta.0

pmalek-sumo:
This change unfortunately requires us to "hardcode" the `source_name` and `source_category` configuration options because the options that were used until now: `.Values.fluentd.logs.containers.sourceName` and `.Values.fluentd.logs.containers.sourceCategory` have keys specified in Sumo conventions.

Those cannot be used since at the stage in the pipeline where source processor resides we are still using OT naming convention.

Change in chart stems from change in OT distro's config (specifically in sourceprocessor): https://github.com/SumoLogic/sumologic-otel-collector/pull/254

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -2835,7 +2835,7 @@ otelcol:
   statefulset:
     image:
       repository: public.ecr.aws/sumologic/sumologic-otel-collector
-      tag: 0.0.27-beta.0
+      tag: 0.0.29-beta.0
       pullPolicy: IfNotPresent
   metadata:
     metrics:
@@ -3023,10 +3023,6 @@ otelcol:
             sending_queue:
               enabled: true
             metadata_attributes:
-              ## Required to be used in headers
-              - _sourceCategory
-              - _sourceHost
-              - _sourceName
               ## Attributes to be send as fields
               - _collector
               - k8s.*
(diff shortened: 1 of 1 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -2835,8 +2835,9 @@ otelcol:
   statefulset:
     image:
       repository: public.ecr.aws/sumologic/sumologic-otel-collector
-      tag: 0.0.27-beta.0
-      pullPolicy: IfNotPresent
+      tag: 0.0.29-beta.0
+      # pullPolicy: IfNotPresent
+      pullPolicy: Always
```
sumo-drosiek, line 2840: Why?

deploy/helm/sumologic/values.yaml
```diff
@@ -3163,8 +3157,8 @@ otelcol:
               - from: build_hostname
           source/containers:
             collector: '{{ .Values.sumologic.collectorName | default .Values.sumologic.clusterName | quote }}'
-            source_name: '{{ .Values.fluentd.logs.containers.sourceName | quote }}'
-            source_category: '{{ .Values.fluentd.logs.containers.sourceCategory | quote }}'
+            source_name: "%{k8s.namespace.name}.%{k8s.pod.name}.%{k8s.container.name}"
```
sumo-drosiek, line 3160: we could use helm function to translate attributes :trollface:
seriously :thinking:

deploy/helm/sumologic/values.yaml
pmalek-sumo: derp, fixed.

astencel-sumo (approved):
LGTM

PR #1800, opened 2021-10-01
feat(helm/otelcol): filter out telegraf.influxdata.com/inputs annotation

sumo-drosiek:
###### Description
As `telegraf.influxdata.com/inputs` can easily exceed [limit of characters](https://help.sumologic.com/Manage/Fields#limitations) we shouldn't send it as a field

---

###### Testing performed


pmalek-sumo (approved):
We could perhaps think of putting this logic into sourceprocessor since it's the source of those fields

sumo-drosiek:
IMHO, It should be handle by exporter as this is related to sending data, not processing them

PR #1803, opened 2021-10-05
feat(otelcol): do not send pod annotations

andrzej-stencel:
Narrow down the metadata regexes to explicitly only match on pod and namespace labels.

This prevents pod annotations from being sent to Sumo, as they all start with `pod_annotations_`.


pmalek-sumo (approved):

sumo-drosiek (changes requested):
Remove
```
              annotations:
                - tag_name: "pod_annotations_%s"
                  key: "*"
```

from k8s_tagger or ensure they won't be send as part of the log body

PR #1866, opened 2021-11-04
fix: change Vagrant IP address for vbox 6.1.28 compat

swiatekm-sumo:
###### Description

Trying to bring up Vagrant with Virtualbox 6.1.28 with the current setup results in the following error:

Screenshot 2021-11-04 at 10 16 02

As it turns out, Virtualbox 6.1.28 restricts the IP addresses permitted in host-only networks to 192.168.56.0/21. Strangely, there isn't any mention of this in the [changelog](https://www.virtualbox.org/wiki/Changelog). This restriction can be lifted via vbox config, but we don't care what our guest IP is, so changing that instead is simpler.

See: https://www.virtualbox.org/manual/ch06.html#network_hostonly for more information. The relevant passage is:

> On Linux, Mac OS X and Solaris Oracle VM VirtualBox will only allow IP addresses in 192.68.56.0/21 range to be assigned to host-only adapters. For IPv6 only link-local addresses are allowed. If other ranges are desired, they can be enabled by creating /etc/vbox/networks.conf and specifying allowed ranges there.

There's also discussion of this problem in the Vagrant forums [here](https://discuss.hashicorp.com/t/vagrant-2-2-18-osx-11-6-cannot-create-private-network/30984). And a bug report where the change is explicitly acknowledged [here](https://www.virtualbox.org/ticket/20626).

Note: Existing networks will continue to work even with Vbox 6.1.28. Not sure why, probably has to do with specifics of how Vagrant interacts with vbox.

###### Testing performed

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,7 +12,9 @@ Vagrant.configure('2') do |config|
   config.disksize.size = '50GB'
   config.vm.box_check_update = false
   config.vm.host_name = 'sumologic-kubernetes-collection'
-  config.vm.network :private_network, ip: "192.168.78.66"
+  # Vbox 6.1.28+ restricts host-only adapters to 192.168.56.0/21
+  # See: https://www.virtualbox.org/manual/ch06.html#network_hostonly
+  config.vm.network :private_network, ip: "192.168.56.2"
 
   config.vm.provider 'virtualbox' do |vb|
     vb.gui = false


Vagrantfile
```diff
@@ -12,7 +12,7 @@ Vagrant.configure('2') do |config|
   config.disksize.size = '50GB'
   config.vm.box_check_update = false
   config.vm.host_name = 'sumologic-kubernetes-collection'
-  config.vm.network :private_network, ip: "192.168.78.66"
+  config.vm.network :private_network, ip: "192.168.56.2"
```
pmalek-sumo, line 15: It won't be as easy as 192.168.SU.MO anymore 😿 

Do you mind adding a comment above this line with e.g. a link to an issue or the note about the limitation?

astencel-sumo: I don't think a comment is needed - `git blame` will tell you the story, right?

pmalek-sumo: I don't usually look at `git blame` output, do you? :) BTW, any change (even a whitespace change) within this line will "stack on top of this change" and it will be even harder to figure out why this was changed to be `... 56.2`

astencel-sumo (approved):
I'm not able to reproduce the issue on my machine, but this change shouldn't hurt. Maybe it's something only reproducible on a clean setup?

Vagrantfile
astencel-sumo: I do 😛

pmalek-sumo (approved):

Vagrantfile
swiatekm-sumo: Added a comment, agreed that it should be there in case someone decides to change the IP in the future, especially since the effects aren't immediately obvious.

PR #1867, opened 2021-11-08
fix(deps): upgrade Fluentd to 1.12.2-sumo-6

astencel-sumo:
See change log at https://github.com/SumoLogic/sumologic-kubernetes-fluentd/releases.

PR #1868, opened 2021-11-08
fix(deps): upgrade Fluentd to 1.12.2-sumo-6

astencel-sumo:
Backports #1867

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -211,7 +211,7 @@ sumologic:
 fluentd:
   image:
     repository: public.ecr.aws/sumologic/kubernetes-fluentd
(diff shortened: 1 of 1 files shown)


pmalek-sumo (approved):
I forgot to ask about this but we can/should add an entry about this into our changelog (which exists only on `main` - which makes sense)

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/CHANGELOG.md

astencel-sumo:
> I forgot to ask about this but we can/should add an entry about this into our changelog (which exists only on `main` - which makes sense)
> 
> https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/CHANGELOG.md

Right, I was going to do it and forgot after all 🤦 Thanks for reminding

PR #1869, opened 2021-11-08
ci(codeowners): use team name only

astencel-sumo:
Backports #1636


pmalek-sumo:
The "problem" with this is that not all the team members will get the notification about a particular PR :(

pmalek-sumo (approved):

PR #1874, opened 2021-11-13
Add integration tests using kind, terratest and kubernetes-sigs/e2e-framework

pmalek-sumo:
Add integration tests using:

- [`kind`][kind]
- [`terratest`][terratest]
- [`kubernetes_e2e_framework`][kubernetes_e2e_framework]

[terratest]: https://github.com/gruntwork-io/terratest
[kubernetes_e2e_framework]: https://github.com/kubernetes-sigs/e2e-framework
[kind]: https://kind.sigs.k8s.io/

---

The idea for now is that:

* every k8s version (as defined per https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/0c1aae5d40346dbc63e175c888c85095f7faa013/tests/integration/kind_images.json) has a separate tests (CI workflow) being run
* each test gets its own cluster (one for each k8s version as defined in `kine_images.json`)


tests/integration/pkg/main_test.go
```diff
@@ -0,0 +1,93 @@
...
+		k := kind.NewCluster(clusterName)
+		// We only provide the config because the API is constructed in such a way
+		// that it requires both the image and the cluster config.
+		kubecfg, err := k.CreateWithConfig(KindImages.Default, "yamls/cluster.yaml")
+		if err != nil {
+			return ctx, err
+		}
+
+		// TODO: find a better way than sleeping here, e.g. to wait for specific
+		// k8s subsystems to be avialable.
```
swiatekm-sumo, line 58: Would doing `kubectl version --request-timeout 3s` work here? Or are we waiting for something specific?

tests/integration/pkg/main_test.go
```diff
@@ -0,0 +1,93 @@
...
+			DestroyKindCluster(kindClusterName),
+		)
+	}
+
+	os.Exit(testenv.Run(m))
+}
+
+type kindContextKey string
+
+func CreateKindCluster(clusterName string) env.Func {
```
swiatekm-sumo, line 47: This is a slightly misleading name, as it actually returns the cluster create function rather than creating one. How about `GetCreateKindClusterFunc`? Same for the destroy one below.

tests/integration/pkg/helm_1_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+)
+
+const (
+	yamlPathReceiverMock = "yamls/receiver-mock.yaml"
```
swiatekm-sumo, line 17: Is this currently used for anything?

tests/integration/pkg/helm_1_test.go
```diff
@@ -0,0 +1,54 @@
...
+			}).
+		Setup(helmDependencyUpdate(_helmSumoLogicChartRelPath)).
+		Setup(helmInstall(HelmSumoLogicChartAbsPath, releaseName)).
+		Teardown(
+			func(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
+				helm.Delete(t, HelmOptions(ctx), releaseName, true)
+				k8s.DeleteNamespace(t, KubectlOptions(ctx), namespaceName)
+				return ctx
+			}).
+		Feature()
```
swiatekm-sumo, line 51: Just so I'm clear, this test just installs the Helm Chart and then uninstalls it without doing anything else, right? So it's a placeholder to show how to write other tests?

swiatekm-sumo (commented):
Looks pretty cool overall, seems to be performant enough to run on every push and equivalent in power to the Vagrant tests. Left some mostly minor comments and questions, as it's my first time seeing terratest and e2e_framework.

tests/integration/pkg/main_test.go
pmalek-sumo: `GetCreateKindClusterFunc` would feel to me like a type name rather than an actual `func`/object. The terminology is the same here as in the original https://github.com/kubernetes-sigs/e2e-framework/blob/9169c8740b064b38d4d1533ae0098473cd22fa1d/pkg/envfuncs/kind_funcs.go#L38.

I needed to add it here in order to store the cluster name I believe

tests/integration/pkg/helm_1_test.go
pmalek-sumo: This can be treated as WIP :)

tests/integration/pkg/helm_1_test.go
pmalek-sumo: Kind of. But the general purpose of this PR was to provide the boilerplate and get things running.

And in a sense this test is valid because it actually checks if the installation succeeds  with (almost) default values.

tests/integration/pkg/main_test.go
pmalek-sumo: We could but we don't have kubectl options readily available here and we'd need to create them (nothing huge but it would need to be done).

+ I'd prefer to add more logic to this "wait" so that we wait e.g. for all subcomponents/daemonsets/deployments(like `coredns`) to be up and running.

tests/integration/pkg/helm_1_test.go
swiatekm-sumo: You want to add anything using it to this PR or merge it as-is?

tests/integration/pkg/main_test.go
swiatekm-sumo: `GetX` feels like a type name to you? :suspect: Either way, this is a minor thing that I can live with.

tests/integration/pkg/main_test.go
swiatekm-sumo: How about: `kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 3s`? That waits until all Pods are ready, which is a pretty good approximation of what we want here, I think. I don't really think we can avoid setting up kubectl here if we want to interact with the cluster meaningfully.

FYI I'm fine merging this as is, I just thought there might be a simple solution.

tests/integration/pkg/main_test.go
pmalek-sumo: `GetCreateKindCluster**Func**` feels like a type name

tests/integration/pkg/helm_1_test.go
pmalek-sumo: I've removed this (I have this being used in a separate test that I've prepared).

.github/workflows/dev_builds.yml
```diff
@@ -13,3 +13,30 @@ jobs:
       - uses: actions/checkout@v2
       - name: Push dev helm chart
         run: make push-helm-chart
+
+  # Add this to dev builds as that's the only way for cache to be shared across branches.
+  # https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
+  integration-tests:
+    runs-on: ubuntu-20.04
+    name: Compile integration tests
```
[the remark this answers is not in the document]
pmalek-sumo: done

swiatekm-sumo (approved):

astencel-sumo:
Would be nice to move all the other contents of the `/tests` directory to a separate subdirectory, what do you think?
image

pmalek-sumo:
> Would be nice to move all the other contents of the `/tests` directory to a separate subdirectory, what do you think? 

Agreed but that's a separate initiative :)

astencel-sumo:
> > Would be nice to move all the other contents of the `/tests` directory to a separate subdirectory, what do you think?
> 
> Agreed but that's a separate initiative :)

This should be a separate commit, sure, but I believe it would be best if this was done before adding the `integration` directory.

astencel-sumo (approved):
I ran `make test` and it didn't fail 🎉 🥇

pmalek-sumo:
@astencel-sumo @swiatekm-sumo Do you mind taking one more look?

I've added:

* running all tests in their separate clusters ( you can see a successful pipeline run in [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/4353938193?check_suite_focus=true)
* some actual assertions
* some helper functions which make the boilerplate a bit easier on the eyes

.github/workflows/pull_requests.yml
```diff
@@ -93,3 +93,64 @@ jobs:
...
+        run: |
+          echo "::set-output name=kind_images::$(jq -cM '.supported' kind_images.json)"
+      - name: Print kind images
+        run: echo "Kind images ${{ steps.set_kind_images.outputs.kind_images }}"
+      - id: set_test_names
+        name: Set test names
+        working-directory: ./tests/integration/
+        # grep magic described in https://unix.stackexchange.com/a/13472
+        run: |
+          echo "::set-output name=test_names::$(grep -rhoP "(?<=^func )(Test[a-zA-z_0-9]+)(?=\(t \*testing.T\) {)" . | jq -R . | jq -cs .)"
```
swiatekm-sumo, line 120: Is there really no cleaner way to do this?

pmalek-sumo: There might be but I don't know about it 🤷 

`go test -list` returns includes the output of the e.g. what `init()` prints out:

```
go test -v -list Test_Helm
2021/11/29 14:38:51 Successfully read kind images spec
2021/11/29 14:38:51 Default kind image: kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047
2021/11/29 14:38:51 Supported kind images: [kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729 kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9 kindest/node:v1.21.2@sha256:9d07ff05e4afefbba983fac311807b3c17a5f36e7061f6cb7e2ba756255b2be4 kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047]
Test_Helm_Default
ok      github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration  0.483s
```

pmalek-sumo: Using JSON doesn't help much:

```
go test -v -list Test_Helm -json
{"Time":"2021-11-29T14:39:54.949815+01:00","Action":"output","Package":"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration","Output":"2021/11/29 14:39:54 Successfully read kind images spec\n"}
{"Time":"2021-11-29T14:39:54.950129+01:00","Action":"output","Package":"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration","Output":"2021/11/29 14:39:54 Default kind image: kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047\n"}
{"Time":"2021-11-29T14:39:54.950139+01:00","Action":"output","Package":"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration","Output":"2021/11/29 14:39:54 Supported kind images: [kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729 kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9 kindest/node:v1.21.2@sha256:9d07ff05e4afefbba983fac311807b3c17a5f36e7061f6cb7e2ba756255b2be4 kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047]\n"}
{"Time":"2021-11-29T14:39:54.950275+01:00","Action":"output","Package":"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration","Output":"Test_Helm_Default\n"}
{"Time":"2021-11-29T14:39:54.950538+01:00","Action":"output","Package":"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration","Output":"ok  \tgithub.com/SumoLogic/sumologic-kubernetes-collection/tests/integration\t0.335s\n"}
{"Time":"2021-11-29T14:39:54.950558+01:00","Action":"pass","Package":"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration","Elapsed":0.335}
```

tests/integration/internal/stepfuncs/step_funcs.go
```diff
@@ -0,0 +1,147 @@
...
+	// controls whether to skip the 'helm dependency update' invocation.
+	// If its set to anything else than an empty string then it's being skipped.
+	envNameHelmNoDependencyUpdate = "HELM_NO_DEPENDENCY_UPDATE"
+)
+
+// HelmDependencyUpdateOpt returns a features.Func that will run helm dependency update using
+// the provided path as an argument.
+//
+// NOTE:
+// It this step will be skipped if the relevant environment variable (envNameHelmNoDependencyUpdate)
```
swiatekm-sumo, line 28: Redundant "It" at the start.

pmalek-sumo: Thanks, done

.github/workflows/pull_requests.yml
swiatekm-sumo: Allright, I guess that's what it'll have to be. Hopefully this doesn't bite us. If it does. we can always consider keeping the test list in a separate file and generating it manually when adding more tests.

swiatekm-sumo (approved):

PR #1878, opened 2021-11-15
Prepare v2.2.0

kkujawa-sumo:
###### Description

Changes: https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v2.1.7...v2.2.0-rc.1

---

###### Testing performed

- [ ] Redeploy fluentd and fluentd-events pods
- [ ] Confirm events, logs, and metrics are coming in

PR #1881, opened 2021-11-16
chore: update changelog

kkujawa-sumo:
###### Description

Changes:
- https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v2.1.7...v2.2.0-rc.1 (unfortunately this view shows also commits which were backported to [release-v2.1](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/release-v2.1))
- https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v2.1.6...v2.1.7

---

###### Testing performed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support otelcol for metrics pipeline [#1702][#1702]
+- Add otelcol as alternative logs metadata provider [#1697][#1697]
+
+  Adds new `values.yaml` property `sumologic.logs.metadata.provider` and associated configuration properties
+  to replace Fluentd with Sumo Logic Distro of OpenTelemetry Collector for logs metadata enrichment.
+
 - otelcol: add systemd logs pipeline [#1767][#1767]
 
   - This change introduces logs metadata enrichment with Sumo Open Telemetry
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -29,17 +35,87 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     In order to set the source name to something else please change
     `otelcol.metadata.logs.config.processors.source/systemd:.source_name`
     configuration value.
+- feat(otelcol/logs/kubelet): add kublet logs pipeline [#1772][#1772]
```
pmalek-sumo, line 38:
```suggestion
- feat(otelcol/logs/kubelet): add kubelet logs pipeline [#1772][#1772]
```

pmalek-sumo (approved):
I believe we can already change `Unreleased` -> `v2.2.0`

BTW we should maintain the changelog for `rc`s (i.e. in the same file)

kkujawa-sumo:
> I believe we can already change `Unreleased` -> `v2.2.0`

I'll change this in next PR: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1878

kkujawa-sumo:
>BTW we should maintain the changelog for rcs (i.e. in the same file)

What changes should be included in the changelog for `rc`s? 
For `2.2.0-rc.0` changes from `2.1.7 ` to `2.2.0-rc.0` and for `2.2.0.rc.1` changes from `2.2.0-rc.0` to `2.2.0.rc.1`?
Should then `2.2.0` include changes only from `2.2.0.rc.1` to `2.2.0` or everything from `2.1.7 ` to `2.2.0`?

Could you propose form of such changelog or send the link to example?

pmalek-sumo:
> > BTW we should maintain the changelog for rcs (i.e. in the same file)
> 
> What changes should be included in the changelog for `rc`s? For `2.2.0-rc.0` changes from `2.1.7 ` to `2.2.0-rc.0` and for `2.2.0.rc.1` changes from `2.2.0-rc.0` to `2.2.0.rc.1`? Should then `2.2.0` include changes only from `2.2.0.rc.1` to `2.2.0` or everything from `2.1.7 ` to `2.2.0`?
> 
> Could you propose form of such changelog or send the link to example?

```
# 2.2.0

Changes

# 2.2.0-rc.1

Changes

# 2.2.0-rc.0

Changes

# 2.1.7

Changes
```

but that would contain most of the changes in `rc.0` 🤔 ... maybe we'd have to think more about this. Maybe put everything into release/Unreleased(before the release happens) and in between the `rc`s corresponding changes?

perk-sumo (approved):

perk-sumo:
Yeah, that's not an easy problem. Here's [what Syncthing does](https://github.com/syncthing/syncthing/releases) - basically RC's have mostly same info as GA.

Kubernetes on the other hand [keeps everything separate](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md) but also they have changelog file per release.

pmalek-sumo:
> Yeah, that's not an easy problem. Here's [what Syncthing does](https://github.com/syncthing/syncthing/releases) - basically RC's have mostly same info as GA.
> 
> Kubernetes on the other hand [keeps everything separate](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md) but also they have changelog file per release.

Nice. I believe we can use the syncthing's approach since our changelog tend to (well despite 2.2) be small. (or rather smaller than kubernetes') And we can indicate that the changes are `FROM_PREVIOUS_RELEASE..TO_THIS`

PR #1890, opened 2021-11-19
docs: remove unsupported k8s versions

astencel-sumo:


swiatekm-sumo (approved):

perk-sumo (approved):

pmalek-sumo (commented):
I suggest we use helm capabilities to prevent users installing in 1.16 and below: https://helm.sh/docs/chart_template_guide/builtin_objects/

PR #1895, opened 2021-11-22
feat: add affinity to fluentd events statefulset

### LINKED ISSUE #1880: Add `affinity` sections to the `fluentd` templates
**Is your feature request related to a problem? Please describe.**
We want to control which node pools the collection's pods run in. `nodeSelector` isn't flexible enough for our needs as we don't have a set of labels that could be used as an AND to define a nodeSelector. Also, we can't use `antiAffinity` to space out replicas across nodes. 

**Describe the solution you'd like**
Add `affinity` sections to the `values.yaml` for each of the templates creating Deployments in the chart, and then pull those through to the templates.

**Describe alternatives you've considered**
Applying more labels so that we can do a 'don't run on nodes with this label' by applying another label to every other node pool would work, but is a lot of overhead for us to manage.

pmalek-sumo:
Fixes #1880


astencel-sumo (approved):
Fixes #1880 if I'm not mistaken @pmalek-sumo?

PR #1906, opened 2021-11-26
chore: remove support for EOL K8s platform versions

swiatekm-sumo:
###### Description

1.17 on EKS and 1.18 on GKE don't allow new clusters to be created; 1.19 on GKE shouldn't either according to GCP's schedule. As we can't test on these platforms, we can't officially support them.

See https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-release-calendar and https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule_for_static_no-channel_versions.

---

###### Testing performed

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -82,15 +82,15 @@ The diagram below illustrates the components of the Kubernetes collection soluti
 
 The following table displays the tested Kubernetes and Helm versions.
 
-| Name          | Version                                  |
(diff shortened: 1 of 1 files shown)


pmalek-sumo (approved):

perk-sumo (approved):
LGTM, thanks!

PR #1916, opened 2021-11-30
chore: filter out receiver-mock's logs in fluent-bit's config

pmalek-sumo:
This has already been done in fluentd:

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/f4d95bc71b9974f32bab700bb79c1725f111ee6d/vagrant/values.yaml#L159-L166

but when using otelcol as logs provider it doesn't make an effect hence this PR.


swiatekm-sumo (approved):
Nice catch!

PR #1922, opened 2021-12-01
Add changelog update item to PR checklist

swiatekm-sumo:
###### Description

I'm annoyed at always forgetting to update the changelog, so here's a small reminder. Ideally, we'd have tooling to check this in the CI, but this is easy to do and should help a fair bit.


pmalek-sumo (approved):
+1 on the tooling (albeit it won't be easy to do reliably)

PR #1923, opened 2021-12-01
tests: add otc metadata integration tests

pmalek-sumo:
Add OTC related integration metadata integration tests.

* use the following [`values.yaml`](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/add-sumo-ot-distro-metadata-integration-tests/tests/integration/values/values_otc_metadata.yaml)
* add some [assertions](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/add-sumo-ot-distro-metadata-integration-tests/tests/integration/helm_otc_metadata_installation_test.go)

### DIFF
--- a/tests/integration/helm_otc_metadata_installation_test.go
+++ b/tests/integration/helm_otc_metadata_installation_test.go
@@ -21,12 +21,12 @@ import (
 	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/stepfuncs"
 )
 
-func Test_Helm_Default(t *testing.T) {
+func Test_Helm_OT_Metadata(t *testing.T) {
 	var (
 		now            = time.Now()
 		namespace      = generateNamespaceName(now)
 		releaseName    = generateReleaseName(now)
-		valuesFilePath = "values/values_default.yaml"
+		valuesFilePath = "values/values_otc_metadata.yaml"
 
 		tickDuration = time.Second
 		waitDuration = time.Minute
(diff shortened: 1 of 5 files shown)


tests/integration/helm_otc_metadata_installation_test.go
```diff
@@ -0,0 +1,182 @@
...
+			func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+				require.Eventually(t, func() bool {
+					pods := k8s.ListPods(t, ctxopts.KubectlOptions(ctx), v1.ListOptions{
+						LabelSelector: fmt.Sprintf("app=%s-sumologic-otelcol-logs", releaseName),
+						FieldSelector: "status.phase=Running",
+					})
+
+					if len(pods) != 3 {
+						return false
+					}
```
swiatekm-sumo, line 65: I think we should move the magic numbers and strings, like the pod count or the app labels, into the variable definitions up top. In all honesty, we should either read these from the generated yaml or explicitly template them into the values.yaml we use for tests, but making them more explicit should be good enough for now.

This applies to all the labels and pod counts in this test.

swiatekm-sumo (approved):
I'd like the labels we use for selecting pods for different apps, and the counts where they're different than 1, organised a bit better in the test case. Other than that, looks great to me.

tests/integration/helm_otc_metadata_installation_test.go
pmalek-sumo: I get the idea and we can perhaps reorganize/templatize this at some point but one thing I wouldn't change which is to not read the (default) values from the yaml files but to assert on them (expecting what's the actual default) if that makes sense.

PR #1865, opened 2021-11-03
feat(helm): add PodDisruptionBudget api version helm chart helpers

pmalek-sumo:
The idea behind this change is to support newer version of `PodDisruptionBudget` (`policy/v1beta1/PodDisruptionBudget` vs `policy/v1/PodDisruptionBudget`)
and to get rid of deprecation warnings when installing our chart:

```
$ helm upgrade --install collection sumologic/sumologic
Release "collection" does not exist. Installing it now.
...
W1118 17:09:13.651675 2835438 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W1118 17:09:13.654276 2835438 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
...
W1118 17:09:14.026749 2835438 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W1118 17:09:14.028299 2835438 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
NAME: collection
LAST DEPLOYED: Thu Nov 18 17:09:12 2021
NAMESPACE: sumologic
STATUS: deployed
REVISION: 1
NOTES:
Thank you for installing sumologic.

PR #1924, opened 2021-12-01
fix(helm): fix helm template render tests

pmalek-sumo:
Helm template render tests where broken in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1865 (by introduction of a subshell in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1865/files#diff-dd0c9cda3917fccebea6cea8d3fb46f46e4fbbeda2d6939864c9aaa9fc9c9571R29-R39)

Hence they were not reporting issues which they should report: e.g. https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/4371102809?check_suite_focus=true#step:4:75

This PR fixes that by using a temporary file to report test success/failures https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1924/files#diff-a13b01bb0c50ed17f62903dbcbb0a1df693e4e0a921eefc8dd078fff56bc57f3R18 and by fixing already accumulated breakage in templates.

### DIFF
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -965,8 +965,8 @@ Example:
 {{- define "kubernetes.sources.envs" -}}
 {{- $ctx := .Context -}}
 {{- $type := .Type -}}
-{{- range $name, $source := (index .Context.sumologic.collector.sources $type) }}
-{{ include "kubernetes.sources.env" (dict "Context" $ctx "Type" $type  "Name" $name ) }}
+{{- range $name, $source := (index .Context.sumologic.collector.sources $type) -}}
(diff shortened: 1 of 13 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -248,6 +248,8 @@ fluentd:
   securityContext:
     ## The group ID of all processes in the statefulset containers. By default this needs to be fluent(999).
     fsGroup: 999
+    runAsUser: 0
+    runAsGroup: 0
```
swiatekm-sumo, line 252: Why are we making this change?

pmalek-sumo: This is a test 1,2,3.
Thanks for staying vigilant! 🤦

swiatekm-sumo (approved):

PR #1932, opened 2021-12-06
Fix pod readiness checks in integration tests

swiatekm-sumo:
##### Description

Use utilities from terratest to check pod availability instead of rolling our own. Our tests would actually pass despite some containers getting OOM killed.

The reason our containers were getting OOM killed were rather strict resource limits set for them in values.yaml we use for tests. Remove those limits, they don't serve any real purpose right now.

### DIFF
--- a/tests/integration/helm_default_installation_test.go
+++ b/tests/integration/helm_default_installation_test.go
@@ -30,7 +30,7 @@ func Test_Helm_Default(t *testing.T) {
 		valuesFilePath = "values/values_default.yaml"
 
 		tickDuration = time.Second
-		waitDuration = time.Minute
+		waitDuration = time.Minute * 2
 	)
 
 	feat := features.New("installation").
--- a/tests/integration/helm_default_installation_test.go
+++ b/tests/integration/helm_default_installation_test.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal"
 	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/ctxopts"
+	k8sinternal "github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/k8s"
 	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/stepfuncs"
 )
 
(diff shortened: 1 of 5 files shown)


tests/integration/helm_default_installation_test.go
```diff
@@ -52,19 +53,15 @@ func Test_Helm_Default(t *testing.T) {
...
-						LabelSelector: fmt.Sprintf("app=%s-sumologic-fluentd-logs", releaseName),
-						FieldSelector: "status.phase=Running",
-					})
-					return len(pods) == 3
-				}, waitDuration, tickDuration)
-
+				filters := v1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s-sumologic-fluentd-logs", releaseName),
+				}
+				utils.WaitUntilPodsAvailable(t, ctxopts.KubectlOptions(ctx), filters, 3, waitDuration, tickDuration)
```
pmalek-sumo, line 61: Plz not `utils` 🙏

swiatekm-sumo: Come up with a better name then? `k8s` like in terratest?

swiatekm-sumo: Changed to `waituntil`

tests/integration/internal/k8s/pods.go
```diff
@@ -0,0 +1,67 @@
...
+	minPods int,
+	sleepDuration time.Duration,
+	sleepBetweenRetries time.Duration,
+) {
+	require.NoError(t, WaitUntilPodsAvailableE(t, options, filters, minPods, sleepDuration, sleepBetweenRetries))
+}
+
+// WaitUntilPodsAvailableE waits until pods satisfying the provided filters are all available.
+// Availability is defined via terrastruct's k8s.IsPodAvailable function, which requires the Pod
+// to be Running and all of its containers to be started and ready.
```
pmalek-sumo, line 28: > ...to be started and ready

It's not technically true, right? We're not waiting for readiness probes here, are we?

swiatekm-sumo: We're waiting for containers to be Ready, which does imply probes afaik:
https://github.com/gruntwork-io/terratest/blob/3d69d62aab234c3eb39adcac1aaff09dd8840697/modules/k8s/pod.go#L140

pmalek-sumo (approved):

PR #1933, opened 2021-12-06
Add an integration test for presence of expected metrics

swiatekm-sumo:
##### Description

Add tests for the presence of metrics we expect to send to the receiver. Uses receiver-mock's `/metrics-list` endpoint. 

Adding it to the existing Features for now, as I don't want to duplicate setup in a new Feature, and it's not yet clear how we want to deal with this structurally.

### DIFF
--- a/tests/integration/helm_otc_metadata_installation_test.go
+++ b/tests/integration/helm_otc_metadata_installation_test.go
@@ -28,8 +28,9 @@ func Test_Helm_OT_Metadata(t *testing.T) {
 		releaseName    = generateReleaseName(now)
 		valuesFilePath = "values/values_otc_metadata.yaml"
 
-		tickDuration = time.Second
-		waitDuration = time.Minute * 2
+		tickDuration    = time.Second
+		waitDuration    = time.Minute * 2
+		expectedMetrics = internal.DefaultExpectedMetrics
 	)
 
 	feat := features.New("installation").
@@ -145,7 +146,17 @@ func Test_Helm_OT_Metadata(t *testing.T) {
 
 				require.EqualValues(t, 0, daemonsets[0].Status.NumberUnavailable)
 				return ctx
-			}).
+		}).
+		Assess("metrics are present", // TODO: extract this out to a separate feature
+			stepfuncs.WaitUntilExpectedMetricsPresent(
+				expectedMetrics,
+				"receiver-mock",
+				"receiver-mock",
+				internal.ReceiverMockServicePort,
+				waitDuration,
+				tickDuration,
+			),
+		).
 		Feature()
 
 	testenv.Test(t, feat)
--- a/tests/integration/internal/constants.go
+++ b/tests/integration/internal/constants.go
@@ -13,12 +13,82 @@ const (
 
 	EnvNameKindImage = "KIND_NODE_IMAGE"
 
-	YamlPathReceiverMock = "yamls/receiver-mock.yaml"
+	YamlPathReceiverMock    = "yamls/receiver-mock.yaml"
+	ReceiverMockServicePort = 3000
+)
+
+// metrics we expect the receiver to get
+// metrics which are expected in current E2E tests but aren't present here, are commented out
+// TODO: figure out why the expected metrics aren't present
+var (
+	KubeStateMetrics = []string{
+		"kube_statefulset_status_observed_generation",
+		"kube_statefulset_status_replicas",
+		"kube_statefulset_replicas",
+		"kube_statefulset_metadata_generation",
+	}
+	KubeDaemonSetMetrics = []string{
+		"kube_daemonset_status_current_number_scheduled",
+		"kube_daemonset_status_desired_number_scheduled",
+		"kube_daemonset_status_number_misscheduled",
+		"kube_daemonset_status_number_unavailable",
+	}
+	KubeDeploymentMetrics = []string{
+		"kube_deployment_status_replicas_available",
+		"kube_deployment_status_replicas_unavailable",
+		"kube_deployment_spec_replicas",
+	}
+	KubeNodeMetrics = []string{
+		"kube_node_info",
+		"kube_node_status_allocatable",
+		"kube_node_status_capacity",
+		"kube_node_status_condition",
+	}
+	KubePodMetrics = []string{
+		"kube_pod_container_info",
+		"kube_pod_container_resource_requests",
+		"kube_pod_container_resource_limits",
+		"kube_pod_container_status_ready",
+		"kube_pod_container_status_terminated_reason",
+		"kube_pod_container_status_restarts_total",
+		"kube_pod_status_phase",
+	}
+	KubeletMetrics = []string{
+		"kubelet_running_containers",
+		"kubelet_running_pods",
+	}
+	CAdvisorMetrics = []string{
+		"container_cpu_usage_seconds_total",
+		// These metrics will be avaiable in containerd after kind upgrades past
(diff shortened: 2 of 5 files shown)


tests/integration/helm_otc_metadata_installation_test.go
```diff
@@ -139,6 +141,30 @@ func Test_Helm_OT_Metadata(t *testing.T) {
 				require.EqualValues(t, 0, daemonsets[0].Status.NumberUnavailable)
 				return ctx
 			}).
+		Assess("metrics are present",
+			func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+				kubectlOpts := *ctxopts.KubectlOptions(ctx)
+				kubectlOpts.Namespace = "receiver-mock"
+				k8s.WaitUntilServiceAvailable(t, &kubectlOpts, "receiver-mock", int(waitDuration), tickDuration)
+				baseUrl := fmt.Sprintf("http://%s:%d", internal.NodePortHostname, internal.ReceiverMockNodePort)
```
pmalek-sumo, line 149: Maybe we could think about using tunnel: https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/tunnel_test.go

swiatekm-sumo: Huh I thought that was a lot more complex for some reason. Let me try this way.

tests/integration/internal/constants.go
```diff
@@ -13,12 +13,80 @@ const (
...
+		"kube_pod_status_phase",
+	}
+	KubeletMetrics = []string{
+		"kubelet_running_containers",
+		"kubelet_running_pods",
+	}
+	CAdvisorMetrics = []string{
+		"container_cpu_usage_seconds_total",
+		// "container_fs_usage_bytes",
+		// "container_fs_limit_bytes",
```
pmalek-sumo, line 63: I believe this one was is related to https://github.com/google/cadvisor/pull/2872

tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+
+	http_helper "github.com/gruntwork-io/terratest/modules/http-helper"
+)
+
+// A HTTP client for the receiver-mock API
+type ReceiverMockClient struct {
+	baseUrl   url.URL
+	tlsConfig tls.Config
+	t         *testing.T
```
pmalek-sumo, line 18: I'd prefer not to store this. Instead we could pass it as first param to exposed public functions (like in other "frameworks")

tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -0,0 +1,70 @@
...
+	metricNameToCount := make(map[string]int)
+	lines := strings.Split(rawMetricsValues, "\n")
+	for _, line := range lines {
+		if len(line) == 0 {
+			continue
+		}
+		// the last colon of the line is the split point
+		splitIndex := strings.LastIndex(line, ":")
+		if splitIndex == -1 || splitIndex == 0 {
+			return nil, fmt.Errorf("failed to parse metrics list line: '%s'", line)
```
pmalek-sumo, line 59:
```suggestion
			return nil, fmt.Errorf("failed to parse metrics list line: %q", line)
```

equivalent

tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -0,0 +1,70 @@
...
+	baseUrl   url.URL
+	tlsConfig tls.Config
+	t         *testing.T
+}
+
+func NewReceiverMockClient(t *testing.T, baseUrl url.URL) *ReceiverMockClient {
+	return &ReceiverMockClient{baseUrl: baseUrl, t: t, tlsConfig: tls.Config{}}
+}
+
+func (client *ReceiverMockClient) GetMetricCounts() (map[string]int, error) {
```
pmalek-sumo, line 25: Might want to define a type for this return type

```go
type MetricCounts map[string]int
```

tests/integration/internal/stepfuncs/assess_funcs.go
```diff
@@ -2,25 +2,78 @@ package stepfuncs
...
+		require.Eventually(t, func() bool {
+			metricCounts, err := client.GetMetricCounts()
+			if err != nil {
+				t.Log(err)
+				return false
+			}
+			for _, expectedMetricName := range expectedMetrics {
+				_, ok := metricCounts[expectedMetricName]
+				if !ok {
+					t.Logf("Couldn't find metric '%s' in %v", expectedMetricName, metricCounts)
```
pmalek-sumo, line 71:
```suggestion
					t.Logf("Couldn't find metric %q in %v", expectedMetricName, metricCounts)
```

tests/integration/internal/stepfuncs/assess_funcs.go
```diff
@@ -2,25 +2,78 @@ package stepfuncs
...
+					return false
+				}
+			}
+			return true
+		}, waitDuration, tickDuration*5)
```
pmalek-sumo, line 76: `tickDuration*5` seems a bit magic to me.

tests/integration/helm_otc_metadata_installation_test.go
```diff
@@ -146,6 +147,16 @@ func Test_Helm_OT_Metadata(t *testing.T) {
 				require.EqualValues(t, 0, daemonsets[0].Status.NumberUnavailable)
 				return ctx
 			}).
+		Assess("metrics are present",
```
pmalek-sumo, line 150: Let's add a TODO to extract this out to a separate feature when we figure out sharing a cluster between features.

pmalek-sumo (changes requested):

tests/integration/internal/constants.go
swiatekm-sumo: So it'll start working after kind gets to containerd 1.5.7: https://github.com/containerd/containerd/issues/5882? I'll add a comment.

tests/integration/internal/stepfuncs/assess_funcs.go
swiatekm-sumo: Yeah, that's a debugging holdover that shouldn't be here. Good catch!

tests/integration/helm_otc_metadata_installation_test.go
```diff
@@ -145,7 +146,17 @@ func Test_Helm_OT_Metadata(t *testing.T) {
 
 				require.EqualValues(t, 0, daemonsets[0].Status.NumberUnavailable)
 				return ctx
-			}).
+						}).
```
pmalek-sumo, line 149: This indentation looks weird

swiatekm-sumo: Weird, I swear go fmt did this...

tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"net/url"
+	"strconv"
+	"strings"
+	"testing"
+
+	http_helper "github.com/gruntwork-io/terratest/modules/http-helper"
+)
+
+type MetricCounts map[string]int
```
pmalek-sumo, line 14: Add a comment on this type what does it represent: I assume it's a mapping of metric names to metric counts but who knows :)

pmalek-sumo (approved):
Apart from one nit 👍

PR #1936, opened 2021-12-06
feat: add option to disable `service` enrichment

astencel-sumo:


swiatekm-sumo (approved):

perk-sumo (approved):

pmalek-sumo (changes requested):
Changelog 🤡

deploy/helm/sumologic/values.yaml
```diff
@@ -220,7 +220,7 @@ sumologic:
 fluentd:
   image:
     repository: public.ecr.aws/sumologic/kubernetes-fluentd
-    tag: 1.12.2-sumo-10
+    tag: 1.12.2-sumo-11
```
pmalek-sumo, line 223: This was caught by the integration test https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/4431276057?check_suite_focus=true#step:6:334

😲 🚀 
image

perk-sumo: I believe they just build longer than they used to - so while it's 🎉 that it was caught - there is no real error, we just need for images to be published (which is happening, just slow).
Ref https://github.com/SumoLogic/sumologic-kubernetes-fluentd/actions/runs/1544893588

pmalek-sumo: Doesn't matter if it's fast or slow. At the moment you merge this PR this image has to be available. No tricks, no postponing, no rushing things. When it's available then we can merge it.

astencel-sumo: Here's something to shorten the Fluentd release time: https://github.com/SumoLogic/sumologic-kubernetes-fluentd/pull/371

swiatekm-sumo: I doubt publishing changes will make a difference. Most of the time is spent building the image for ARM, which tends to be very slow in QEMU. If we want it to be faster, we can add a build cache and/or try not to build native extensions during the image build process, which is what takes the lion's share of the time

pmalek-sumo (approved):

deploy/helm/sumologic/values.yaml
astencel-sumo: There's a recent blog from Docker on this very topic https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/, but I'm afraid it might not apply for our Ruby image build.

swiatekm-sumo: Yeah, cross-compiling ruby extensions is probably not the easiest way of improving this process. I bet we could get quite far by configuring caching correctly for buildx, which isn't very difficult, but ECR doesn't support it yet.

PR #1951, opened 2021-12-08
docs: comment change in values.yaml - tracing is not experimental anymore

perk-sumo:
##### Description

Comment change in values.yaml - tracing is not experimental anymore

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -211,7 +211,7 @@ sumologic:
       provider: fluentd
 
   ### Traces configuration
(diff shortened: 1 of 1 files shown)


pmalek-sumo (commented):
Description please 🙏

pmm-sumo:
Following the thread, we should also update default value

perk-sumo:
> Following the thread, we should also update default value

Ok, that makes sense but we cannot backport that (it's not a patch) so it needs another PR.

astencel-sumo (approved):

PR #1968, opened 2021-12-11
Minor integration test fixes

swiatekm-sumo:
##### Description

* Only print the list of expected metrics at the end of the test, reducing the number of large log lines in integration tests.
* Change the name of the Helm default test so it's not a prefix of other test names. Due to how `go test` matches test runs, this caused us to run more tests than necessary when picking the default test.
* Remove an unnecessary string test. We should think about running tests for our test framework separately, so as to not spawn too many runs in GHA.


tests/integration/internal/strings/strings_test.go
```diff
@@ -6,10 +6,6 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_ValuesFileFromT(t *testing.T) {
```
pmalek-sumo, line 9: I'd leave this because this shows the reader that both this and the test below return the same values file name.

pmalek-sumo (approved):

tests/integration/internal/strings/strings_test.go
swiatekm-sumo: Ok, restored it and added a comment explaining it instead.

PR #1975, opened 2021-12-13
Fix Otel Agent Template

mat-rumian:
##### Description

Fill in your description here.

---

##### Checklist

Remove items which don't apply to your PR.

- [ ] Changelog updated

###### Testing performed

Tested on:
```
sumologic:
  accessId: "dummy"
  accessKey: "dummy"
  endpoint: http://receiver-mock.receiver-mock:3000/terraform/api/
  clusterName: "cluster-1"
  traces:
    enabled: true

otelagent:
  enabled: true
```

PR #1977, opened 2021-12-14
test(traces): add integration test for enabled traces

pmalek-sumo:
Related to recent problem fixed in: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1975

This PR adds a test with traces enabled to verify that traces work with the chart by actually installing it in a cluster.

Depends on https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1971

### DIFF
--- a/tests/integration/values/values_helm_traces_enabled.yaml
+++ b/tests/integration/values/values_helm_traces_enabled.yaml
@@ -1,0 +1,68 @@
+sumologic:
+  setupEnabled: true
+  accessId: "dummy"
+  accessKey: "dummy"
+  endpoint: http://receiver-mock.receiver-mock:3000/terraform/api/
+
+  traces:
+    enabled: true
+
+# Prevent snowball effect by filtering out receiver mock logs
+fluent-bit:
+  config:
+    filters: |
+      [FILTER]
+          Name    grep
+          Match   containers.var.log.containers.receiver-mock*
+          Exclude log .*
+
+# Request less resources so that this fits on Github actions runners environment
+kube-prometheus-stack:
+  prometheus:
+    prometheusSpec:
+      resources:
+        requests:
+          cpu: 100m
+          memory: 128Mi
+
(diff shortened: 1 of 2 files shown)


mat-rumian:
Great work! Really like the tests :rocket:

mat-rumian (approved):

tests/integration/values/values_helm_traces_enabled.yaml
```diff
@@ -0,0 +1,68 @@
...
+  setupEnabled: true
+  accessId: "dummy"
+  accessKey: "dummy"
+  endpoint: http://receiver-mock.receiver-mock:3000/terraform/api/
+
+  traces:
+    enabled: true
+
+# Prevent snowball effect by filtering out receiver mock logs
+fluent-bit:
```
perk-sumo, line 11: Why do we set this (and others like Prometheus or Fluentd) for tracing tests?

perk-sumo: I think we should include `sumologic.traces.enabled: true` and the `otelcol` and `otelagent` sections only 🤔

pmalek-sumo: This is to prevent fluent-bit ingesting receiver-mock logs, which would then be printed out on stdout and be ingested by fluent-bit and so on.

We could either keep this filter or disable logs for this particular testcase.

I'm open to suggestions but I thought that we should check that everything seems to work fine when we change a different config option like `traces.enabled`.

pmalek-sumo: > Why do we set this (and others like Prometheus or Fluentd) for tracing tests?

There are comments explaining why we need them :) TLDR, with default requests those pods wouldn't fit on the github runner's node

perk (approved):

perk-sumo (approved):
🏅

PR #1986, opened 2021-12-16
feat: add experimental otelcol log collector

swiatekm-sumo:
##### Description

This is an alpha feature without a stable API. It only supports container logs from docker-shim. Fluent-bit needs to be manually disabled to avoid duplicate logs.

The namespacing in the internal configuration is currently a bit messy -  we have a top-level `sumologic.logs` key, which should probably be `sumologic.metadata.logs`, same with metrics. Same thing is true about the directory structure within the Helm chart. I'd like to change both of those in a separate PR after merging this one. Then this collector can live under `sumologic.logs.*` instead of `sumologic.logs.collector.*`.

The configuration for otelcol was originally prepared by @astencel-sumo, he gets credit for it actually doing the right thing.

---

##### Checklist

- [x] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/templates/logs/collector/otelcol/configmap.yaml
+++ b/deploy/helm/sumologic/templates/logs/collector/otelcol/configmap.yaml
@@ -1,0 +1,11 @@
+{{- if .Values.otellogs.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "sumologic.logs.collector.name.configmap" . }}
+  labels:
+    app: {{ template "sumologic.labels.app.logs.collector.configmap" . }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+data:
+  {{- (tpl (.Files.Glob "conf/logs/collector/otelcol/config.yaml").AsConfig .) | nindent 2 }}
+{{- end }}
--- a/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
+++ b/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
@@ -1,0 +1,109 @@
+{{- if .Values.otellogs.enabled }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ template "sumologic.logs.collector.name.daemonset" . }}
+  labels:
+    app: {{ template "sumologic.labels.app.logs.collector.daemonset" . }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "sumologic.labels.app.logs.collector.pod" . }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/logs/collector/otelcol/configmap.yaml") . | sha256sum }}
+{{- if .Values.sumologic.podAnnotations }}
+{{ toYaml .Values.sumologic.podAnnotations | indent 8 }}
+{{- end }}
+{{- if .Values.otellogs.daemonset.podAnnotations }}
+{{ toYaml .Values.otellogs.daemonset.podAnnotations | indent 8 }}
+{{- end }}
+      labels:
+        app.kubernetes.io/name: {{ template "sumologic.labels.app.logs.collector.pod" . }}
+        {{- include "sumologic.labels.common" . | nindent 8 }}
+{{- if .Values.sumologic.podLabels }}
+{{ toYaml .Values.sumologic.podLabels | indent 8 }}
+{{- end }}
+{{- if .Values.otellogs.daemonset.podLabels }}
+{{ toYaml .Values.otellogs.daemonset.podLabels | indent 8 }}
+{{- end }}
+    spec:
+      securityContext:
+        {{- toYaml .Values.otellogs.daemonset.securityContext | nindent 8 }}
+      {{- if .Values.otellogs.daemonset.priorityClassName }}
+      priorityClassName: {{ .Values.metadata.logs.daemonset.priorityClassName | quote }}
+      {{- end }}
+      containers:
+      - args:
+        - --config=/etc/otelcol/config.yaml
+        image: {{ .Values.otellogs.image.repository }}:{{ .Values.otellogs.image.tag }}
+        imagePullPolicy: {{ .Values.otellogs.image.pullPolicy }}
+        name: otelcol
+        livenessProbe:
+          httpGet:
+            path: /
+            port: 13133 # Health Check extension default port.
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 13133 # Health Check extension default port.
+        volumeMounts:
+        - mountPath: /etc/otelcol
+          name: otelcol-config
+        - mountPath: /var/log/pods
+          name: varlogpods
+          readOnly: true
+        - mountPath: /var/lib/docker/containers
+          name: varlibdockercontainers
+          readOnly: true
+        - mountPath: {{ .Values.otellogs.config.extensions.file_storage.directory }}
+          name: file-storage
+        env:
+        - name: LOGS_METADATA_SVC
+          valueFrom:
+            configMapKeyRef:
+              name: sumologic-configmap
+              key: fluentdLogs
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        securityContext: 
+          {{- toYaml .Values.otellogs.daemonset.containers.otelcol.securityContext | nindent 10 }}
+      initContainers: # ensure the host path is owned by the otel user group
+      - name: changeowner
+        image: busybox
+        command:
+        - "sh"
+        - "-c"
+        - |
+          chown -R \
+            {{ .Values.otellogs.daemonset.securityContext.fsGroup }}:{{ .Values.otellogs.daemonset.securityContext.fsGroup }} \
+            {{ .Values.otellogs.config.extensions.file_storage.directory }}
+        volumeMounts:
+        - mountPath: {{ .Values.otellogs.config.extensions.file_storage.directory }}
+          name: file-storage
+      volumes:
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: config.yaml
+            path: config.yaml
+          name: {{ template "sumologic.logs.collector.name.configmap" . }}
+        name: otelcol-config
+      - hostPath:
+          path: /var/log/pods
+          type: ""
+        name: varlogpods
+      - hostPath:
+          path: /var/lib/docker/containers
+          type: ""
+        name: varlibdockercontainers
+      - hostPath:
+          path: /var/lib/otc
+          type: DirectoryOrCreate
+        name: file-storage
+      serviceAccountName: {{ template "sumologic.logs.collector.name.serviceaccount" . }}
+{{- end }}
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -4023,7 +4023,7 @@ metadata:
           - file_storage
           # - sumologic
         pipelines:
-          logs:
+          logs/fluent/containers:
             receivers:
               - fluentforward
             processors:
@@ -4038,7 +4038,7 @@ metadata:
               - batch
             exporters:
               - sumologic/containers
-          logs/systemd:
+          logs/fluent/systemd:
             receivers:
               - fluentforward
             processors:
@@ -4052,7 +4052,7 @@ metadata:
               - batch
             exporters:
               - sumologic/systemd
-          logs/kubelet:
+          logs/fluent/kubelet:
             receivers:
               - fluentforward
             processors:
@@ -4065,7 +4065,7 @@ metadata:
               - batch
             exporters:
               - sumologic/systemd
-          logs/otc/containers:
+          logs/otlp/containers:
             receivers:
               - otlp
             processors:
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -3776,6 +3776,10 @@ metadata:
       receivers:
         fluentforward:
           endpoint: 0.0.0.0:24321
+        otlp:
+          protocols:
+            http:
+              endpoint: 0.0.0.0:4318
       extensions:
         health_check: {}
         ## Configuration for File Storage extension
@@ -4061,7 +4065,19 @@ metadata:
               - batch
             exporters:
               - sumologic/systemd
-
+          logs/otc/containers:
+            receivers:
+              - otlp
+            processors:
+              - memory_limiter
+              - attributes/containers
+              - groupbyattrs/containers
+              - k8s_tagger
+              - source/containers
+              - resource/containers_copy_node_to_host
+              - batch
+            exporters:
+              - sumologic/containers
     statefulset:
       nodeSelector: {}
       tolerations: []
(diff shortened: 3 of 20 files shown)


perk-sumo:
> I'd like to change both of those in a separate PR after merging this one.

Please create an issue for v3 as that will be a breaking change. Also this most probably will be changed globally for all the keys.

deploy/helm/sumologic/values.yaml
```diff
@@ -4125,6 +4129,212 @@ metadata:
       ## To use maxUnavailable, set minAvailable to null and uncomment the below:
       # maxUnavailable: 1
 
+## Configure log collection with Otelcol
+## This is currently highly experimental and not recommended unless you really know what you're doing
+## In particular, this configuration currently only covers container logs, and only supports docker-shim
+## This is an alpha feature, and may change in the near future
+logs:
+  enabled: false
```
perk-sumo, line 4137: Now that I look at this key it seems to me it will be too misleading. Let's change the whole key from `logs` into `otelcollogs` (for lack of better ideas) for the time being 🙏

pmalek-sumo: I'd argue not to to something "just for now".

It will be a pain and unnecessary effort later on to migrate/change this

deploy/helm/sumologic/values.yaml
```diff
@@ -4022,9 +4026,9 @@ metadata:
           logs:
             receivers:
               - fluentforward
+              - otlp
             processors:
               - memory_limiter
-              - filter/include_fluent_tag_containers
```
perk-sumo, line 4027: Will this still work for logs sent by `Fluent Bit`?

swiatekm-sumo:
> > I'd like to change both of those in a separate PR after merging this one.
> 
> Please create an issue for v3 as that will be a breaking change. Also this most probably will be changed globally for many other keys.

I meant the internal keys defined in `_helpers.tpl`, which aren't a breaking change and are simpler structurally (since they only cover resources explicitly defined in our chart).

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Yes. This filter only lets through logs from fluent-bit, actually. I'm not entirely sure why it's here, given that we should probably process any record that we get here. @kkujawa-sumo do you know what the purpose was originally?

deploy/helm/sumologic/values.yaml
swiatekm-sumo: It is an alpha feature as is, so we can change it without breaking changes. it'd be nice to have a clear vision of how we want this to look in 3.0, though.

deploy/helm/sumologic/values.yaml
```diff
@@ -4037,9 +4041,9 @@ metadata:
           logs/systemd:
             receivers:
               - fluentforward
+              - otlp
```
pmalek-sumo, line 4044: Let's not touch systemd/kubelet related pipelines in this PR.

deploy/helm/sumologic/values.yaml
```diff
@@ -4022,9 +4026,9 @@ metadata:
           logs:
             receivers:
               - fluentforward
+              - otlp
```
pmalek-sumo, line 4029: Let's create a separate pipeline for this as discussed elsewhere.

Also please rename this pipeline to `logs/containers` (and accordingly the other one can be called `logs/containers_otlp`)

pmalek-sumo (changes requested):

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Update: The filter is here to control which logs pass through which pipeline (since they share a receiver), and the fact that they filter out non-fluent-bit logs is a side effect. I'll roll this back and add a separate pipeline for otlp container logs.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Isn't changing the name technically a breaking change?

pmalek-sumo: It is but not in this case because using OT distro gathering metadata is still in beta and we agreed that beta features can undergo breaking changes.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Changed to `otellogs`, it's consistent with `otelagent`, and `otelcollogs` just looks weird to me.

deploy/helm/sumologic/values.yaml
```diff
@@ -4125,6 +4141,215 @@ metadata:
...
+    exporters:
+      otlphttp:
+        endpoint: http://${LOGS_METADATA_SVC}.${NAMESPACE}.svc.cluster.local:4318
+  daemonset:
+    ## Set securityContext for containers running in pods in log collector daemonset
+    securityContext:
+    # in order to reliably read logs from mounted node logging paths, we need to run as root
+      fsGroup: 0
+      runAsUser: 0
+      runAsGroup: 0
```
swiatekm-sumo, line 4329: I'm not sure if there's any other generic way of guaranteeing we can read the node logs. It seems like K8s distributions have different ideas on what the ACLs should be for these logs. On EKS and microk8s they're world-readable, but on KOPS they're not.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Changed them all to `logs/fluent-bit/*` and the otlp one to `logs/otlp/containers`. With this change, I think the purpose of the fluent include filters is much clearer as well.

deploy/helm/sumologic/values.yaml
```diff
@@ -4019,7 +4023,7 @@ metadata:
           - file_storage
           # - sumologic
         pipelines:
-          logs:
+          logs/fluent-bit/containers:
```
perk-sumo, line 4026: ❤️

deploy/helm/sumologic/values.yaml
perk-sumo: Thank you!

deploy/helm/sumologic/values.yaml
```diff
@@ -4125,6 +4141,215 @@ metadata:
...
+## This is currently highly experimental and not recommended unless you really know what you're doing
+## In particular, this configuration currently only covers container logs, and only supports docker-shim
+## This is an alpha feature, and may change in the near future
+otellogs:
+  enabled: false
+
+  ## Configure image for Opentelemetry Collector
+  image:
+    repository: public.ecr.aws/sumologic/sumologic-otel-collector
+    tag: 0.0.42-beta.0
```
perk-sumo, line 4154: We have a new image now :)

tests/helm/metadata_logs_otc/static/templates.output.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   name: RELEASE-NAME-sumologic-otelcol-logs
   labels:
     app: RELEASE-NAME-sumologic-otelcol-logs
-    chart: "sumologic-%CURRENT_CHART_VERSION%"
```
perk-sumo, line 9: Hm I'm pretty sure that's not correct 🤔

tests/helm/metadata_logs_otc/static/templates.output.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
...
+    chart: "sumologic-2.4.0-alpha.1"
```
perk-sumo, line 9: Hm I'm pretty sure that's not correct 🤔

swiatekm-sumo:
@astencel-sumo maybe you can take a look as well?

tests/helm/logs_otc/static/basic.input.yaml
```diff
@@ -0,0 +1,2 @@
+otellogs:
+  enabled: true
```
pmalek-sumo, line 2: Empty new line

tests/helm/logs_otc_daemonset/static/basic.input.yaml
```diff
@@ -0,0 +1,2 @@
+otellogs:
+  enabled: true
```
pmalek-sumo, line 2: ditto

deploy/helm/sumologic/values.yaml
```diff
@@ -4125,6 +4141,215 @@ metadata:
...
+## This is currently highly experimental and not recommended unless you really know what you're doing
+## In particular, this configuration currently only covers container logs, and only supports docker-shim
+## This is an alpha feature, and may change in the near future
+otellogs:
+  enabled: false
+
+  ## Configure image for Opentelemetry Collector
+  image:
+    repository: public.ecr.aws/sumologic/sumologic-otel-collector
+    tag: 0.0.43-beta.0
```
pmalek-sumo, line 4154:
```suggestion
    tag: 0.0.45-beta.0
```

deploy/helm/sumologic/values.yaml
```diff
@@ -4061,7 +4065,19 @@ metadata:
               - batch
             exporters:
               - sumologic/systemd
-
+          logs/otc/containers:
```
pmalek-sumo, line 4068: Can you update the pipelines in vagrant's values.yaml as well so that they are not "duplicated" in vagrant env?

perk-sumo (approved):
LGTM except for nits from Patryk 🚀

deploy/helm/sumologic/templates/logs/collector/otelcol/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.otellogs.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "sumologic.logs.collector.name.configmap" . }}
+  labels:
+    app: {{ template "sumologic.labels.app.logs.collector.configmap" . }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+data:
+  {{- (tpl (.Files.Glob "conf/logs/collector/otelcol/config.yaml").AsConfig .) | nindent 2 }}
```
pmalek-sumo, line 10: I believe using `.Files.Get` would be enough here, given that we only provide one concrete path.

https://helm.sh/docs/chart_template_guide/accessing_files/#glob-patterns

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
```diff
@@ -0,0 +1,109 @@
...
+      volumes:
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: config.yaml
+            path: config.yaml
+          name: {{ template "sumologic.logs.collector.name.configmap" . }}
+        name: otelcol-config
+      - hostPath:
+          path: /var/log/pods
```
pmalek-sumo, line 97: Did we establish that we want to use `/var/log/pods` instead of [`/var/log/containers`](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/cddb1fc41d8fa277a069f1061657f2229c8dbbbe/deploy/helm/sumologic/values.yaml#L973)?

perk-sumo: It should be the same as for `Fluent Bit` IMO 🤔

swiatekm-sumo: I'm not sure. @astencel-sumo prepared the OTC configuration originally, and it made sense to me as `/var/log/containers` is getting deprecated eventually afaik. Do you see any problems with using `/var/log/pods`?

pmalek-sumo: I don't have a strong preference but I would prefer to keep it the same as it is and **then** change it if need be later on.

Where did you see the info about it being deprecated?

deploy/helm/sumologic/templates/logs/collector/otelcol/configmap.yaml
swiatekm-sumo: `.Files.Get` returns a string rather than a `Files` object, and we'd need to manually format data for the configmap instead of calling `.AsConfig`. I think this is the simpler way.

pmalek-sumo: 👍 thanks for checking this

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
swiatekm-sumo: Basically, the official log path for CRI is `/var/log/pods`, and these are linked to `/var/log/containers` for backwards compatibility. I haven't been able to find an official deprecation timeline for the latter, was referring to [this comment](https://github.com/kubernetes/kubernetes/issues/53022#issuecomment-332150416).

I can switch it back to `/var/log/containers` for now, but it's going to be a bit annoying, as it exposes less metadata via the filepath, and I'll need to fiddle with the receiver operators again. 😒

pmalek-sumo: Ok, it's fine with me but I'd like this information to be easily discoverable by users, i.e. perhaps mention this in `values.yaml` somewhere next to the `otellogs` key?

deploy/helm/sumologic/values.yaml
pmalek-sumo, line 4068:
```suggestion
          logs/otlp/containers:
```

I'd rename this as the important bit here is actually the underlying protocol being used - `otlp` - not the agent.

deploy/helm/sumologic/values.yaml
```diff
@@ -4125,6 +4141,215 @@ metadata:
...
+              - move:
+                  from: log
+                  to: $
+    exporters:
+      otlphttp:
+        endpoint: http://${LOGS_METADATA_SVC}.${NAMESPACE}.svc.cluster.local:4318
+  daemonset:
+    ## Set securityContext for containers running in pods in log collector daemonset
+    securityContext:
+    # in order to reliably read logs from mounted node logging paths, we need to run as root
```
pmalek-sumo, line 4326:
```suggestion
    ## In order to reliably read logs from mounted node logging paths, we need to run as root
```

deploy/helm/sumologic/values.yaml
pmalek-sumo, line 4026: And perhaps (but I'm not sure) change this to `logs/fluent/containers` 🤔 Or something similar

swiatekm-sumo: If we're going to name these for input protocol, then that would be the right name, yes. Seems good to me.

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
swiatekm-sumo: Allright, I'll add a comment then.

deploy/helm/sumologic/values.yaml
```diff
@@ -4125,6 +4141,217 @@ metadata:
...
+  enabled: false
+
+  ## Configure image for Opentelemetry Collector
+  image:
+    repository: public.ecr.aws/sumologic/sumologic-otel-collector
+    tag: 0.0.45-beta.0
+    pullPolicy: IfNotPresent
+
+  logLevel: info
+  config:
```
pmalek-sumo, line 4160: Just a thought: since there's not all that much that we'd like to allow folks to customize in here I believe it would be much simpler (not littering `values.yaml` and not complicating it even further) to put this into a config map and then expose what will end up necessary to be exposed.

WDYT ?

swiatekm-sumo: Hm, since we're going to expose OTC configuration for metadata anyway, I don't think this causes any issues. We allow it for Fluent-Bit, whose configuration is less user-friendly. And users will want to modify this configuration in the future, to set up multiline logs correctly, if nothing else. I think trying to hide the details in this case would ultimately create more complexity for both us and the users.

pmalek-sumo: 🤔 Perhaps this makes sense w.r.t to multiline logs.

pmalek-sumo (approved):

PR #1991, opened 2021-12-17
test(helm): don't use yq for template tests

swiatekm-sumo:
##### Description

It's easier to run the Helm tests locally if they use the latest major version of yq, as the versions are actually incompatible.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -76,6 +76,8 @@ jobs:
 
   test:
     runs-on: ubuntu-20.04
+    env:
+      YQ_VERSION: 3.4.1
     needs:
       - markdownlint
       - shellcheck
@@ -88,7 +90,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: yq
         run: |
-            curl --retry 10 --retry-max-time 120 --retry-delay 5 -L -o /tmp/yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
+            curl --retry 10 --retry-max-time 120 --retry-delay 5 -L -o /tmp/yq https://github.com/mikefarah/yq/releases/download/${{ env.YQ_VERSION }}/yq_linux_amd64
             chmod +x /tmp/yq
             sudo mv /tmp/yq /usr/local/bin/yq
       - name: test
--- a/tests/helm/functions.sh
+++ b/tests/helm/functions.sh
@@ -29,7 +29,7 @@ function set_variables() {
   # Path to the temporary output file
   TEST_OUT="${TEST_TMP_PATH}/new_values.yaml"
   # Current version wchich should override the placeholder in test
-  CURRENT_CHART_VERSION=$(yq r "${TEST_SCRIPT_PATH}/../../../deploy/helm/sumologic/Chart.yaml" version)
+  CURRENT_CHART_VERSION=$(grep "^version: .*" "${TEST_SCRIPT_PATH}/../../../deploy/helm/sumologic/Chart.yaml" | cut -f 2 -d " ")
 }
 
 # Update helm chart and remove the tmpcharts eventually


swiatekm-sumo:
So, we can't easily upgrade the `yq` version because the upgrade scripts use it. Leaving it as is isn't that great either, as v3 is currently unmaintained. The options available to us, as I see it, are:
* Convert the upgrade scripts to v4. A significant amount of work, it looks like.
* Use v4 in Helm template tests, remove upgrade script tests from the CI and `make test`.
* Leave things as they are and be forced to use yq v3 for running helm tests locally.
What do you think @SumoLogic/open-source-collection-team ?

pmalek-sumo:
> * Leave things as they are and be forced to use yq v3 for running helm tests locally.
> 
> What do you think @SumoLogic/open-source-collection-team ?

Let's leave this as is and work with v3. I've already been there, not worth it :)

swiatekm-sumo:
> > * Leave things as they are and be forced to use yq v3 for running helm tests locally.
> > 
> > What do you think @SumoLogic/open-source-collection-team ?
> 
> Let's leave this as is and work with v3. I've already been there, not worth it :)

Well, the only thing the helm tests need it for is extracting the helm chart version, which should be easy to do with grep. You don't often need to run the migration tests locally, so that should at least make things a bit less annoying. How about doing that?

On a side note, we should update the migration scripts with a max version for yq, right now they make it seem like v4 is ok, which is not true.

pmalek-sumo:
> > > * Leave things as they are and be forced to use yq v3 for running helm tests locally.
> > > 
> > > What do you think @SumoLogic/open-source-collection-team ?
> > 
> > 
> > Let's leave this as is and work with v3. I've already been there, not worth it :)
> 
> Well, the only thing the helm tests need it for is extracting the helm chart version, which should be easy to do with grep. You don't often need to run the migration tests locally, so that should at least make things a bit less annoying. How about doing that?
> 
> On a side note, we should update the migration scripts with a max version for yq, right now they make it seem like v4 is ok, which is not true.

I don't believe we should spend time on this. There are [some simple checks around yq version](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/81b5b949bffcc7899f4cf5305c3e4643f89dc150/deploy/helm/sumologic/upgrade-2.0.0.sh#L117-L132) in the v2 upgrade script ([it should be something between 3.4 and 4.0](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/81b5b949bffcc7899f4cf5305c3e4643f89dc150/deploy/helm/sumologic/upgrade-2.0.0.sh#L12-L13))

As long as people don't complain about this I'm good with the current state of affairs.

Let's focus on the "next" version of the upgrade script instead :) Whenever we'll need that :)

swiatekm-sumo:
> As long as people don't complain about this I'm good with the current state of affairs.
I'm complaining that I want to run the template tests without installing an unmaintained version of yq XD. And generally speaking, fewer test dependencies is always better. I switched the template tests to grep, that ok with you?

pmalek-sumo:
I've checked out this branch, set `yq` to be in version: 4.16.1 and I've` run `make test` and got errors running tests for upgrade script: 

```
Test upgrade script...
[.] falco_enabled
mv: rename new_values.yaml to /Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/tmp/new_values.yaml: No such file or directory
diff: /Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/tmp/new_values.yaml: No such file or directory
rm: /Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/tmp/new_values.yaml: No such file or directory
        Log diff (/Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/static/falco_enabled.log):
0a1,2
> Error: unknown command "r" for "yq"
> Run 'yq --help' for usage.
2,5c4,20
< [INFO]    falco will be enabled. Change "falco.enabled" to "false" if you want to disable it (default for 1.0)
<
< Thank you for upgrading to v1.0.0 of the Sumo Logic Kubernetes Collection Helm chart.
< A new yaml file has been generated for you. Please check the current directory for new_values.yaml.
---
> Error: unknown command "r" for "yq"
...
```

1.0 upgrade script but still.

swiatekm-sumo:
> I've checked out this branch, set `yq` to be in version: 4.16.1 and I've`run`make test` and got errors running tests for upgrade script:
> 
> ```
> Test upgrade script...
> [.] falco_enabled
> mv: rename new_values.yaml to /Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/tmp/new_values.yaml: No such file or directory
> diff: /Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/tmp/new_values.yaml: No such file or directory
> rm: /Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/tmp/new_values.yaml: No such file or directory
>         Log diff (/Users/pmalek/code/sumologic-kubernetes-collection/tests/helm/upgrade_script/static/falco_enabled.log):
> 0a1,2
> > Error: unknown command "r" for "yq"
> > Run 'yq --help' for usage.
> 2,5c4,20
> < [INFO]    falco will be enabled. Change "falco.enabled" to "false" if you want to disable it (default for 1.0)
> <
> < Thank you for upgrading to v1.0.0 of the Sumo Logic Kubernetes Collection Helm chart.
> < A new yaml file has been generated for you. Please check the current directory for new_values.yaml.
> ---
> > Error: unknown command "r" for "yq"
> ...
> ```
> 
> 1.0 upgrade script but still.

I'm not actually changing the yq version anymore for this reason, just making the template tests not use it at all. So you still need the right version of yq to run the migration tests, but the template tests need only helm itself.

pmalek-sumo (approved):

PR #1992, opened 2021-12-17
fix(helm): add metrics port to otelcol pods

pmalek-sumo:
This PR adds port 8888 (aka `metrics`) to otelcol's pods (both metrics and logs) so that it can be used to gather otelcol's metrics

This should actually be done already but it seems we missed this somehow. Our docs already mention this:

```
 `otelcol.metrics.enabled` | Enable or disable generation of the metrics from Collector.
```


swiatekm-sumo (approved):

perk-sumo (approved):
> This should actually be done already but it seems we missed this somehow.

I believe that was done for the tracing otelcol only. But there is no reason to add this to logs and tracing ones as well.

pmalek-sumo:
> But there is no reason to add this to logs and tracing ones as well.

"there is no reason to add" or "there is no reason **not** to add"

?? @perk-sumo 🤔

PR #1998, opened 2021-12-18
test: add receiver-mock logs to printclusterstateopt

pmalek-sumo:


swiatekm-sumo:
This looks like it could be super noisy. We already log a lot of text.

pmalek-sumo:
> This looks like it could be super noisy. We already log a lot of text.

True, but this by default will only dump receiver-mock logs when the test failed and that's the scenario when you want as much information as possible in order to debug the failure.

swiatekm-sumo:
> > This looks like it could be super noisy. We already log a lot of text.
> 
> True, but this by default will only dump receiver-mock logs when the test failed and that's the scenario when you want as much information as possible in order to debug the failure.

1000 lines is a lot even with that, and receiver-mock logs are quite noisy by design, so I'm concerned this will hurt more than it'll help. Maybe we should add some less verbose logging mode to receiver-mock first?

Have you actually found this useful in practice?

pmalek-sumo:
> > > This looks like it could be super noisy. We already log a lot of text.
> > 
> > 
> > True, but this by default will only dump receiver-mock logs when the test failed and that's the scenario when you want as much information as possible in order to debug the failure.
> 
> 1000 lines is a lot even with that, and receiver-mock logs are quite noisy by design, so I'm concerned this will hurt more than it'll help. Maybe we should add some less verbose logging mode to receiver-mock first?
> 
> Have you actually found this useful in practice?

We can decrease this to e.g. 100 lines or so (but for sure that won't be that helpful in debugging a failure).

My primary goal with this is to have a "tool" when we encounter a spurious CI failure for some reason.

swiatekm-sumo:
> > > > This looks like it could be super noisy. We already log a lot of text.
> > > 
> > > 
> > > True, but this by default will only dump receiver-mock logs when the test failed and that's the scenario when you want as much information as possible in order to debug the failure.
> > 
> > 
> > 1000 lines is a lot even with that, and receiver-mock logs are quite noisy by design, so I'm concerned this will hurt more than it'll help. Maybe we should add some less verbose logging mode to receiver-mock first?
> > Have you actually found this useful in practice?
> 
> We can decrease this to e.g. 100 lines or so (but for sure that won't be that helpful in debugging a failure).
> 
> My primary goal with this is to have a "tool" when we encounter a spurious CI failure for some reason.

My main worry is that this is going to be a major pain for local debugging. Right now the test output is quite readable, even though there's a lot of it. If we dump 1000 lines of receiver-mock logs at the end, it's going to be a lot less so. So maybe make this a flag and only enable it for the CI?

pmalek-sumo:
> 
> My main worry is that this is going to be a major pain for local debugging. Right now the test output is quite readable, even though there's a lot of it. If we dump 1000 lines of receiver-mock logs at the end, it's going to be a lot less so. So maybe make this a flag and only enable it for the CI?

I'd say I'd like to have the same information available both on CI and locally in order to debug a spurious failure so I'd rather not make this (more complicated and) configurable.

pmalek-sumo:
@swiatekm-sumo This is an exemplar CI run (that failed) that I'd like to have this information (logs) available for: https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/4712698994?check_suite_focus=true :)

swiatekm-sumo (approved):
Allright, but I reserve the right to add an option to disable this if it annoys me enough when working locally.

PR #2000, opened 2021-12-20
test(integration): add tests for log count and basic metadata

swiatekm-sumo:
##### Description

Test log collection and delivery by generating a set of logs using logs-generator and then making assertions about the receiver having received them using the receiver's log counting API.

No systemd logs because:
* they don't have any metadata to query over, need https://github.com/SumoLogic/sumologic-kubernetes-tools/pull/193 first
* By default, kind doesn't enable persistence for journald, and the logs live in `/run/logs/journal` instead of `/var/log/journal` - and we don't have the former mounted in fluent-bit containers

### DIFF
--- a/tests/integration/helm_default_installation_test.go
+++ b/tests/integration/helm_default_installation_test.go
@@ -30,8 +30,9 @@ import (
 
 func Test_Helm_Default_FluentD_Metadata(t *testing.T) {
 	const (
-		tickDuration = 3 * time.Second
-		waitDuration = 3 * time.Minute
+		tickDuration            = 3 * time.Second
+		waitDuration            = 3 * time.Minute
+		logsGeneratorCount uint = 1000
 	)
 	var (
 		expectedMetrics = internal.DefaultExpectedMetrics
@@ -268,5 +269,56 @@ func Test_Helm_Default_FluentD_Metadata(t *testing.T) {
 		).
 		Feature()
 
-	testenv.Test(t, featInstall, featMetrics)
+	featLogs := features.New("logs").
+		Setup(stepfuncs.GenerateLogsWithDeployment(
+			logsGeneratorCount,
+			internal.LogsGeneratorName,
+			internal.LogsGeneratorNamespace,
+			internal.LogsGeneratorImage,
+		)).
+		Assess("logs from log generator present", stepfuncs.WaitUntilExpectedLogsPresent(
+			logsGeneratorCount,
+			map[string]string{
+				"namespace":      internal.LogsGeneratorName,
+				"pod_labels_app": internal.LogsGeneratorName,
+			},
+			internal.ReceiverMockNamespace,
+			internal.ReceiverMockServiceName,
+			internal.ReceiverMockServicePort,
+			waitDuration,
+			tickDuration,
+		)).
+		Assess("expected container log metadata is present", stepfuncs.WaitUntilExpectedLogsPresent(
+			logsGeneratorCount,
+			map[string]string{
+				"namespace":      internal.LogsGeneratorName,
+				"pod_labels_app": internal.LogsGeneratorName,
+				"container":      internal.LogsGeneratorName,
+				"deployment":     internal.LogsGeneratorName,
+				"replicaset":     "",
+				"namespace_id":   "",
+				"pod":            "",
+				"pod_id":         "",
+				"container_id":   "",
+				"host":           "",
+				"master_url":     "",
+				"node":           "",
+			},
+			internal.ReceiverMockNamespace,
+			internal.ReceiverMockServiceName,
+			internal.ReceiverMockServicePort,
+			waitDuration,
+			tickDuration,
+		)).
+		Teardown(
+			func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+				opts := *ctxopts.KubectlOptions(ctx)
+				opts.Namespace = internal.LogsGeneratorNamespace
+				k8s.RunKubectl(t, &opts, "delete", "deployment", internal.LogsGeneratorName)
+				return ctx
+			}).
+		Teardown(stepfuncs.KubectlDeleteNamespaceOpt(internal.LogsGeneratorNamespace)).
+		Feature()
+
+	testenv.Test(t, featInstall, featMetrics, featLogs)
 }
--- a/tests/integration/internal/logsgenerator/logsgenerator.go
+++ b/tests/integration/internal/logsgenerator/logsgenerator.go
@@ -1,0 +1,134 @@
+package logsgenerator
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+/// This package wraps the logsgenerator utility in sumologic-kubernetes-tools in a procedural API
+/// easily consumable by tests.
+
+const deploymentSleepTime = time.Hour * 24 // how much time we spend sleeping after generating logs in a Deployment
+
+type LogsGeneratorOptions struct {
+	// For all of these options, 0 and "" respectively are treated as "not set"
+
+	// how long we should run in total
+	// the way Duration and TotalLogs interact, is that the program exits on the first condition achieved
+	// in particular, there isn't a way to make it continue running after generating a fixed number of logs
+	Duration uint
+	// how many log lines to generate in total
+	TotalLogs uint
+	// maximum number of logs generated per second
+	LogsThroughput uint
+	// maximum number of bytes generater per second
+	BytesThroughput uint
+	// should we print diagnostic messages?
+	Verbose bool
+
+	// Pattern options
+	// See https://github.com/SumoLogic/sumologic-kubernetes-tools/tree/main/src/rust/logs-generator#patterns
+
+	// These options control random pattern generation
+	RandomPatterns      uint
+	MinPatternLength    uint
+	MaxPatternLength    uint
+	KnownWordsInRatio   uint
+	RandomWordsInRatio  uint
+	RandomDigitsInRatio uint
+
+	// This option allows the pattern to be controlled directly
+	Pattern string
+}
+
+func NewDefaultGeneratorOptions() *LogsGeneratorOptions {
+	return &LogsGeneratorOptions{
+		Duration:            0,
+		TotalLogs:           0,
+		LogsThroughput:      0,
+		BytesThroughput:     0,
+		Verbose:             false,
+		RandomPatterns:      10,
+		MinPatternLength:    5,
+		MaxPatternLength:    20,
+		KnownWordsInRatio:   7,
+		RandomWordsInRatio:  2,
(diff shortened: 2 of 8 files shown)


tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -122,6 +122,41 @@ func (client *ReceiverMockClient) GetMetricsSamples(
 	return metricsSamples, nil
 }
 
+type LogsCountResponse struct {
+	Count uint
+}
+
+func (client *ReceiverMockClient) GetLogsCount(t *testing.T, metadataFilters MetadataFilters) (uint, error) {
```
pmalek-sumo, line 129: Comment

tests/integration/internal/stepfuncs/assess_funcs.go
```diff
@@ -75,3 +77,49 @@ func WaitUntilExpectedMetricsPresent(
...
+	return func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+		kubectlOpts := *ctxopts.KubectlOptions(ctx)
+		kubectlOpts.Namespace = receiverMockNamespace
+		terrak8s.WaitUntilServiceAvailable(t, &kubectlOpts, receiverMockServiceName, int(waitDuration), tickDuration)
+
+		client, closeTunnelFunc := receivermock.NewClientWithK8sTunnel(ctx, t)
+		defer closeTunnelFunc()
+
+		assert.Eventually(t, func() bool {
+			logsCount, err := client.GetLogsCount(t, expectedLogsMetadata)
```
pmalek-sumo, line 102: If you do this in `assert.Eventually` and this fails once (marking `t` as failed) then the whole test will fail regardless of the final result. IIRC

tests/integration/internal/stepfuncs/logs.go
```diff
@@ -0,0 +1,46 @@
...
+			logsGeneratorImage,
+			generatorOptions,
+		)
+
+		// create the namespace
+		namespace := corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: logsGeneratorNamespace}}
+		err := client.Resources().Create(ctx, &namespace)
+		if err != nil {
+			t.Fatalf(err.Error())
+		}
```
pmalek-sumo, line 39:
```suggestion
		require.NoErrot(t, client.Resources().Create(ctx, &namespace))
```

but actually I'm not sure if this (either of those versions) wouldn't prevent the teardowns to run 🤔 can you check that?

tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -122,6 +122,41 @@ func (client *ReceiverMockClient) GetMetricsSamples(
...
+	if statusCode != 200 {
+		return 0, fmt.Errorf("received status code %d in response to receiver request", statusCode)
+	}
+
+	var response LogsCountResponse
+	err = json.Unmarshal([]byte(body), &response)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return response.Count, nil
```
pmalek-sumo, line 157: Why `t.Fatal`? Isn't that just an error which you can return from this func?

tests/integration/internal/logsgenerator/logsgenerator.go
```diff
@@ -0,0 +1,133 @@
...
+		RandomDigitsInRatio: 1,
+		Pattern:             "",
+	}
+}
+
+func GetLogsGeneratorDeployment(
+	namespace string,
+	name string,
+	image string,
+	options *LogsGeneratorOptions,
```
pmalek-sumo, line 69: Any reason to use a pointer?

tests/integration/internal/logsgenerator/logsgenerator.go
```diff
@@ -0,0 +1,133 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+/// This package wraps the logsgenerator utility in sumologic-kubernetes-tools in a procedural API
+/// easily consumable by tests.
+
+const deploymentSleepTime = 60 * 60 * 24 // how much time we spend sleeping after generating logs in a Deployment
```
pmalek-sumo, line 15: Can we declare that as `time.Duration` with a time unit and then where needed convert to (seconds?) whatever is used here to prevent "implicit" time units usage.

pmalek-sumo (changes requested):

tests/integration/internal/receivermock/receiver_mock.go
swiatekm-sumo: Same reason we Fatal if url parsing fails, it almost always implies a bug in the code, and shouldn't just be retried. It's simpler this way because the caller doesn't need to inspect the error and decide whether they should Fatal themselves or continue.

tests/integration/internal/logsgenerator/logsgenerator.go
swiatekm-sumo: I don't think so, I think I was working on Rust code in tools at the same time and got semantics mixed up. Will switch to a value.

tests/integration/internal/stepfuncs/assess_funcs.go
swiatekm-sumo: True, but it should only fail in extreme circumstances (like if we get a 200 and can't decode the data, for example). Some errors should cause the whole test to fail, and I think it's easier if the function itself decides this instead of having to inspect the error here.

tests/integration/internal/stepfuncs/logs.go
swiatekm-sumo: It's possible to stop teardowns from running by erroring out of a feature? 🙀

tests/integration/internal/receivermock/receiver_mock.go
pmalek-sumo: Got it. Kinda makes sense but then why do we use the helper that does the http get which  requires `t` (on which it can set the test as failed)? This might be a transient error that we don't want to fail our whole testcase on, right?

tests/integration/internal/receivermock/receiver_mock.go
```diff
@@ -122,6 +122,41 @@ func (client *ReceiverMockClient) GetMetricsSamples(
...
 
+type LogsCountResponse struct {
+	Count uint
+}
+
+func (client *ReceiverMockClient) GetLogsCount(t *testing.T, metadataFilters MetadataFilters) (uint, error) {
+	path, err := url.Parse("logs/count")
+	if err != nil {
+		t.Fatal(err)
+	}
```
pmalek-sumo, line 133: It might be worth thinking about putting this e.g. in the `ReceiverMockClient` struct itself (or as package wide unexported variable parsed at `init`)

tests/integration/internal/stepfuncs/logs.go
pmalek-sumo: `testing` framework and `e2e` framework work a bit on a different level and I'm not sure. Depends how e2e's teardown are implemented

tests/integration/internal/receivermock/receiver_mock.go
swiatekm-sumo: Good point, I should switch to the helper that doesn't fail tests and handle the error explicitly. I like the terratest helpers in general because the logging they provide is useful for debugging.

tests/integration/internal/receivermock/receiver_mock.go
swiatekm-sumo: I'll leave a TODO for it.

tests/integration/internal/stepfuncs/logs.go
swiatekm-sumo: Ok, so the way this works is that assessments are run as subtests, so calling `t.FailNow()` (which is what both `t.Fatal()` and `require.NoError()` do) only fails the given assessment and moves on. Setups and Teardowns, however, are run in the containing test context, which means that by doing any of the aforementioned, we abort the test completely.

I'm not sure we can do any better though. If a setup or teardown action fails, we can't continue, and we don't know how to clean up after ourselves. If e2e_framework supported pairing setup and teardown actions, it could clean up correctly - maybe we should propose this feature to them?

In the interim, we either do what we're doing or roll our own finalization.

pmalek-sumo: Thanks for checking this. I'm good with the current state of affairs. I'm not sure if pairing up setups and teardowns is a necessary feature for the e2e framework 🤔 It would complicate things a bit for a debatable gain

tests/integration/helm_default_installation_test.go
```diff
@@ -268,5 +269,56 @@ func Test_Helm_Default_FluentD_Metadata(t *testing.T) {
...
+			tickDuration,
+		)).
+		Assess("expected container log metadata is present", stepfuncs.WaitUntilExpectedLogsPresent(
+			logsGeneratorCount,
+			map[string]string{
+				"namespace":      internal.LogsGeneratorName,
+				"pod_labels_app": internal.LogsGeneratorName,
+				"container":      internal.LogsGeneratorName,
+				"deployment":     internal.LogsGeneratorName,
+				"replicaset":     "",
```
pmalek-sumo, line 298: Just a song of the future: these could be made a bit more sophisticated so that the function itself accepts not `map[string]string` but an interface that instead of matching only 1:1 could also be implemented as regex matcher.

tests/integration/internal/logsgenerator/logsgenerator.go
```diff
@@ -0,0 +1,134 @@
...
+		ObjectMeta: metadata,
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: appLabels,
+			},
+			Template: podTemplateSpec,
+		},
+	}
+	return deployment
```
pmalek-sumo, line 111:
```suggestion
	return appsv1.Deployment{
		ObjectMeta: metadata,
		Spec: appsv1.DeploymentSpec{
			Replicas: &replicas,
			Selector: &metav1.LabelSelector{
				MatchLabels: appLabels,
			},
			Template: podTemplateSpec,
		},
	}
```

pmalek-sumo (approved):

tests/integration/internal/stepfuncs/logs.go
swiatekm-sumo: The gain would be the ability to clean up resources created in previous setups if one of them fails. It'd be pretty straightforward in that you'd basically pass in a second "teardown" stepfunc to Setup, and the framework would simply `defer` that stepfunc whenever it successfully completed its respective setup stepfunc. `t.FailNow` already calls all deffereds while existing.

Either way, we can merge this as-is, right?

tests/integration/helm_default_installation_test.go
swiatekm-sumo: Could they? We're restricted by what receiver-mock allows, and what we give it are just strings.

pmalek-sumo: You could get creating and limit search by one set of rules and then match via regex the results.

Or implement regex matching in receiver-mock 🚀

PR #2002, opened 2021-12-20
docs: add troubleshooting section about fluentd buffer permissions issue

pmalek-sumo:


deploy/docs/Troubleshoot_Collection.md
```diff
@@ -477,3 +479,79 @@ kube-prometheus-stack:
...
+When you encounter the following (or a similar) error message in fluentd logs:
+
+```
+2021-11-23 07:05:56 +0000 [error]: #0 unexpected error error_class=Errno::EACCES error="Permission denied @ dir_s_mkdir - /fluentd/buffer/logs"
+```
+
+this means that most likely the volume that has been provisioned as PersistentVolume
+for your fluentd has incorrect ownership and/or permissions set.
+
+You can verify that with the following `kubectl` command:
```
swiatekm-sumo, line 494: `that` -> `that this is the case` - reads nicer.

deploy/docs/Troubleshoot_Collection.md
```diff
@@ -477,3 +479,79 @@ kube-prometheus-stack:
...
+
+```
+$ kubectl exec -it -n <NAMESPACE> <RELEASE_NAME>-<NAMESPACE>-fluentd-logs-0 \
+  --container fluentd -- ls -ld /fluentd/buffer
+drwx------ 6 root root 4096 Dec 17 16:01 /fluentd/buffer
+```
+
+In the above snippet you can observe that `/fluentd/buffer/` is owned by `root`
+and only it can access it.
```
swiatekm-sumo, line 503: `is owned by the root user, and only that user can access it.`

deploy/docs/Troubleshoot_Collection.md
```diff
@@ -477,3 +479,79 @@ kube-prometheus-stack:
...
+
+There can be a multitude of reason why that's the case, this can depend on the
+cloud provider that you use and the StorageClasses that are available/set in your cluster.
```
swiatekm-sumo, line 506: `There are many possible reasons for this behaviour, (...)`? "Multitude of reasons" just sounds strange to me.

swiatekm-sumo (commented):
A few minor style comments, looks good to me otherwise.

deploy/docs/Troubleshoot_Collection.md
pmalek-sumo, line 494:
```suggestion
You can verify that this is the case with the following `kubectl` command:
```

deploy/docs/Troubleshoot_Collection.md
pmalek-sumo, line 503:
```suggestion
In the above snippet you can observe that `/fluentd/buffer/` is owned by `root`, and
only that user can access it.
```

deploy/docs/Troubleshoot_Collection.md
pmalek-sumo, line 506:
```suggestion
There are many possible reasons for this behaviour, this can depend on the
cloud provider that you use and the StorageClasses that are available/set in your cluster.
```

swiatekm-sumo (approved):

PR #2007, opened 2021-12-21
ci: use tmpfs for etcd storage in kind CI tests

swiatekm-sumo:
##### Description

Supposedly this improves runtimes on Github Runners.


swiatekm-sumo:
Can't see any difference, closing.

PR #2015, opened 2022-01-03
test(integration): add test for multiline logs in containerd log format

kkujawa-sumo:
##### Description

test(integration): add test for multiline logs in containerd log format

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

- run `make test TEST_NAME="Test_Helm_FluentBit_Containerd_Multiline_Logs"`

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -17,8 +17,12 @@ This directory contains `sumologic-kubernetes-collection` integration tests util
   - [Creating `kind` cluster](#creating-kind-cluster)
   - [Reusing a preexisting cluster](#reusing-a-preexisting-cluster)
 - [Runtime options](#runtime-options)
-- [Running specified tests](#running-specified-tests)
+- [Filtering tests](#filtering-tests)
+  - [Running specific tests](#running-specific-tests)
+  - [Running specific features/assessments](#running-specific-featuresassessments)
 - [K8s node images matrix](#k8s-node-images-matrix)
+- [Known limitations/issues](#known-limitationsissues)
+  - [Maximum size of arguments for processes](#maximum-size-of-arguments-for-processes)
 
 ---
 
@@ -159,3 +163,27 @@ Node images (k8s versions) on which tests are being run are defined in
 Where `.default` sets what is being used by default and `.supported` indicates on which
 images the tests should pass - CI configuration will ensure that tests are being on
 each of those image versions.
+
+## Known limitations/issues
+
+### Maximum size of arguments for processes
+
+Maximum size of arguments for processes is limited by:
+
+- `ARG_MAX` - maximum length of arguments
+- `MAX_ARG_STRLEN` - maximum size for one argument
+
+__Note__: `ARG_MAX` and `MAX_ARG_STRLEN` vary on different systems.
+
+Because of limited size of arguments it is not possible to prepare Pod with very long list of arguments which is generated in the code, e.g.
+Pod with very long list of `echo` commands that have very long string passed as an argument.
+
+When process receives too long list of arguments following error occurs:
+
+```
+2022-02-22T13:57:55.662403778Z stderr F standard_init_linux.go:228: exec user process caused: argument list too long
+```
+
+To learn more about `ARG_MAX` and `MAX_ARG_STRLEN` please read [ARG_MAX, maximum length of arguments for a new process][ARG_MAX_ARTICLE]
+
+[ARG_MAX_ARTICLE]: https://www.in-ulm.de/~mascheck/various/argmax/#maximum_number
--- a/tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
+++ b/tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
@@ -1,0 +1,112 @@
+package integration
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal"
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/ctxopts"
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/stepfuncs"
+)
+
+func Test_Helm_FluentBit_Containerd_Multiline_Logs(t *testing.T) {
+	const (
+		tickDuration           = 3 * time.Second
+		waitDuration           = 3 * time.Minute
+		logRecords             = 4   //	number of log records in single loop, see: tests/integration/yamls/pod_multiline_long_lines.yaml
+		logLoops               = 500 // number of loops in which logs are generated, see: tests/integration/yamls/pod_multiline_long_lines.yaml
+		multilineLogCount uint = logRecords * logLoops
+	)
+
+	featInstall := features.New("installation").
+		Assess("sumologic secret is created",
+			func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+				k8s.WaitUntilSecretAvailable(t, ctxopts.KubectlOptions(ctx), "sumologic", 60, tickDuration)
+				secret := k8s.GetSecret(t, ctxopts.KubectlOptions(ctx), "sumologic")
+				require.Len(t, secret.Data, 10)
+				return ctx
+			}).
+		Assess("fluentd logs statefulset is ready",
+			stepfuncs.WaitUntilStatefulSetIsReady(
+				waitDuration,
+				tickDuration,
+				stepfuncs.WithNameF(
+					stepfuncs.ReleaseFormatter("%s-sumologic-fluentd-logs"),
+				),
+				stepfuncs.WithLabelsF(
+					stepfuncs.LabelFormatterKV{
+						K: "app",
+						V: stepfuncs.ReleaseFormatter("%s-sumologic-fluentd-logs"),
+					},
+				),
+			),
+		).
+		Assess("fluentd logs buffers PVCs are created",
+			func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+				namespace := ctxopts.Namespace(ctx)
+				releaseName := ctxopts.HelmRelease(ctx)
+				kubectlOptions := ctxopts.KubectlOptions(ctx)
+
+				t.Logf("kubeconfig: %s", kubectlOptions.ConfigPath)
+				cl, err := k8s.GetKubernetesClientFromOptionsE(t, kubectlOptions)
+				require.NoError(t, err)
+
+				assert.Eventually(t, func() bool {
+					pvcs, err := cl.CoreV1().PersistentVolumeClaims(namespace).
+						List(ctx, v1.ListOptions{
+							LabelSelector: fmt.Sprintf("app=%s-sumologic-fluentd-logs", releaseName),
+						})
+					if !assert.NoError(t, err) {
+						return false
+					}
+
+					return err == nil && len(pvcs.Items) == 3
+				}, waitDuration, tickDuration)
+				return ctx
+			}).
+		Assess("fluent-bit daemonset is running",
+			func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+				var daemonsets []appsv1.DaemonSet
+				require.Eventually(t, func() bool {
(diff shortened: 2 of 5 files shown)


tests/integration/internal/stepfuncs/logs.go
```diff
@@ -42,3 +42,9 @@ func GenerateLogsWithDeployment(
 		return ctx
 	}
 }
+
+func DeployPodGeneratingMultilineLogs() features.Func {
+	return func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+		return ctx
+	}
+}
```
pmalek-sumo, line 50: 🤔

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
```diff
@@ -0,0 +1,318 @@
...
+		multilineLogCount uint = logRecords * logLoops
+	)
+	var (
+		expectedMetrics = internal.DefaultExpectedMetrics
+	)
+
+	// TODO:
+	// Refactor this: we should find a way to inject this into step func helpers
+	// like stepfuncs.WaitUntilPodsAvailable() instead of relying on an implementation
+	// detail.
```
pmalek-sumo, line 46: There's a PR that tries to address this: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2012 

Do you mind waiting with this until it gets merged :) ?

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
```diff
@@ -0,0 +1,318 @@
...
+						})
+					if !assert.NoError(t, err) {
+						return false
+					}
+
+					return err == nil && len(pvcs.Items) == 3
+				}, waitDuration, tickDuration)
+				return ctx
+			}).
+		Assess("fluentd metrics pods are available",
```
pmalek-sumo, line 90: I'd remove all the code related to metrics both from the test code and from values.yaml.

This is already tested elsewhere. Just leave here whatever is related to containerd multiline logs.

I'd also remove the tests related to logs as that is tested also in a separate test.

tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
```diff
@@ -0,0 +1,222 @@
...
+          Systemd_Filter  _SYSTEMD_UNIT=rpc-statd.service
+          Systemd_Filter  _SYSTEMD_UNIT=rpcbind.service
+          Systemd_Filter  _SYSTEMD_UNIT=set-aws-environment.service
+          Systemd_Filter  _SYSTEMD_UNIT=system-cloudinit.service
+          Systemd_Filter  _SYSTEMD_UNIT=systemd-timesyncd.service
+          Systemd_Filter  _SYSTEMD_UNIT=update-ca-certificates.service
+          Systemd_Filter  _SYSTEMD_UNIT=user-cloudinit.service
+          Systemd_Filter  _SYSTEMD_UNIT=var-lib-etcd2.service
+          Max_Entries     1000
+          Read_From_Tail  true
```
pmalek-sumo, line 144: Do we need this for this test?

pmalek-sumo (changes requested):

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
kkujawa-sumo: I think that this PR can wait until #2012 is merged

tests/integration/internal/stepfuncs/logs.go
kkujawa-sumo: Thanks for catching this, it is no longer needed.

tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
kkujawa-sumo: OK, we can remove configuration for systemd logs, I removed also tests for non-multiline logs from `Test_Helm_FluentBit_Containerd_Multiline_Logs`

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
kkujawa-sumo: done

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
pmalek-sumo: @kkujawa-sumo https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2012 is merged 🎊

kkujawa-sumo: @pmalek-sumo please take another look

tests/integration/yamls/pod_multiline_long_lines.yaml
```diff
@@ -0,0 +1,28 @@
...
+        echo "Dec 13 09:41:08 1st single line...";
+        echo "Dec 13 09:41:08 2nd single line...";
+        echo "Dec 14 06:41:08 Exception in thread "main" java.lang.RuntimeException: Something has gone wrong, aborting! ${LONG_STRING} end of the 1st long line
+          at com.myproject.module.MyProject.badMethod(MyProject.java:22)
+          at com.myproject.module.MyProject.oneMoreMethod(MyProject.java:18)
+          at com.myproject.module.MyProject.anotherMethod(MyProject.java:14)
+          at com.myproject.module.MyProject.someMethod(MyProject.java:10)";
+        echo "    at com.myproject.module.MyProject.verylongLine(MyProject.java:100000) ${LONG_STRING} end of the 2nd long line";
+        echo "    at com.myproject.module.MyProject.main(MyProject.java:6)
+      Dec 15 09:41:08 another line in loop ${i}";
```
pmalek-sumo, line 27: Just for my memory: does that setup have this limitation that if we didn't send this very last log line then none of the previous logs would be flushed?

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/stepfuncs"
+)
+
+func Test_Helm_FluentBit_Containerd_Multiline_Logs(t *testing.T) {
+	const (
+		tickDuration           = 3 * time.Second
+		waitDuration           = 3 * time.Minute
+		logRecords             = 4
+		logLoops               = 500
+		multilineLogCount uint = logRecords * logLoops
```
pmalek-sumo, line 29: Maybe add some comments that these values are hardcoded in the pod definition yaml?

Or maybe even better: instead of applying a `.yaml` that's saved on disk we could define the pod in code and then we could have control over the loop counts and we wouldn't have to rely on "implicit dependency" between these `consts` here and the pod `.yaml` file?

WDYT?

swiatekm-sumo: As a point of reference, we've defined a similar thing in code here: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/internal/logsgenerator/logsgenerator.go

tests/integration/yamls/pod_multiline_long_lines.yaml
kkujawa-sumo: In this solution buffering of log lines is controlled in `tail` input plugin so it s as good a previously used configurations, lua filter just cleans logs by removing information specific for containerd log format and joins multipart logs into single line (which are sent as part of multiline log).

I see that there is `Flush` parameter in `Service` section https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/configuration-file#config_section and it seems that it prevents the situation that nothing is sent.

tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
```diff
@@ -0,0 +1,134 @@
...
+  setupEnabled: true
+  accessId: "dummy"
+  accessKey: "dummy"
+  endpoint: http://receiver-mock.receiver-mock:3000/terraform/api/
+  metrics:
+    enabled: false
+
+fluent-bit:
+  luaScripts:
+    parse_logs.lua: |
```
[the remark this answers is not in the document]
pmalek-sumo: bump: @kkujawa-sumo Do you mind adding this?

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
pmalek-sumo: I've created an issue for this so that we can perhaps make this use less hardcoded and (very) loosely coupled `.yaml` files.

https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/2055

Ideally we could use that approach elsewhere as well, e.g. in the place where you mentioned it @swiatekm-sumo

tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
```diff
@@ -0,0 +1,136 @@
+sumologic:
+  setupEnabled: true
+  accessId: "dummy"
+  accessKey: "dummy"
+  endpoint: http://receiver-mock.receiver-mock:3000/terraform/api/
+  metrics:
+    enabled: false
```
pmalek-sumo, line 7: As one last thing I'd ask to rebase this PR and remove this as well as the fluentd resource section since that is already covered by https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2041

And then only override what's not set in common yaml.

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
kkujawa-sumo: It will be good to have everything in the code but unfortunately size of arguments for processes is limited, related articles [1](https://www.in-ulm.de/~mascheck/various/argmax/), [2](https://www.baeldung.com/linux/argument-list-too-long-error) 

Following error occurred:
```
cat /var/log/containers/example-pod-multiline-logs-long-lines_multiline-logs-generator_example-pod-multiline-logs-long-lines-3cd7f6f3fd6f2a4eda236a9c55882beb05efe78f3e18e2695b09d7b635489433.log 
2022-02-22T13:57:55.662403778Z stderr F standard_init_linux.go:228: exec user process caused: argument list too long
```
and because of that we cannot pass explicitly very long strings  as arguments to `echo` and we cannot create very long number of arguments as `args` in Pod specification

Experiments with changing Pod in `yaml` to `go` are available on branch: https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/kk-test-multiline-logs-trial-to-replace-yaml-with-go

tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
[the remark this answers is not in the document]
kkujawa-sumo: Infromation is added [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2015/files#diff-1e8f2418efe03ffd07e138c023c67c17153d806f70176c3f03426e088670240cR2)

tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
kkujawa-sumo: done

tests/integration/helm_fluentbit_containerd_multiline_logs_test.go
pmalek-sumo: Thanks for checking if it's possible 🙏 

Ideally I'd like to have that documented somewhere, perhaps an issue? Perhaps a comment? So that we know we've investigated that and we know what are the limitations.

Maybe a "known limitations/issues" section in tests readme?

kkujawa-sumo:
@pmalek-sumo please take another look

tests/integration/README.md
```diff
@@ -159,3 +163,27 @@ Node images (k8s versions) on which tests are being run are defined in
...
+
+When process receives too long list of arguments following error occurs:
+
+```
+2022-02-22T13:57:55.662403778Z stderr F standard_init_linux.go:228: exec user process caused: argument list too long
+```
+
+To learn more about `ARG_MAX` and `MAX_ARG_STRLEN` please read [ARG_MAX, maximum length of arguments for a new process][ARG_MAX_ARTICLE]
+
+[ARG_MAX_ARTICLE]: https://www.in-ulm.de/~mascheck/various/argmax/#maximum_number
```
pmalek-sumo, line 189: 🚀

tests/integration/internal/constants.go
```diff
@@ -22,6 +22,10 @@ const (
 	LogsGeneratorNamespace = "logs-generator"
 	LogsGeneratorName      = "logs-generator"
 	LogsGeneratorImage     = "sumologic/kubernetes-tools:2.9.0"
+
+	MultilineLogsNamespace   = "default"
```
pmalek-sumo, line 26: Can we make it a bit more concrete/unique? So that in case we need to debug we don't have logs from other stuff mixed in the `default` namespace? We can call it e.g. `multiline-tests` or something like that

kkujawa-sumo: ok

PR #2036, opened 2022-01-25
chore: update telegraf to 1.21

swiatekm-sumo:
##### Description

1.14 is pretty old and contains quite a few vulnerabilities. I ran E2E tests on 1.21 and they passed without issue.

---

##### Checklist

- [x] Changelog updated

###### Testing performed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- chore: update the Telegraf image to 1.21.2 [#2036][#2036]
+
 ### Fixed
 
 [Unreleased]: https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v2.4.0...main
+[#2036]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2036
 
 ## [v2.4.0][v2_4_0]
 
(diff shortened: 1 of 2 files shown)


pmalek-sumo (changes requested):
Are there any breaking changes between 1.14.4 and 1.21.2?

I'd be more inclined to bump the operator version (that handles the image version itself) instead of bumping the image only.

swiatekm-sumo:
> Are there any breaking changes between 1.14.4 and 1.21.2?
> 
> I'd be more inclined to bump the operator version (that handles the image version itself) instead of bumping the image only.
> 
> rel: https://github.com/influxdata/telegraf-operator/releases

No breaking changes in Telegraf itself: https://github.com/influxdata/telegraf/blob/master/CHANGELOG.md. They seem to be following semantic versioning even if they don't state so explicitly in their documentation.

For Telegraf Operator, we're already running the latest version, and the default Telegraf version in its Helm Chart is 1.19. Do you recall if the 1.14.4 image was hosted on our ECR as a matter of policy, or if it was customized in some way?

pmalek-sumo (commented):
I believe the changelog would need some rebasing. Wouldn't it? I wonder why github doesn't show it like that.

swiatekm-sumo:
> I believe the changelog would need some rebasing. Wouldn't it? I wonder why github doesn't show it like that.

Good catch. Github does the right thing, since from its perspective there's no conflict in the diff, just an unrelated change in a different line (Unreleased -> 2.4.0).

pmalek-sumo (approved):

PR #2039, opened 2022-01-25
docs: remove the 'unreleased' section from the v2.4.0 changelog

perk-sumo:


pmalek-sumo (approved):
I'm actually thinking that this could stay since with a very first change that will go to Changelog we'll need to add it back.

PR #2041, opened 2022-01-25
test(integration): use common values.yaml for resources, access keys/IDs and receiver-mock URL

pmalek-sumo:


tests/integration/values/values_common.yaml
```diff
@@ -0,0 +1,73 @@
+sumologic:
```
swiatekm-sumo, line 1: Can you also add a short comment here, explaining what this is for?

tests/integration/internal/stepfuncs/debug.go
```diff
@@ -32,6 +32,8 @@ func PrintClusterStateOpt(force ...bool) features.Func {
 			)
 
 			k8s.RunKubectl(t, ctxopts.KubectlOptions(ctx), "get", "all")
+			k8s.RunKubectl(t, ctxopts.KubectlOptions(ctx), "get", "pvc")
+			k8s.RunKubectl(t, ctxopts.KubectlOptions(ctx), "get", "pv")
```
swiatekm-sumo, line 36: Can you move this to a separate commit?

swiatekm-sumo (changes requested):
Good idea, left some minor comments.

tests/integration/internal/stepfuncs/debug.go
pmalek-sumo: Done

tests/integration/values/values_common.yaml
pmalek-sumo: Done

swiatekm-sumo (approved):

PR #2043, opened 2022-01-26
docs: keep versions information in one place

perk-sumo:


deploy/README.md
```diff
@@ -1,4 +1,4 @@
-# Deployment Guide
+# Deployment Guide for release v2.4
```
pmalek-sumo, line 1: Not sure if that's a good idea, this will get old very quick.

astencel-sumo (approved):

deploy/README.md
perk-sumo: Yes, that's the question what to put here in the main branch.
Maybe it should be `for unreleased version`?

This particular change here was to show how I think we should keep things in release branches.

perk-sumo: Yes, that's the question what to put here in the main branch.
Maybe it should be `for unreleased version`?

This particular change here was to show how I think we should keep things in release branches.

perk-sumo: Changed.

pmalek-sumo (approved):

PR #2052, opened 2022-01-27
[Backport release-v2.4] chore: update Telegraf to 1.21.2

swiatekm-sumo:
##### Description

Backport the Telegraf update to 2.4.

---

##### Checklist

- [X] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+### Changed
+
+- chore: update the Telegraf image to 1.21.2 [#2036][#2036]
+
+### Fixed
+
+[Unreleased]: https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v2.4.0...main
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
pmalek-sumo, line 8: Secondly I just thought about (IMHO) a nice idea how to mark the next release, that yet unreleased. See the below suggestion.

```suggestion
## v2.4.1

### Unreleased
	
```

WDYT? (obviously we can tweak that)

pmalek-sumo (commented):
I'm not sure if we didn't establish that we won't be backporting this 🤔 .

swiatekm-sumo:
> I'm not sure if we didn't establish that we won't be backporting this 🤔 .

We agreed on 2.4, but not earlier versions. In general, I think it's good to backport relatively minor (in scope) security upgrades to the latest major version.

CHANGELOG.md
swiatekm-sumo: I think this might be a bit confusing if someone were to look at the branch and see the version in the changelog without an accompanying release tag. I think that's the reason we don't do this in main either.

pmalek-sumo (approved):

PR #2072, opened 2022-02-02
test(integration): decrease expected kubelet logs count

pmalek-sumo:
It seems that sometimes we don't get 10 logs from kubelet ([failed CI](https://github.com/SumoLogic/sumologic-kubernetes-collection/runs/5034342403?check_suite_focus=true)) so decrease the number so that we don't get any more false failures.

### DIFF
--- a/tests/integration/helm_default_installation_test.go
+++ b/tests/integration/helm_default_installation_test.go
@@ -333,7 +333,7 @@ func Test_Helm_Default_FluentD_Metadata(t *testing.T) {
 			tickDuration,
 		)).
 		Assess("logs from kubelet present", stepfuncs.WaitUntilExpectedLogsPresent(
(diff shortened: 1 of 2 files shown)


sumo-drosiek:
Is there any reason why 1 is not sufficient?

pmalek-sumo:
> Is there any reason why 1 is not sufficient?

Not really.

sumo-drosiek:
> > Is there any reason why 1 is not sufficient?
> 
> Not really.

Lets use 1 then as it means `we got any log`

pmalek-sumo:
> > > Is there any reason why 1 is not sufficient?
> > 
> > 
> > Not really.
> 
> Lets use 1 then as it means `we got any log`

k, done

sumo-drosiek (approved):

PR #2124, opened 2022-02-14
docs: using a custom CA certificate

swiatekm-sumo:
##### Description

Document how to integrate the Helm Chart with a TLS intercepting proxy and a custom CA certificate.

Also add a brief note about upgrading while rehosting container images.

### DIFF
--- a/deploy/docs/Security_Best_Practices.md
+++ b/deploy/docs/Security_Best_Practices.md
@@ -91,3 +91,85 @@ fluentd:
   metrics:
     ...
 ```
+
+## Using a custom root CA for TLS interception
+
+Unfortunately, there isn't a uniform way of adding root certificates to application containers.
+Generally speaking, this can be done in two ways:
+
+- At container image build time, making use of the underlying OS certificate management facilities
+- At Pod start time, using `initContainers` and `extraVolumes`, also making use of the underlying OS certificate management facilities
+
+### Adding a custom root CA certificate by rebuilding container images
+
+Adding certificates during container build using the underlying Linux distribution's management facilities
+tends to be straightforward. There are three different base distributions among container images used by the
+collection Helm Chart: `debian`, `alpine` and `scratch`.
+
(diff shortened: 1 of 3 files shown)


deploy/docs/Security_Best_Practices.md
```diff
@@ -91,3 +91,85 @@ fluentd:
...
+
+```Dockerfile
+COPY cert.pem /usr/local/share/ca-certificates/cert.crt
+RUN cat /usr/local/share/ca-certificates/cert.crt >> /etc/ssl/certs/ca-certificates.crt
+```
+
+Keep in mind that this needs to be done as the root user, and then the user should be switched back to the original image's
+default.
+
+### Adding a custom root CA certificate using initContainers
```
pmalek-sumo, line 120:
```suggestion
### Adding a custom root CA certificate using `initContainers`
```

deploy/docs/Security_Best_Practices.md
```diff
@@ -91,3 +91,85 @@ fluentd:
...
+    statefulset:
+      initContainers:
+        - name: update-certificates
+          image: odise/busybox-curl
+          command: 
+            - 'sh'
+            - '-c'
+            - |
+              "cp /etc/ssl/certs/ca-certificates.crt /certs/ca-certificates.crt && \
+              cat /root/ca-cert/cert.pem >> /certs/ca-certificates.crt"
```
pmalek-sumo, line 140: Wouldn't this just work?

```suggestion
            - sh
            - -c
            - |
              cp /etc/ssl/certs/ca-certificates.crt /certs/ca-certificates.crt
              cat /root/ca-cert/cert.pem >> /certs/ca-certificates.crt
```

swiatekm-sumo: I think it would, yeah.

pmalek-sumo (approved):

PR #2144, opened 2022-02-23
feat(metrics): drop container label for non-container kube state metrics

### LINKED ISSUE #2097: Explorer view in Sumo UI shows kube-state-metrics container added to every pod
**Describe the bug**
The issue can be observed in the Explore view in Sumo UI (go to Sumo UI, create new Explore tab), when viewing a k8s cluster's data e.g. by Kubernetes Namespace View. You can see that there is a kube-state-metrics container displayed in the left-hand navigation tree for every single pod in the cluster.

**Description of the cause of the issue and possible fix**
This is due to Prometheus labelling the metrics from kube-state-metrics with container=kube-state-metrics. We are already removing Prometheus-decorated pod and service labels from non-pod kube-state-metrics metrics, but it looks like we should also remove the container label from non-container pod metrics like kube_pod_phase_status and other kube_pod_*, but not from container metrics like kube_pod_container_*.

**Configuration**
Default configuration used.

**To Reproduce**
Install the Helm Chart and go to Explore view in Sumo UI.

**Expected behavior**
`kube-state-metrics` should not be shown for evey pod in the cluster.

image

sumo-drosiek:
##### Description

Drop container label for non-container kube state metrics. It requires to extract `kube_pod_status_phase` which shouldn't have container label but should have `pod`/`service` labels.

Fixes #2097

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - chore: bump sumo ot distro to 0.0.50-beta.0 [#2127][#2127]
+- feat(metrics): drop container label for non-container kube state metrics [#2144][#2144]
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - feat: add otelcol's liveness and readiness probes configuration [#2105][#2105]
 - docs: add fluentd buffers vs DPM calculations info for metrics [#2128][#2128]
 - feat(otelcol/metrics): adjust metric otelcol configuration [#2134][#2134]
+- feat(metrics): drop container label for non-container kube state metrics [#2144][#2144]
```
pmalek-sumo, line 15: I think it's more of a `Changed` entry, isn't it?

PR #2154, opened 2022-02-28
feat: enable metrics/traces collection from instrumentation by default

mat-rumian:
##### Description

It's time to enable tracing ingestion by default. OTC will be deployed with SKC by default.

---

##### Checklist

- [x] Changelog updated

###### Testing performed

- [x] Deploy chart to cluster


deploy/helm/sumologic/values.yaml
```diff
@@ -252,9 +252,9 @@ sumologic:
         failureThreshold: 3
 
   ### Traces configuration
-  ## Set the enabled flag to true to enable traces ingestion.
+  ## Set the enabled flag to false to disable traces ingestion.
   traces:
-    enabled: false
+    enabled: true
```
sumo-drosiek, line 257: @SumoLogic/open-source-collection-team WDYT?

deploy/helm/sumologic/values.yaml
```diff
@@ -177,7 +177,7 @@ sumologic:
           name: traces
           config-name: endpoint-traces
           properties:
-            content_type: Zipkin
+            content_type: OtlpHttpProto
```
perk-sumo, line 180: Is that a breaking change? We need to make sure it's not if that is to be included in v2.

perk-sumo (changes requested):
Hey, before we can proceed with this need confirmation that this will not break current users who have tracing enabled.

mat-rumian:
As we decided - PR postponed till 3.0

deploy/helm/sumologic/values.yaml
perk-sumo: Hey @mat-rumian should we call that change out in the changelog as well?

mat-rumian: Hey, I've reverted the change of the content_type.

swiatekm-sumo (approved):
LGTM

PR #2161, opened 2022-02-28
chore(otellogs): increase send_batch_size to 10240

sumo-drosiek:
increase it to prevent data loss
during reading huge file from beginning I observed some data loss
because previous value was relatively low


pmalek-sumo (approved):
Perhaps add a changelog entry?

PR #2173, opened 2022-03-04
feat: add pprof extension for otelcol

swiatekm-sumo:
##### Description

The performance cost of this extension is negligible, and it really helps with live profiling and debugging.

---

##### Checklist

- [x] Changelog updated

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -3643,6 +3643,7 @@ metadata:
         file_storage:
           directory: /var/lib/storage/otc
           timeout: 10s
+        pprof: {}
       exporters:
         ## Configuration for Sumo Logic Exporter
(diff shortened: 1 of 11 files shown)


sumo-drosiek (approved):

deploy/helm/sumologic/values.yaml
```diff
@@ -4273,7 +4276,7 @@ metadata:
         extensions:
           - health_check
           - file_storage
-          # - sumologic
```
pmalek-sumo, line 4276: I believe we can leave this comment here

swiatekm-sumo: If we do, I'd like to add an explanation to it, because without context it's just confusing.

pmalek-sumo: Agreed to the fact that this would need a bit more configuration anyway so it's fine to remove this comment.

PR #1647, opened 2021-06-10
Sumologic exporter and metric pipelines into OT Collector & agent

mat-rumian:
###### Description

Addition of the [Sumologic exporter](https://github.com/SumoLogic/opentelemetry-collector-contrib/tree/main/exporter/sumologicexporter) to the Tracing OT Collector pipeline. This will add possibility to send received metrics on the OTC receiver (OTLP protocol) side to Sumo. 

---

###### Testing performed

- [ ] Redeploy fluentd and fluentd-events pods
- [ ] Confirm events, logs, and metrics are coming in

PR #2182, opened 2022-03-14
fix(helm): always create default metrics source if traces are enabled

pmalek-sumo:
After introducing: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1647 we got into a bit of a situation for users who enable traces but don't enable metrics.

The problem's root cause is that traces pipeline now requires default metrics endpoint to be available https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/9bf45ca0304d7eaf690d900d90e52a54af1f6fd9/deploy/helm/sumologic/values.yaml#L3498 but doesn't take into account if metrics are enabled in the chart (which in turn controls which [sources will be created](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/9bf45ca0304d7eaf690d900d90e52a54af1f6fd9/deploy/helm/sumologic/values.yaml#L134-L180) which in turn controls what will be placed in [k8s secret ](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/9af6eb6d2d3064e34d3e4661e0311be1d2dd52e0/deploy/helm/sumologic/conf/setup/resources.tf#L32-L61).

This PR addresses that by always creating default metrics endpoint when traces are enabled in the chart.


deploy/helm/sumologic/conf/setup/resources.tf
```diff
@@ -29,13 +38,22 @@ resource "kubernetes_secret" "sumologic_collection_secret" {
...
-    {{- end }}
+          {{- end }}
+        {{- end }}
+      {{- else if and (eq $type "metrics") $ctx.sumologic.traces.enabled }}
+        {{- /*
+        If traces are enabled and metrics are disabled, create default metrics source anyway
+        */}}
+        {{- range $key, $source := $sources }}
+          {{- if eq $key "default"}}
+    {{ include "terraform.sources.data" (dict "Endpoint" (include "terraform.sources.config-map-variable" (dict "Type" $type "Context" $ctx "Name" $key)) "Name" (include "terraform.sources.name" (dict "Name" $key "Type" $type))) }}
```
sumo-drosiek, line 53: I think it would be more readable and less changes if we always create this endpoint (regardless of metrics and tracing configuration). If we have to add it for traces we can add it for logs as well and we just won't use it

PR #2184, opened 2022-03-15
fix: change filtering logs example to a real world one

perk-sumo:


pmalek-sumo (approved):

pmalek-sumo:
> I believe it should be moved to `BestPractices`. I would never expect anyone to look into this file 🤷

That's actually a fair statement 😅

PR #2201, opened 2022-03-17
feat(otelcol): add startupProbe config option

pmalek-sumo:


sumo-drosiek:
1. how can it be disabled?
2. add it to `README.md` please

pmalek-sumo:
1. with this approach you cannot but it's the same for liveness and readiness probes
2. added 

PTAL @sumo-drosiek

swiatekm-sumo (approved):

PR #2209, opened 2022-03-22
feat(notes): add information about tracing receiver endpoints

mat-rumian:
##### Description

Show available endpoints list to which traces can be sent after chart installation based on the installation type otelagent vs otelcol.

---

##### Checklist

- [x] Changelog updated

###### Testing performed


sumo-drosiek (approved):
It would be nice to add some kind of comment to otc service, that it should update `NOTES.txt` in case of changing ports ports, but generally LGTM

PR #2222, opened 2022-04-07
fix: set source name and category in the FluentD output for events

swiatekm-sumo:
##### Description
We've been setting source name and category directly on the source in the setup Terraform manifest. Instead, do it in the output plugin and make it configurable.

This makes it both easier to test and makes it work correctly if setup is disabled.

Also add a basic integration test for events.

---

##### Checklist

- [x] Changelog updated

###### Testing performed


swiatekm-sumo:
I'll update the changelog after we release 2.7. In the meantime, this is ready for review.

pmalek-sumo (changes requested):
Please add a changelog entry

sumo-drosiek (approved):

deploy/helm/sumologic/values.yaml
```diff
@@ -909,8 +908,10 @@ fluentd:
     ## Override output section for events. Leave empty for the default output section
     overrideOutputConf: |-
 
+    ## Source name for the Events source. Default: "events"
+    sourceName: "events"
     ## Source category for the Events source. Default: "{clusterName}/events"
-    sourceCategory: ""
+    # sourceCategory: "clustername/events"
```
sumo-drosiek, line 914: `"clustername/events"` can be misleading as this is not the actual default

swiatekm-sumo: What would you prefer, `<cluster_name>/events`?

sumo-drosiek: I think `kubernetes/events`

deploy/helm/sumologic/conf/events/events.conf
```diff
@@ -50,6 +50,8 @@
     @id sumologic.endpoint.events
     sumo_client {{ include "sumologic.sumo_client" . | quote }}
     endpoint "#{ENV['SUMO_ENDPOINT_DEFAULT_EVENTS_SOURCE']}"
+    source_name {{ .Values.fluentd.events.sourceName }}
+    source_category {{ .Values.fluentd.events.source_category | default (printf "%s/%s"  (include "sumologic.clusterNameReplaceSpaceWithDash" .) (.Values.fluentd.events.sourceName ))}}
```
pmalek-sumo, line 54: Usage of `sumologic.clusterNameReplaceSpaceWithDash` I believe should be documented somewhere, shouldn't it?

pmalek-sumo (approved):

deploy/helm/sumologic/conf/events/events.conf
swiatekm-sumo: It should, and we use it basically everywhere we need the cluster name for something. There's even a warning from setup for it I believe. I've added a comment about it to the README and all the other docs explicitly mentioning it.

PR #2232, opened 2022-04-12
docs: FluentD buffer size configuration

swiatekm-sumo:
##### Description

Make it clear that the FluentD buffer size needs to be increased alongside the volume size.

---

##### Checklist

- [ ] Changelog updated


perk-sumo (approved):
🚀

PR #2238, opened 2022-04-14
fix: use custom ServiceMonitor for Prometheus' own metrics

astencel-sumo:
Having the `kube-prometheus-stack` subchart generate the ServiceMonitor to scrape Prometheus' own metrics does not work correctly in a setup with an external Prometheus operator in a different (higher) version.
By "does not work correctly" I mean that the ServiceMonitor points to a Service that does not see any pods and in effect, Prometheus cannot scrape its own metrics.
This is because different versions of Prometheus operator create different labels on Prometheus pods.
To fix this, let's disable the `kube-prometheus-stack` subchart's built-in ServiceMonitor and create our own, just like we do for other collection resources (Fluent Bit, Fluentd, Otelcol).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- fix: use custom ServiceMonitor for Prometheus' own metrics [#2238]
+
 ### Fixed
 
 - fix: set source name and category in the FluentD output for events [#2222][#2222]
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):

swiatekm-sumo (commented):
Can you add these metrics to https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/c7255e41afd9c1733e23e78b012f2efb79f05bd0/tests/integration/internal/constants.go#L38 so integration tests verify their presence?

kkujawa-sumo (approved):

PR #2251, opened 2022-04-28
feat(metadata/otc): upgrade to v0.50.0-sumo-0

sumo-drosiek:
##### Description

- Upgrade otc version to [v0.49.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/releases/tag/v0.49.0-sumo-0)
- Remove usage of `metadata_attributes`
- Stop moving out record attributes for systemd
- Move `_collector` to resource attributes for containers
- remove `k8s.pod.uid` for vagrant
- update docs for using containerd and fluent-bit 
- remove `http_listener_v2_path` in routing processor instead of exporter for metadata/metrics

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - chore: change Fluent Bit image to `public.ecr.aws/sumologic/fluent-bit:1.6.10-sumo-2`,
   it is Fluent Bit 1.6.10 with updated dependencies,
   image repository: https://github.com/SumoLogic/fluent-bit-docker-image [#2254][#2254]
-- feat(metadata/otc): upgrade to v0.49.0-sumo-0 [#2251][#2251]
+- feat(metadata/otc): upgrade to v0.50.0-sumo-0 [#2251][#2251]
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - chore: change Fluent Bit image to `public.ecr.aws/sumologic/fluent-bit:1.6.10-sumo-2`,
   it is Fluent Bit 1.6.10 with updated dependencies,
   image repository: https://github.com/SumoLogic/fluent-bit-docker-image [#2254][#2254]
+- feat(metadata/otc): upgrade to v0.49.0-sumo-0 [#2251][#2251]
 
 ### Fixed
 
@@ -36,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2246]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2246
 [#2254]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2254
 [#2255]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2255
+[#2251]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2251
 [Unreleased]: https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v2.6.0...main
 
 ## [v2.7.0][v2.7.0]
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -3543,10 +3543,6 @@ otelcol:
         source_name: "exporters.sumologic.source_name.replace"
         ## Desired host name. Useful if you want to override the source host configured for the source.
         source_host: "%{host.name}"
-        ## List of regexes for attributes which should be send as metadata
-        metadata_attributes:
-          - k8s.*
-          - host.name
         ## Timeout for every attempt to send data to Sumo Logic backend. Maximum connection timeout is 55s.
         timeout: 5s
         retry_on_failure:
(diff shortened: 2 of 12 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -4279,11 +4229,8 @@ metadata:
                   value: kubelet.service
         groupbyattrs/systemd:
           keys:
-            - _SYSTEMD_UNIT
-            - _HOSTNAME
-            - SYSLOG_FACILITY
-            - PRIORITY
```
swiatekm-sumo, line 4285: These are used in the `exclude` section of `sourceprocessor`, so we can't just remove them. I replaced them with filter processors here: https://github.com/SumoLogic/sumologic-kubernetes-collection/commit/b9a4875ad6b9e0f2a497db9a41ba52592394dc43

swiatekm-sumo (commented):
We also need to get rid of `annotations:` blocks from k8s_tagger configurations. These were unnecessary before, but now they're going to cause us to send more metadata than we need.

sumo-drosiek:
> We also need to get rid of annotations: blocks from k8s_tagger configurations. These were unnecessary before, but now they're going to cause us to send more metadata than we need.

isn't breaking change?

deploy/helm/sumologic/values.yaml
sumo-drosiek: I cherry-picked your commit

sumo-drosiek:
> > > We also need to get rid of annotations: blocks from k8s_tagger configurations. These were unnecessary before, but now they're going to cause us to send more metadata than we need.
> > 
> > 
> > isn't this a breaking change?
> 
> It is if the user added them to their `metadata_attributes`. In that sense, almost all changes in this PR are breaking.

I was thinking rather that `we have been sending pod annotations and we stop sending them with this change`

swiatekm-sumo:
> I was thinking rather that we have been sending pod annotations and we stop sending them with this change

We haven't, though. We didn't have the annotations in `metadata_attributes` anywhere, so it didn't get sent.

sumo-drosiek:
> > I was thinking rather that we have been sending pod annotations and we stop sending them with this change
> 
> We haven't, though. We didn't have the annotations in `metadata_attributes` anywhere, so it didn't get sent.

We are using `json`, so if the annotations were in the any of the attributes, they were send as part of log

swiatekm-sumo:
> > > I was thinking rather that we have been sending pod annotations and we stop sending them with this change
> > 
> > 
> > We haven't, though. We didn't have the annotations in `metadata_attributes` anywhere, so it didn't get sent.
> 
> We are using `json`, so if the annotations were in the any of the attributes, they were send as part of log

If `metadata_attributes` were defined, then only attributes that matched any of the entries were sent, to my best understanding. See: https://github.com/SumoLogic/sumologic-otel-collector/blob/9923b89a506617689fcec94c1f85c0a3e7770c86/pkg/exporter/sumologicexporter/exporter.go#L292.

swiatekm-sumo:
Ok, you're talking about record attributes, makes sense. In that sense it is a breaking change and quite annoying to deal with, as I don't think there's a processor to move attributes from the resource level to the record level.

sumo-drosiek:
also we use `pod_annotations` in sourceprocessor in order to overwrite `_source*` configuration

swiatekm-sumo:
> also filtering needs to be fixed as "" regexp catches ecerything

I'm pretty sure that's the same behaviour as exclusion in `sourceprocessor`. Both use `MatchString`:
https://github.com/SumoLogic/sumologic-otel-collector/blob/a825c0c60c4bf37e1abbdb42952eccdeaa6d397f/pkg/processor/sourceprocessor/source_processor.go#L344
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/033bc949bec36233c6c4a050a951f119dd00b16d/internal/coreinternal/processor/filterset/regexp/regexpfilterset.go#L64

sumo-drosiek:
@swiatekm-sumo for `source_processor` we iterate over empty list, so we do not execute any regexp

https://github.com/SumoLogic/sumologic-otel-collector/blob/a825c0c60c4bf37e1abbdb42952eccdeaa6d397f/pkg/processor/sourceprocessor/source_processor.go#L140-L145

https://github.com/SumoLogic/sumologic-otel-collector/blob/a825c0c60c4bf37e1abbdb42952eccdeaa6d397f/pkg/processor/sourceprocessor/source_processor.go#L97-L101

swiatekm-sumo (changes requested):
I added dropping the Pod annotation from resource attributes in [fix: drop Pod annotation attributes for container logs](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2251/commits/8b1f36834f2d002be8cc7581c22928e9d5b647c9), the rest looks good to me.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - chore: change Fluent Bit image to `public.ecr.aws/sumologic/fluent-bit:1.6.10-sumo-2`,
   it is Fluent Bit 1.6.10 with updated dependencies,
   image repository: https://github.com/SumoLogic/fluent-bit-docker-image [#2254][#2254]
+- feat(metadata/otc): upgrade to v0.49.0-sumo-0 [#2251][#2251]
```
sumo-drosiek, line 22:
```suggestion
- feat(metadata/otc): upgrade to v0.50.0-sumo-0 [#2251][#2251]
```

swiatekm-sumo (approved):

PR #2253, opened 2022-04-28
doc: add information about Fluentd autoscaling to troubleshooting documentation

kkujawa-sumo:


deploy/docs/Troubleshoot_Collection.md
```diff
@@ -94,6 +99,18 @@ If it is, there will be `budget.exceeded` messages from Sumo in Fluentd or Otelc
 
 [ingest_budgets]: https://help.sumologic.com/Manage/Ingestion-and-Volume/Ingest_Budgets
 
+### Check Fluentd autoscaling
```
astencel-sumo, line 102: Should we mention Otelcol as Fluentd replacement too?

astencel-sumo (approved):

deploy/docs/Troubleshoot_Collection.md
kkujawa-sumo: We should also write documentation about Otelcol autoscaling, I don't see any documentation about this. 
I prefer to extend this documentation in the next PR.

PR #2275, opened 2022-05-06
chore(fluent-bit): upgrade to 1.8.15

sumo-drosiek:

##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - chore: update falco chart to 1.18.6 [#2376]
 - chore: update telegraf-operator chart to 1.3.5 [#2387]
 - feat: update otellogs to 0.52.0-sumo-0 [#2338][#2338]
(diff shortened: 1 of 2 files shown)


pdelewski (approved):

astencel-sumo (changes requested):
This version of Fluent Bit crashes on configurations that 1.6.10 works fine with. Are we sure we want to do this?

sumo-drosiek:
What configuration?

ReillyBrogan:
Might as well use 1.9.7 instead now. Hopefully this can be updated given that 1.6.10 is now more than a year and a half old.

swiatekm-sumo:
@sumo-drosiek do we still want to do this? We ran into problems every time we tried to upgrade, and at this point it's probably easier to use otel.

sumo-drosiek:
> @sumo-drosiek do we still want to do this? We ran into problems every time we tried to upgrade, and at this point it's probably easier to use otel.

I doubt about fluent-bit stability. `1.6.10` was the most stable so far. I believe we can close it for now and recommend to use OpenTelemetry Collector. Also there is nothing which can stop customers from using newer version

PR #2281, opened 2022-05-13
Making replica count for fluentd events configurable

gizmo-rt:
##### Description

Replica count for fluentd events currently cannot be driven by config. Hence, manual scaling is required to enable them.
This PR sets the default replica count for events to 1 and enables driving the replica requirement via config

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


swiatekm-sumo:
If you actually do this, you'd have each of the replicas collect the same set of data. Is that useful for anything?

PR #2284, opened 2022-05-17
fix: set cluster field in metadata pipelines

swiatekm-sumo:
##### Description

The `cluster` metadata field was the only field set on the collector in Sumo, rather than in metadata pipelines. In the metadata pipelines, how it was handled varied a lot:
* FluentD metrics set it correctly.
* FluentD logs didn't set it at all.
* OT pipelines were actively deceptive, pretending to set it in the k8s_tagger processor, which doesn't actually do anything. 

The K8s metadata field that allows the k8s_tagger to set the cluster name is being deprecated in K8s upstream due to being actively confusing and not very useful. See: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/9886 for more context, and also https://github.com/SumoLogic/sumologic-otel-collector/pull/578.

Additionally, setting the field on the collector actually behaves subtly differently from sending it as a metric label, which caused one nasty bug for a client trialing OT.

This change explicitly sets the field based on the configuration value and removes it from the collector.

---

##### Checklist

- [x] Changelog updated

###### Testing performed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- fix: set cluster field in metadata pipelines [#2284][#2284]
+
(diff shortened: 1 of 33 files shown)


swiatekm-sumo:
I noticed I also need to add this for FluentD events, which use the log source. Tagging this `don't merge` until that's done.

astencel-sumo (approved):

PR #2287, opened 2022-05-23
chore(deps): upgrade fluentd to 1.14.6-sumo-3

swiatekm-sumo:
##### Description

The new version reloads ServiceAccount tokens periodically, correctly handling token expiration enabled in K8s 1.21.

This fixes #2280 

---

##### Checklist

- [X] Changelog updated

###### Testing performed

- [X] Redeploy fluentd and fluentd-events pods
- [X] Confirm events, logs, and metrics are coming in

PR #2289, opened 2022-05-24
docs(changelog): fix invalid entry for #2287

swiatekm-sumo:
I accidentally added the changelog entry for #2287 to v2.8.0 instead of the unreleased section.


astencel-sumo (approved):
😄

PR #2291, opened 2022-05-24
fix(otellogs): set resources on Otelcol logs collector daemonset

astencel-sumo:
The resources were already defined in `values.yaml`
in the `otellogs.daemonset.resources` property,
but this property was not being used in the daemonset template.

PR #2298, opened 2022-05-25
fix(otellogs): set resources on Otelcol logs collector daemonset

astencel-sumo:
The resources were already defined in `values.yaml` in the `otellogs.daemonset.resources` property, but this property was not being used in the daemonset template.

Backports #2291 to `v2.8`.

### DIFF
--- a/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
+++ b/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
@@ -64,6 +64,8 @@ spec:
           httpGet:
             path: /
             port: 13133 # Health Check extension default port.
+        resources:
+          {{- toYaml .Values.otellogs.daemonset.resources | nindent 10 }}
         volumeMounts:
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
```diff
@@ -64,6 +64,8 @@ spec:
           httpGet:
             path: /
             port: 13133 # Health Check extension default port.
+        resources:
+          {{- toYaml .Values.otellogs.daemonset.resources | nindent 10 }}
```
pdelewski, line 68: What this 10 means in this case?

pdelewski (approved):

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
astencel-sumo: "Indent 10 spaces" 🤷

PR #2364, opened 2022-06-07
feat(otellogs/systemd): add support for systemd logs to otellogs

sumo-drosiek:
##### Description

Add support for journald receiver. This behaves like fluent-bit with one small exception. Units for filtering are taken from different place.

```
        ## This is not a full equivalent of fluent-bit filtering as fluent-bit filters by `_SYSTEMD_UNIT`
        ## Here is filtering by `UNIT`
```

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -4138,6 +4138,9 @@ metadata:
           json_logs:
             add_timestamp: true
             timestamp_key: timestamp
+            ## use flatten_body, but OTLP won't require any flattening
+            ## fluent based logs will be all send as record attributes
+            ## otellogs based logs will be all send as body attributes
             flatten_body: true
           endpoint: ${SUMO_ENDPOINT_DEFAULT_LOGS_SOURCE}
           source_name: "%{_sourceName}"
@@ -4386,6 +4389,11 @@ metadata:
           source_category: '{{ .Values.fluentd.logs.systemd.sourceCategory | quote }}'
           source_category_prefix: '{{ .Values.fluentd.logs.systemd.sourceCategoryPrefix | quote }}'
           source_category_replace_dash: '{{ .Values.fluentd.logs.systemd.sourceCategoryReplaceDash | quote }}'
+        ## Remove all attributes, so body won't by nested by SumoLogic receiver in case of using otlp format
+        transform/remove_attributes:
+          logs:
+            queries:
+              - limit(attributes, 0)
 
         ## kubelet related processors
         filter/include_kubelet:
@@ -4450,10 +4458,9 @@ metadata:
           #     - batch
           #   exporters:
           #     - sumologic/containers
-          logs/systemd:
+          logs/fluent/systemd:
             receivers:
               - fluentforward
-              - otlp
             processors:
               - memory_limiter
               - filter/include_fluent_tag_host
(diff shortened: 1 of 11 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -4145,6 +4145,22 @@ metadata:
           sending_queue:
             enabled: true
             persistent_storage_enabled: '{{ .Values.metadata.persistence.enabled }}'
+        ## This is the same like sumologic/systemd with additional `flatten_body` parameter
+        sumologic/systemd_flatten:
+          log_format: json
+          json_logs:
+            add_timestamp: true
+            timestamp_key: timestamp
+            flatten_body: true
```
sumo-drosiek, line 4154: we can consider it for fluentd pipeline as well, but we need further investigation for that

sumo-drosiek: Question is if the systemd log can be nested json

tests/helm/logs_otc_daemonset/static/complex.output.yaml
```diff
@@ -0,0 +1,144 @@
...
+        name: file-storage
+      - hostPath:
+          path: /var/log/journal/
+          type: ""
+        name: varlogjournal
+      - name: es-certs
+        secret:
+          defaultMode: 420
+          secretName: es-certs
+        serviceAccountName: RELEASE-NAME-sumologic-otelcol-logs-collector
```
sumo-drosiek, line 144:
```suggestion
        serviceAccountName: RELEASE-NAME-sumologic-otelcol-logs-collector

```

deploy/helm/sumologic/values.yaml
sumo-drosiek: https://stackoverflow.com/questions/45604563/injecting-structured-json-logs-into-journald

```
journald doesn't accept arbitary JSON. Just Key/Value pairs. So it's not possible to send nested data structures
```

deploy/helm/sumologic/values.yaml
```diff
@@ -4144,7 +4145,7 @@ metadata:
           ## ref: https://github.com/open-telemetry/opentelemetry-collector/tree/release/v0.37.x/exporter/exporterhelper#configuration
           sending_queue:
             enabled: true
-            persistent_storage_enabled: '{{ .Values.metadata.persistence.enabled }}'
+            persistent_storage_enabled: '{{ .Values.metadata.persistence.enabled }}'\
```
sumo-drosiek, line 4148:
```suggestion
            persistent_storage_enabled: '{{ .Values.metadata.persistence.enabled }}'
```

deploy/helm/sumologic/values.yaml
```diff
@@ -4136,6 +4136,7 @@ metadata:
           json_logs:
             add_timestamp: true
             timestamp_key: timestamp
+            flatten_body: true
```
sumo-drosiek, line 4139: systemd is key value pair, so this option is safe for fluentd. It's necessary for otel, as it sends body as key-value pair. This body would be put inside `log` key without this option

deploy/helm/sumologic/values.yaml
```diff
@@ -4488,6 +4497,53 @@ metadata:
               - batch
             exporters:
               - sumologic/systemd
+          ## This is the same pipeline like logs/fluent/systemd with the following changes:
```
sumo-drosiek, line 4500:
```suggestion
          ## This is the same pipeline like logs/fluent/systemd, but with the following changes:
```

deploy/helm/sumologic/values.yaml
```diff
@@ -4488,6 +4497,53 @@ metadata:
...
+              - attributes/extract_systemd_source_fields
+              - attributes/remove_fluent_tag
+              - groupbyattrs/systemd
+              - resource/add_cluster
+              - source/systemd
+              - transform/remove_attributes
+              - batch
+            exporters:
+              - sumologic/systemd
+          ## This is the same pipeline like logs/fluent/kubelet with the following changes:
```
sumo-drosiek, line 4524:
```suggestion
          ## This is the same pipeline like logs/fluent/kubelet, but with the following changes:
```

astencel-sumo:
"I have a dream"... that we have a piece of high-level documentation describing collecting systemd logs, a bit like what we now have for collecting Kubernetes events: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/docs/collecting-kubernetes-events.md.

deploy/helm/sumologic/values.yaml
```diff
@@ -5014,6 +5078,59 @@ otellogs:
           - type: move
             from: body.log
             to: body
+      journald:
+        directory: /var/log/journal
+        ## This is not a full equivalent of fluent-bit filtering as fluent-bit filters by `_SYSTEMD_UNIT`
+        ## Here is filtering by `UNIT`
+        units:
```
astencel-sumo, line 5085: Does this list mean that if the user wants to customize this list, they just need to overwrite this property in their `values.yaml` file? 😱  For example, if they want to exclude a specific unit, they need to replicate this whole list with the one unit excluded.
I don't know what the common use cases are (if there are any). Maybe there are no good reasons to customize this list. In that case, do we need to expose it?

sumo-drosiek:
> "I have a dream"... that we have a piece of high-level documentation describing collecting systemd logs, a bit like what we now have for collecting Kubernetes events: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/docs/collecting-kubernetes-events.md.

I would love to have this simple option:

```
sumologic:
  logs:
    systemd:
      enabled: false
```

but it is not possible now. We need to take it in `v3`. I think the documentation should be a part of that change too. (For now we do not document any otellogs configuration)

If you think differently I can make some documentation draft as part of this PR :)

sumo-drosiek:
> but it is not possible now

TBH, maybe it is, but we need to have some final consensus about which approach to configuration we take

deploy/helm/sumologic/values.yaml
sumo-drosiek: Yes, this PR is fluent-bit replacement. There is a lot of improvements we can take, but maybe there should be separate issues, so it will be more organised way to track them?

sumo-drosiek: > I don't know what the common use cases are (if there are any).

I don't think there are any, but we need to have some plan in case they'll appear

astencel-sumo (approved):

PR #2379, opened 2022-06-15
feat: add experimental OT event collection

swiatekm-sumo:
##### Description

This adds event collection using OT. As we don't have an `events` subsection in the `sumologic` section, I can't implement a simple switch like we have for metadata, and instead need to add a new section and depend on the user not to enable both at once.

One thing of note is the approach to configuring OT. Instead of simply embedding the configuration in values.yaml, we instead treat it as a normal template, and then allow overriding it via a configuration field whose function will be similar to the current approach, but which is intended as a last resort escape hatch rather than the default way of configuring OT.

TODO:
- [x] documentation
- [x] comments for the more interesting features

---

##### Checklist

- [x] Changelog updated

###### Testing performed


swiatekm-sumo:
This depends on https://github.com/SumoLogic/sumologic-otel-collector/pull/620, and therefore probably on the 0.54.0 OT distro release.

deploy/helm/sumologic/templates/events/otelcol/statefulset.yaml
```diff
@@ -0,0 +1,166 @@
...
+{{- end }}
+      labels:
+        app: {{ template "sumologic.labels.app.events.pod" . }}
+        {{- include "sumologic.labels.common" . | nindent 8 }}
+{{- if .Values.sumologic.podLabels }}
+{{ toYaml .Values.sumologic.podLabels | indent 8 }}
+{{- end }}
+{{- if .Values.otelevents.statefulset.podLabels }}
+{{ toYaml .Values.otelevents.statefulset.podLabels | indent 8 }}
+{{- end }}
```
sumo-drosiek, line 34: It would be nice to test this configurations as well in helm tests

sumo-drosiek:
I would insist to test all possible values for `statefulset` (including `sumologic` key and so on)

deploy/helm/sumologic/templates/events/otelcol/statefulset.yaml
swiatekm-sumo: Sure, though I think we don't test this for any OT statefulset 😅

sumo-drosiek: We should, as this ends with fix PR afterwards :dagger:

sumo-drosiek (approved):

sumo-drosiek:
We can consider `provider` approach, to avoid both otelevents and fluentd being run at the same time

swiatekm-sumo:
> We can consider `provider` approach, to avoid both otelevents and fluentd being run at the same time

That's a breaking change though, as events don't have a section under `sumologic`. We can switch to that in 3.0.

sumo-drosiek:
it is not until `fluentd` will be default choice

swiatekm-sumo:
> it is not until `fluentd` will be default choice

It is. If someone disables events by setting `fluentd.events.enabled` to `false`, and I add `sumologic.events.enabled` defaulting to `true`, it's either enabled, in which case it's a breaking change, or not, in which case we have a situation very similar to right now.

sumo-drosiek:
I was thinking rather about flag `sumologic.events.provider`

swiatekm-sumo:
> I was thinking rather about flag `sumologic.events.provider`

Does that actually help? We run into problem if someone enabled both FluentD and OT events. In that case, `sumologic.events.provider` could be a tiebreaker, but it's confusing that you need to set this *and* separately enable the thing you want.

sumo-drosiek:
we could enable otc events by default and control it by `sumologic.events.provider` only, especially if we plan to have this flag in the future. If not, we can skip it

PR #2388, opened 2022-06-21
chore(docs): clean up the configuration table

swiatekm-sumo:
##### Description

The configuration table in the Chart README has become unmanagably large, primarily due to large values in the defaults column. Change the largest of those to "See values.yaml", and trim down some long descriptions.


sumo-drosiek:
Could you link `values.yaml`?
Also I think that we should auto-generate the table in the future :fearful:

swiatekm-sumo:
> > I can link to the file, but linking to specific lines is too brittle imo.
> 
> agree, do not expect it :)

Done.

PR #2390, opened 2022-06-24
chore(tracing): move k8s_tagger to otelagent

mat-rumian:
##### Description

This PR:
* moves `k8s_tagger` processor from `otelcol` to `otelagent`. 
* enable `otelagent` and `otelgateway` by default when `traces.enabled=true`

---

##### Checklist

- [x] Changelog updated

###### Testing performed


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+This release changes the OpenTelemetry Collector `Traces` service endpoint `CHART_NAME-sumologic-otelcol.NAMESPACE`
+to `deprecated`. Service will still work and point to `CHART_NAME-sumologic-otelagent.NAMESPACE`.
```
sumo-drosiek, line 11: so we deprecate it, but still use it?

sumo-drosiek: `otelcol` vs `otelagent`

sumo-drosiek (commented):
I believe we can also add to release information that from now, `otelagent` and `otelgateway` are going to be enable by default. It would be also nice to have info how to disable it in the docs somewhere

sumo-drosiek (approved):
traces are disabled by default, so I'm fine :)

CHANGELOG.md
mat-rumian: yes, the reason is that there are customers which deployed tracing without `otelagent`s, so to avoid issues with missing data/export errors we routed traffic to `otelagent` daemonset using `otelcol` service.

PR #2407, opened 2022-07-05
feat: adjust event collection configuration

astencel-sumo:
This pull request builds on event collection with OpenTelemetry Collector introduced in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2379, but changes how event collection is configured. It does so by removing the property `otelevents.enabled` and instead adding `sumologic.events.enabled` and `sumologic.events.provider`.

Also adding some more documentation around this and fixing some typos or misleading docs.

For easier review, see individual commits.


deploy/helm/sumologic/values.yaml
```diff
@@ -191,6 +191,10 @@ sumologic:
           properties:
             content_type: Zipkin
 
+  ### Configuration for collection of Kubernetes events
+  events:
+    provider: fluentd
+
```
swiatekm-sumo, line 197: Shouldn't we also have `enabled: true` here?

astencel-sumo: It would be good to have it, but we cannot - for backward compatibility with the `fluentd.events.enabled` property.

What we want is for the `fluentd.events.enabled` property to enable/disable events collection with Fluentd as long as the `sumologic.events.provider` is not set to `otelcol` and the `sumologic.events.enabled` property is not set by the user.

sumo-drosiek: this is how imho it should work

sumo-drosiek: But we can use it to manage otelcol beahvior

sumo-drosiek: if unset is equal to `tru` why not to set it?

astencel-sumo: Here's how I see it:

1. Default setup

```yaml
sumologic:
  accessId: xxx
  accessKey: yyy
```

Result: Events enabled, using Fluentd

2. Events disabled by the user using the deprecated property `fluentd.events.enabled`:

```yaml
fluentd:
  events:
    enabled: false
sumologic:
  accessId: xxx
  accessKey: yyy
```

Result: Events disabled

3. Events disabled by the user using the newly introduced property `sumologic.events.enabled`:

```yaml
sumologic:
  accessId: xxx
  accessKey: yyy
  events:
    enabled: false
```

Result: Events disabled

4. Events provider changed to Otelcol:

```yaml
sumologic:
  accessId: xxx
  accessKey: yyy
  events:
    provider: otelcol
```

Result: Events enabled, using Otelcol

PR #2414, opened 2022-07-08
docs: add information about installation on OpenShift 4.9

kkujawa-sumo:
docs: add information about installation on OpenShift 4.9
docs: fix command to install on OpenShift with existing Prometheus - for normal k8s cluster kube-prometheus-stack is disabled


astencel-sumo (approved):

kkujawa-sumo:
> be honest, I don't like our current docs on installing Prometheus "side by side" or installing on OpenShift at all.
> I believe we need to write those from scratch.

I totally agree :(

PR #2436, opened 2022-07-15
feat: refactor OT log collector configuration

swiatekm-sumo:
##### Description

Change how log collection using OT is configured. The idea here is for the purely functional configuration to live under the `sumologic.logs` key and contain no application-specific configuration, and for the K8s specific and application specific (like raw OT configuration) configuration to be separate - currently I've left it under `otellogs`, but I'm open to alternatives.

TODO:
- [x] Add template tests for the new options
- [ ] setting for tailling additional files
- [x] documentation
---

##### Checklist

- [x] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/docs/opentelemetry_collector.md
+++ b/deploy/docs/opentelemetry_collector.md
@@ -1,6 +1,6 @@
-# Opentelemetry Collector
+# OpenTelemetry Collector
 
-Opentelemetry Collector is a software to receive, process and export logs, metrics and traces.
+OpenTelemetry Collector is a software to receive, process and export logs, metrics and traces.
 We offer it as drop-in replacement for Fluentd in our collection.
 
 > :warning: **This feature is currently in beta and its configuration can change. It is nonetheless production-ready and will become the default in the next major version.**
@@ -14,11 +14,11 @@ We offer it as drop-in replacement for Fluentd in our collection.
   - [SystemD Logs](#systemd-logs)
 - [Persistence](#persistence)
   - [Enabling persistence](#enabling-persistence)
-    - [Enabling Opentelemetry Collector persistence by recreating StatefulSet](#enabling-opentelemetry-collector-persistence-by-recreating-statefulset)
-    - [Enabling Opentelemetry Collector persistence by creating temporary instances and removing earlier created](#enabling-opentelemetry-collector-persistence-by-creating-temporary-instances-and-removing-earlier-created)
+    - [Enabling OpenTelemetry Collector persistence by recreating StatefulSet](#enabling-opentelemetry-collector-persistence-by-recreating-statefulset)
+    - [Enabling OpenTelemetry Collector persistence by creating temporary instances and removing earlier created](#enabling-opentelemetry-collector-persistence-by-creating-temporary-instances-and-removing-earlier-created)
   - [Disabling persistence](#disabling-persistence)
-    - [Disabling Opentelemetry Collector persistence by recreating StatefulSet](#disabling-opentelemetry-collector-persistence-by-recreating-statefulset)
-    - [Disabling Opentelemetry Collector persistence by creating temporary instances nd removing earlier created](#disabling-opentelemetry-collector-persistence-by-creating-temporary-instances-nd-removing-earlier-created)
+    - [Disabling OpenTelemetry Collector persistence by recreating StatefulSet](#disabling-opentelemetry-collector-persistence-by-recreating-statefulset)
+    - [Disabling OpenTelemetry Collector persistence by creating temporary instances nd removing earlier created](#disabling-opentelemetry-collector-persistence-by-creating-temporary-instances-nd-removing-earlier-created)
 - [Traces](#traces)
   - [Load balancing using the gateway](#load-balancing-using-the-gateway)
 - [Kubernetes Events](#kubernetes-events)
@@ -26,9 +26,9 @@ We offer it as drop-in replacement for Fluentd in our collection.
 
 ## Metrics
 
-We are using Opentelemetry Collector like Fluentd to enrich metadata and to filter data.
+We are using OpenTelemetry Collector like Fluentd to enrich metadata and to filter data.
 
-To enable Opentelemetry Collector for metrics, please use the following configuration:
+To enable OpenTelemetry Collector for metrics, please use the following configuration:
 
 ```yaml
 sumologic:
@@ -39,20 +39,20 @@ sumologic:
 
 As we are providing drop-in replacement, most of the configuration from
 [`values.yaml`][values] should work
-the same way for Opentelemetry Collector like for Fluentd.
+the same way for OpenTelemetry Collector like for Fluentd.
 
 ### Metrics Configuration
 
-All Opentelemetry Collector configuration for metrics is located in
+All OpenTelemetry Collector configuration for metrics is located in
 [`values.yaml`][values] as `metadata.metrics.config`.
 
-If you want to modify it, please see [Sumologic Opentelemetry Collector configuration][configuration]
+If you want to modify it, please see [Sumologic OpenTelemetry Collector configuration][configuration]
 for more information.
 
 ## Logs
 
-Opentelemetry Collector can be used for both log collection and metadata enrichment. For these roles,
-it replaces respectively Fluent-Bit and FluentD.
+OpenTelemetry Collector can be used for both log collection and metadata enrichment. For these roles,
+it replaces respectively Fluent Bit and Fluentd.
 
 For log collection, it can be enabled by setting:
 
@@ -67,7 +67,7 @@ fluent-bit:
   enabled: false
 ```
 
-> **NOTE** Fluent-Bit must be disabled for Opentelemetry Collector to be enabled, they are mutually exclusive.
+> **NOTE** Fluent Bit must be disabled for OpenTelemetry Collector to be enabled, they are mutually exclusive.
 
 For metadata enrichment, it can be enabled by setting:
 
@@ -78,12 +78,12 @@ sumologic:
       provider: otelcol
 ```
 
-If you haven't modified the FluentD or Fluent-Bit configuration, this should be a drop-in replacement with no
+If you haven't modified the Fluentd or Fluent Bit configuration, this should be a drop-in replacement with no
 further changes required.
 
 ### Logs Configuration
 
-High level Opentelemetry Collector configuration for logs is located in
+High level OpenTelemetry Collector configuration for logs is located in
 [`values.yaml`][values] under the `sumologic.logs` key.
 
 Configuration specific to the log collector DaemonSet can be found under the `otellogs` key.
@@ -92,7 +92,7 @@ Finally, configuration specific to the metadata enrichment StatefulSet can be fo
 
 In both of the aforementioned cases, the raw configuration can be overridden - this is done respectively by using
 the `otellogs.config.override` and `metadata.logs.config` sections. Only use these if your use case isn't covered
-by the high-level settings. See [Sumologic Opentelemetry Collector configuration][configuration]
+by the high-level settings. See [Sumologic OpenTelemetry Collector configuration][configuration]
 for more information
 
 [configuration]: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/Configuration.md
@@ -107,7 +107,7 @@ sumologic:
   logs:
     multiline:
       enabled: true
-      first_line_regex: "^\\[?\\d{4}-\\d{1,2}-\\d{1,2}.\\d{2}:\\d{2}:\\d{2}.*"
+      first_line_regex: "^\\[?\\d{4}-\\d{1,2}-\\d{1,2}.\\d{2}:\\d{2}:\\d{2}"
 ```
 
 where `first_line_regex` is a regular expression used to detect the first line of a multiline log.
@@ -147,7 +147,7 @@ sumologic:
 
 ## Persistence
 
-The persistence for Opentelemetry Collector can be configured in [`values.yaml`][values] by making changes under the `metadata.persistence`:
+The persistence for OpenTelemetry Collector can be configured in [`values.yaml`][values] by making changes under the `metadata.persistence`:
 
 ```yaml
 metadata:
@@ -158,15 +158,15 @@ metadata:
 along with changes in configuration under `metadata.metrics.config` and `metadata.logs.config`
 according to [Persistent Queue][persistent_queue] documentation.
 
-When Opentelemetry Collector persistence is to be changed (enabled or disabled)
-it is required to recreate or delete existing Opentelemetry Collector StatefulSets,
+When OpenTelemetry Collector persistence is to be changed (enabled or disabled)
+it is required to recreate or delete existing OpenTelemetry Collector StatefulSets,
 as it is not possible to add/remove `volumeClaimTemplate` for StatefulSet.
 
 [persistent_queue]: https://github.com/open-telemetry/opentelemetry-collector/tree/release/v0.37.x/exporter/exporterhelper#persistent-queue
 
 ### Enabling persistence
 
-To enable persistence for Opentelemetry Collector set following configuration in [`values.yaml`][values]:
+To enable persistence for OpenTelemetry Collector set following configuration in [`values.yaml`][values]:
 
 ```yaml
 metadata:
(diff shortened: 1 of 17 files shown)


deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
```diff
@@ -1 +1,245 @@
...
+      ## Output Body (JSON): { "log": "2001-02-03 04:05:06 first line\n  second line\n  third line\n", "stream": "stdout" }
+{{- if .Values.sumologic.logs.multiline.enabled }}
+      - id: merge-multiline-logs
+        type: recombine
+        output: extract-metadata-from-filepath
+        source_identifier: attributes["log.file.path"]
+        combine_field: body.log
+        combine_with: ""
+        is_first_entry: body.log matches {{ .Values.sumologic.logs.multiline.start_regex | quote }}
+{{- end }}
```
sumo-drosiek, line 131: I would reconsider supporting multiple mutltiline detections in this PR as this is common use case for customers. I can also create Issue if we prefer to add it in the future

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
```diff
@@ -1 +1,245 @@
...
+        from: attributes.uid
+        to: attributes["k8s.pod.uid"]
+      - type: add
+        field: attributes["fluent.tag"]
+        value: EXPR("containers." + attributes["k8s.container.name"])
+      ## Use remove operator when available in opentelemetry collector:
+      ## https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9524
+      - type: move
+        from: attributes["log.file.path"]
+        to: body["log.file.path"]
```
sumo-drosiek, line 200: remove operator is already supported

```suggestion
      - type: remove
        field: attributes["log.file.path"]
```

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -1350,6 +1350,26 @@ Example Usage:
 {{ $enabled }}
 {{- end -}}
 
+{{/*
+Check if otelcol logs collector is enabled.
+It's enabled if logs are enabled but Fluent-Bit is disabled
+
```
sumo-drosiek, line 1356: We should have possibility to disable otellogs even if logs are enabled

deploy/helm/sumologic/values.yaml
```diff
@@ -4862,9 +4924,6 @@ otelevents:
 ## which is consistent with CRI, but may possibly cause issues on older K8s versions.
 ## This is an alpha feature, and may change in the near future.
 otellogs:
-  ## In order to enable this feature, a configuration block under metadata.logs.config.service.pipelines.logs/otlp/containers
-  ## needs to be uncommented
-  enabled: false
```
sumo-drosiek, line 4867: I would leave this option

tests/integration/values/values_helm_otelcol_logs.yaml
```diff
@@ -36,29 +36,14 @@ metadata:
...
+        filelog/containers:
           exclude:
-            match_type: strict
-            record_attributes:
-              - key: k8s.container.name
-                value: receiver-mock
-    receivers:
-      journald:
-        directory: /run/log/journal
+            - /var/log/pods/receiver-mock_*/*/*.log
```
sumo-drosiek, line 46: I like it :)

sumo-drosiek (commented):
Overall this is fine. Are we going to document whose changes in this PR or separate one?

astencel-sumo:
I believe we should start with the documentation (this PR doesn't add or modify any). Writing the documentation should help us make sure the configuration properties that we define in `values.yaml` cover the scenarios we want to cover and that the names/structure make sense.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Moved to `sumologic.logs.collector`.

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
swiatekm-sumo: I think that should be added in a separate PR. Sounds like a complicated use case.

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
swiatekm-sumo: Done.

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
-      provider: otelcol
+    container:
+      enabled: false
+```
 
-## Enable processing logs from Opentelemetry as log collector
-metadata:
+### SystemD Logs
+
+Container logs are collected by default. This can be disabled by setting:
```
sumo-drosiek, line 127:
```suggestion
SystemD logs are collected by default. This can be disabled by setting:
```

sumo-drosiek (approved):
LGTM, please get approval from someone else as well :)

deploy/docs/opentelemetry_collector.md
```diff
@@ -47,11 +49,26 @@ All Opentelemetry Collector configuration for metrics is located in
 If you want to modify it, please see [Sumologic Opentelemetry Collector configuration][configuration]
 for more information.
 
-## Logs Metadata
+## Logs
 
-We are using Opentelemetry Collector like Fluentd to enrich metadata and to filter data.
+Opentelemetry Collector can be used for both log collection and metadata enrichment. For these roles,
+it replaces respectively Fluent-Bit and FluentD.
```
astencel-sumo, line 55: Micro-nit: I think it's "Fluentd", not "FluentD" :)

astencel-sumo: Also I believe Fluent Bit does not use a hyphen in the name: https://fluentbit.io/
image

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
-the same way for Opentelemetry Collector like for Fluentd.
+If you haven't modified the FluentD or Fluent-Bit configuration, this should be a drop-in replacement with no
+further changes required.
 
 ### Logs Configuration
 
-All Opentelemetry Collector configuration for logs is located in
-[`values.yaml`][values] as `metadata.logs.config`.
+High level Opentelemetry Collector configuration for logs is located in
+[`values.yaml`][values] under the `sumologic.logs`.
```
astencel-sumo, line 86:
```suggestion
[`values.yaml`][values] under the `sumologic.logs` key.
```

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
 
-If you want to modify it, please see [Sumologic Opentelemetry Collector configuration][configuration]
-for more information.
+Configuration specific to the log collector Daemonset can be found under the `otellogs` key.
```
astencel-sumo, line 88:
```suggestion
Configuration specific to the log collector DaemonSet can be found under the `otellogs` key.
```

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
+
+Finally, configuration specific to the metadata enrichment Statefulset can be found under the `metadata.logs` key.
```
astencel-sumo, line 90:
```suggestion
Finally, configuration specific to the metadata enrichment StatefulSet can be found under the `metadata.logs` key.
```

deploy/helm/sumologic/values.yaml
```diff
@@ -218,6 +218,68 @@ sumologic:
       ## Set provider service (either fluentd or otelcol).
       provider: fluentd
 
+    collector:
+      enabled: false
+
+    multiline:
+      enabled: true
+      start_regex: "^\\[?\\d{4}-\\d{1,2}-\\d{1,2}.\\d{2}:\\d{2}:\\d{2}.*"
```
astencel-sumo, line 226: What do you think about the name `first_line_regex` instead of `start_regex`? I feel that "start" is a bit more ambiguous to me.

Also is the `.*` at the end of the regex required?

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
+by the high-level settings. See [Sumologic Opentelemetry Collector configuration][configuration]
+for more information
 
 [configuration]: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/Configuration.md
 [values]: ../helm/sumologic/values.yaml
 
-## Scraping Containers Logs
+### Multiline log parsing
+
+Multiline log parsing for Opentelemetry Collector can be configured using the `sumologic.logs.multiline` section in [values.yaml][values].
```
astencel-sumo, line 102:
```suggestion
Multiline log parsing for OpenTelemetry Collector can be configured using the `sumologic.logs.multiline` section in [values.yaml][values].
```

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
-      provider: otelcol
+    container:
+      enabled: false
+```
 
-## Enable processing logs from Opentelemetry as log collector
-metadata:
+### SystemD Logs
+
+SystemD logs are collected by default. This can be disabled by setting:
```
astencel-sumo, line 127: Again, I don't think there's a capital "D" in "systemd".
```suggestion
Systemd logs are collected by default. This can be disabled by setting:
```

deploy/helm/sumologic/values.yaml
sumo-drosiek: > Also is the .* at the end of the regex required?

it is not afaik

deploy/docs/opentelemetry_collector.md
swiatekm-sumo: This is actually wrong elsewhere in the document as well, I can fix it while I'm at it.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: > What do you think about the name `first_line_regex` instead of `start_regex`? I feel that "start" is a bit more ambiguous to me.

Sounds good to me, I didn't really put much thought into the name of this option. I don't think there's any standard terminology for this.

deploy/helm/sumologic/values.yaml
```diff
@@ -218,6 +218,68 @@ sumologic:
...
+
+    container:
+      enabled: true
+
+    systemd:
```
astencel-sumo, line 231: Correct me if I'm wrong, but it looks like there's now no simple way to set e.g. a custom source category on logs from systemd?

It's just an observation - I don't think it's a blocker for this PR, as this can be added in a later PR, right?

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Done.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: That's something that the metadata layer does, so we'd have to add it there, ideally while also moving configuration to a template like I've done here. This PR is only concerned with configuration for the log collector.

deploy/docs/opentelemetry_collector.md
```diff
@@ -60,148 +77,71 @@ sumologic:
...
+```yaml
+sumologic:
+  logs:
+    multiline:
+      enabled: true
+      first_line_regex: "^\\[?\\d{4}-\\d{1,2}-\\d{1,2}.\\d{2}:\\d{2}:\\d{2}.*"
+```
 
-**Note:** Opentelemetry Collector does not support systemd logs yet.
+where `first_line_regex` is a regular expression used to detect the start of the log line.
```
astencel-sumo, line 112:
```suggestion
where `first_line_regex` is a regular expression used to detect the first line of a multiline log.
```

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -1369,6 +1369,27 @@ Example Usage:
 {{ $enabled }}
 {{- end -}}
 
+{{/*
+Check if otelcol logs collector is enabled.
+It's enabled if both logs in general and the collector specifically are enabled.
+If both the collector and Fluent-Bit are enabled, we error.
```
astencel-sumo, line 1375: Do we really want to explicitly forbid the combination of both OT logs collector and Fluent Bit?

I can imagine a user might e.g. want to collect container logs with OT and systemd logs with FB (or the other way around) for some reason and I'm not convinced that hardcoding the lack of this possibility is a good idea.

sumo-drosiek: I think that customers more likely forget to disable one of the collector during migration than will try to use both of them. Especially when we deprecate fluent-bit

swiatekm-sumo: I think the value of catching accidental misconfigurations is greater than the value of allowing users to have both simultaneously. Now that OT supports systemd logs, I don't think there's any reason to run them side-by-side.

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
```diff
@@ -77,7 +77,7 @@ spec:
         - mountPath: /var/lib/docker/containers
           name: varlibdockercontainers
           readOnly: true
-        - mountPath: {{ .Values.otellogs.config.extensions.file_storage.directory }}
+        - mountPath: /var/lib/storage/otc
```
astencel-sumo, line 80: So it is now not possible to customize the path to OTC storage? We can add this option later though, right?

swiatekm-sumo: I used to have a separate setting for this, but I ultimately decided it didn't add any value, and only made things more confusing in the context of the log collector, where there's no persistent queue, the volume is a hostPath, and it's only used for storing the filelog receiver state.

For metadata, I think this configuration can be moved to `sumologic.logs`, like in #2444 .

deploy/docs/opentelemetry_collector.md
swiatekm-sumo: Fixed

deploy/helm/sumologic/values.yaml
```diff
@@ -218,6 +218,68 @@ sumologic:
       ## Set provider service (either fluentd or otelcol).
       provider: fluentd
 
+    collector:
+      enabled: false
```
perk-sumo, line 222: Can we change the name of this key to make it obvious it is about otelcol?
Current name would be OK in v4 when we'll have only single agent type but right now I think we should be more explicit.

I suggest make it:
```
sumologic:
  logs:
    collector:
      otelcol:
        enabled: true
```

swiatekm-sumo: Done.

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
astencel-sumo: Fair enough, it seems there is no need to modify this path 👍

astencel-sumo (approved):
LGTM as long as we agree that we want to forbid the simultaneous usage of Fluent Bit and OT logs collector. I was against forbidding it, but if everyone is fine with it, I'm good.

After all, this can be changed later if we change our minds (or the customers change our minds).

deploy/helm/sumologic/templates/_helpers.tpl
swiatekm-sumo: I added it as `allowSideBySide`, along with a documentation section.

swiatekm-sumo: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2436/commits/b7f0f80e64ce8de261f5b0417223fab4ca4038c0

deploy/docs/opentelemetry_collector.md
```diff
@@ -145,6 +145,28 @@ sumologic:
...
+    collector:
+      otelcol:
+        enabled: true
+      allowSideBySide: true
+  
+fluent-bit:
+  enabled: true
+```
+
+> **WARNING** This will cause each log line to be ingested twice, with all the associated costs.
```
astencel-sumo, line 168: This is true, but only when the user hasn't modified the configuration of Otelcol and Fluent Bit. How about this?
 
```
> **WARNING** Without further modifications to Otelcol and Fluent Bit configuration, this will cause each log line to be ingested twice, potentially doubling the log ingestion cost.
```

astencel-sumo (approved):

PR #2456, opened 2022-07-27
feat!: migrate to OTLP sources for all data types

swiatekm-sumo:
##### Description

Add OTLP sources and use them in OT configuration. This covers all data types. I'm not sure if this makes it easier or harder to review - in the latter case, I can split it into separate PRs.

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


deploy/helm/sumologic/values.yaml
```diff
@@ -4139,11 +4026,8 @@ metadata:
         pprof: {}
       exporters:
         sumologic/containers:
-          log_format: json
-          json_logs:
-            add_timestamp: true
-            timestamp_key: timestamp
-          endpoint: ${SUMO_ENDPOINT_DEFAULT_LOGS_SOURCE}
+          log_format: otlp
```
sumo-drosiek, line 4029: `otlp` is default format for `logs`, `metrics`, `traces`

swiatekm-sumo: I know, but I thought being explicit about this would make the PR easier to review.

PR #2466, opened 2022-08-10
feat: store Sumo credentials for the setup Job in a Secret

swiatekm-sumo:
##### Description

Adds some additional security. The setup Job is run as a Helm pre-install hook, so it gets deleted after it finishes. While it was running, it was possible to read the credentials from the Pod definition. With this change, they're stored in a Secret instead.

---

##### Checklist

- [x] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/templates/setup/job.yaml
+++ b/deploy/helm/sumologic/templates/setup/job.yaml
@@ -61,16 +61,10 @@ spec:
           mountPath: /etc/terraform
         - name: custom
           mountPath: /customer-scripts
-        {{- if .Values.sumologic.envFromSecret }}
         envFrom:
         - secretRef:
-            name: {{ .Values.sumologic.envFromSecret }}
-        {{ else }}
+            name: {{ .Values.sumologic.envFromSecret | default (include "sumologic.metadata.name.setup.secret" .)}}
         env:
-        - name: SUMOLOGIC_ACCESSID
-          value: {{ required "A valid .Values.sumologic.accessId entry required!" .Values.sumologic.accessId }}
-        - name: SUMOLOGIC_ACCESSKEY
-          value: {{ required "A valid .Values.sumologic.accessKey entry required!" .Values.sumologic.accessKey }}
         - name: SUMOLOGIC_BASE_URL
           value: {{ .Values.sumologic.endpoint }}
         - name: HTTP_PROXY
@@ -83,7 +77,6 @@ spec:
         - name: DEBUG_MODE
           value: "true"
 {{- end }}
-        {{ end }}
       securityContext:
         runAsUser: 999
 {{- end }}
--- a/deploy/helm/sumologic/templates/setup/secret.yaml
+++ b/deploy/helm/sumologic/templates/setup/secret.yaml
@@ -1,0 +1,14 @@
+{{- if not .Values.sumologic.envFromSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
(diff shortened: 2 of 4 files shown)


sumo-drosiek (approved):

deploy/helm/sumologic/templates/setup/secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: {{ template "sumologic.metadata.name.setup.secret" . }}
+  annotations:
+{{ include "sumologic.annotations.app.setup.helmsh" "3" | indent 4 }}
+  labels:
+    app: {{ template "sumologic.labels.app.setup.secret" . }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+data:
+  sumologic_accessid: {{ required "A valid .Values.sumologic.accessId entry required!" .Values.sumologic.accessId | b64enc }}
+  sumologic_accesskey: {{ required "A valid .Values.sumologic.accessKey entry required!" .Values.sumologic.accessKey | b64enc }}
```
sumo-drosiek, line 12: I'm thinking about making it optional, so customer can provide it by hand before installation

swiatekm-sumo: Hm, there's already https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/7ff4d7aa7b6025f613e7dbe1d7aaa043ef27c9d6/deploy/helm/sumologic/templates/setup/job.yaml#L64, I guess I could only create the secret if it's not set, and document appropriately.

sumo-drosiek: :+1: 

It will avoid raising `A valid .Values.sumologic.accessId entry required!` if secret is set by hand

deploy/helm/sumologic/templates/setup/job.yaml
```diff
@@ -68,9 +68,16 @@ spec:
...
+            secretKeyRef:
+              name: {{ template "sumologic.metadata.name.setup.secret" . }}
+              key: sumologic_accessid
         - name: SUMOLOGIC_ACCESSKEY
-          value: {{ required "A valid .Values.sumologic.accessKey entry required!" .Values.sumologic.accessKey }}
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "sumologic.metadata.name.setup.secret" . }}
+              key: sumologic_accesskey
+        {{ end }}
```
sumo-drosiek, line 80: We are cutting `env:` by moving this `{{end}}` for rest of environmental values

deploy/helm/sumologic/templates/setup/job.yaml
```diff
@@ -61,16 +61,10 @@ spec:
           mountPath: /etc/terraform
         - name: custom
           mountPath: /customer-scripts
-        {{- if .Values.sumologic.envFromSecret }}
         envFrom:
         - secretRef:
-            name: {{ .Values.sumologic.envFromSecret }}
-        {{ else }}
+            name: {{ .Values.sumologic.envFromSecret | default (include "sumologic.metadata.name.setup.secret" .)}}
```
sumo-drosiek, line 66: :heart_eyes:

deploy/helm/sumologic/templates/setup/job.yaml
swiatekm-sumo: Yeah, in the end I just changed the keys in the secret and added an unconditional `envFrom` with only the secret name changing.

sumo-drosiek (approved):

PR #2474, opened 2022-08-18
Adjust queueing and batching configurarion for OTC according to testi…

pdelewski:
…ng results

##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -3772,7 +3772,7 @@ metadata:
             num_consumers: 10
             ## setting queue_size a high number, so we always use maximum space of the storage
             ## minimal alert non-triggering queue size (if only one exporter is being used): 10GB/16MB = 640
-            queue_size: 50_000
+            queue_size: 10_000
           max_request_body_size: 16_777_216  # 16 MB before compression
           ## set timeout to 30s due to big requests
           timeout: 30s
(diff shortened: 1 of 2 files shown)


deploy/helm/sumologic/values.yaml
```diff
@@ -3977,7 +3977,7 @@ metadata:
         ## ref: https://github.com/SumoLogic/opentelemetry-collector/blob/main/processor/batchprocessor
         batch:
           ## Number of spans after which a batch will be sent regardless of time
-          send_batch_size: 16_384
+          send_batch_size: 1_024
           ## Time duration after which a batch will be sent regardless of size
           timeout: 30s
```
pdelewski, line 3982: Should we update timeout as well?

swiatekm-sumo: Yeah, a lower value like `1s` is better imo.

swiatekm-sumo (approved):
LGTM, do you plan on doing logs in a separate PR?

pdelewski:
> LGTM, do you plan on doing logs in a separate PR?

From what I see, originally it was done only for metrics, so I updated only metrics. I can prepare separate PR for logs

deploy/helm/sumologic/values.yaml
pdelewski: I will update that

PR #2484, opened 2022-08-22
docs(configuration): prepare configuration introduction

sumo-drosiek:
add document which describes configuration basics. This document should be used to explain how to use configuration instead of explaining it inline in multiple places

### DIFF
--- a/deploy/docs/configuration.md
+++ b/deploy/docs/configuration.md
@@ -1,0 +1,98 @@
+# Configuration
+
+This document describes how configuration works and what are the outcomes.
+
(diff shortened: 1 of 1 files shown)


pdelewski (commented):
Will this be somehow connected with other documentation files as currently our documentation is distributed into several files and it would be good to smoothly go from introduction to more detailed topics

sumo-drosiek:
@pdelewski I don't have idea how to connect it (nicely) yet, but those are information which are repeated in several places, so I wanted to start with intro and then link it from those places.

PR #2487, opened 2022-08-23
chore: upgrade otelcol for metadata and events to 0.56.0-sumo-0

sumo-drosiek:
upgrade otelcol for metadata and events to 0.56.0-sumo-0


deploy/helm/sumologic/conf/events/otelcol/config.yaml
```diff
@@ -39,6 +39,8 @@ processors:
     source_name: {{ .Values.fluentd.events.sourceName | quote}}
     source_category: {{ .Values.fluentd.events.sourceCategory | default (printf "%s/%s"  (include "sumologic.clusterNameReplaceSpaceWithDash" .) (.Values.fluentd.events.sourceName )) | quote}}
     source_category_prefix: ""
+  sumologic_schema:
+    add_cloud_namespace: false
```
astencel-sumo, line 43: Don't you need to set `translate_attributes: false` on the `sumologic` exporter?

sumo-drosiek: It should be idempotent and we can skip additional changes when they will be dropped.

swiatekm-sumo (approved):

PR #2491, opened 2022-08-24
test(integration): improve teardown

swiatekm-sumo:
##### Description

Add explicit teardown steps for resources created by integration tests.E2E framework's AfterEachTest actions are not actually called if the test errors out or panics, so we add explicit Finish actions for all of our resources.

### DIFF
--- a/tests/integration/main_test.go
+++ b/tests/integration/main_test.go
@@ -11,6 +11,7 @@ import (
 	"sigs.k8s.io/e2e-framework/klient"
 	"sigs.k8s.io/e2e-framework/pkg/env"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
 	"sigs.k8s.io/e2e-framework/support/kind"
 
 	"github.com/gruntwork-io/terratest/modules/k8s"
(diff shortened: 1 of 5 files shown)


tests/integration/main_test.go
```diff
@@ -102,6 +104,16 @@ func ConfigureTestEnv(testenv env.Environment) {
...
+
+	// Teardown
+	// TODO: Uninstall the Helm Chart here as well
+	testenv.Finish(envfuncs.DeleteNamespace(openTelemetryOperatorNamespaces[0]))
+	testenv.Finish(envfuncs.DeleteNamespace(openTelemetryOperatorNamespaces[1]))
+	testenv.Finish(envfuncs.DeleteNamespace(receiverMockNamespace))
+	testenv.Finish(func(ctx context.Context, envConf *envconf.Config) (context.Context, error) {
+		namespace := ctxopts.Namespace(ctx)
+		return envfuncs.DeleteNamespace(namespace)(ctx, envConf)
+	})
```
sumo-drosiek, line 116: why not just
testenv.Finish(envfuncs.DeleteNamespace(ctxopts.Namespace(ctx)))

swiatekm-sumo: Because you need to get the namespace from the context, and we get the context only at runtime. Your suggestion acts as if `ctx` is available in the function we're setting these Finish actions in, and it's not.

sumo-drosiek (approved):

tests/integration/main_test.go
sumo-drosiek: uch, ok

PR #2501, opened 2022-09-02
feat(tests/integration): add integration tests for otlp traces

aboguszewski-sumo:
This PR adds integration tests for otlp tracing.
The traces are generated using [customer-trace-tester](https://github.com/SumoLogic/sumologic-kubernetes-tools#customer-trace-tester) and then tested in a similar way to how logs are tested.

### DIFF
--- a/tests/integration/internal/stepfuncs/traces.go
+++ b/tests/integration/internal/stepfuncs/traces.go
@@ -1,0 +1,53 @@
+package stepfuncs
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/tracesgenerator"
+)
+
+// Generate logsCount logs using a deployment.
+func GenerateTraces(
+	tracesPerExporter uint,
+	spansPerTrace uint,
+	tracesGeneratorName string,
+	tracesGeneratorNamespace string,
+	tracesGeneratorImage string,
+) features.Func {
+	return func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+		client := envConf.Client()
+		generatorOptions := *tracesgenerator.NewDefaultGeneratorOptions()
+		generatorOptions.TracesPerExporter = tracesPerExporter
+		generatorOptions.SpansPerTrace = spansPerTrace
+
+		var namespace corev1.Namespace
+		err := client.Resources().Get(ctx, tracesGeneratorNamespace, "", &namespace)
+		if err != nil {
+			// create the namespace
+			namespace := corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: tracesGeneratorNamespace}}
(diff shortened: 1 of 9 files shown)


tests/integration/yamls/receiver-mock.yaml
```diff
@@ -26,16 +26,18 @@ spec:
         - ports:
             - containerPort: 3000
             - containerPort: 3001
-          image: sumologic/kubernetes-tools:2.12.0
```
aboguszewski-sumo, line 29: This can't be merged until a new version of k8s-tools is released. Currently it is still undergoing some changes.

tests/integration/internal/stepfuncs/traces.go
```diff
@@ -0,0 +1,54 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/tracesgenerator"
+)
+
+// Generate logsCount logs using the designated implementation type:
+// either deployment or a daemonset.
```
swiatekm-sumo, line 18: The comment says we can use either a Deployment or Daemonset, but the implementation just uses a Daemonset unconditionally. Is there any benefit to using a Daemonset here to begin with?

swiatekm-sumo (commented):
I'm not sure why we're using a DaemonSet for the trace generator instead of a Deployment. Looks good otherwise, great work!

tests/integration/internal/stepfuncs/traces.go
aboguszewski-sumo: To be honest, it's a leftover from copying and paste :)
I was hoping that someone would stratify this in a review, but I forgot to mention that in a comment, sorry.
Is there some overhead related to using Daemonset? I guess that changing to deployment won't hurt.

swiatekm-sumo: There isn't overhead, but DaemonSets and Deployments have different semantics. In particular, a DaemonSet starts a Pod for each Node in the cluster, while a Deployment will start as many as you tell it to. The kind cluster used for tests only has one Node, so they're effectively equivalent in this case, but the test would start failing if someone adds more Nodes, which is not ideal.

In general, I don't think there's any reason to use a DaemonSet here. @mat-rumian WDYT?

mat-rumian: As we use the same version of OT Distro for all our deployments I think it's not necessary to run tests with Daemonset. Single deployment testing traces pipeline should be enough :)

aboguszewski-sumo: I fixed this. However, I believe that there is one difference: Deployment causes the generator to run multiple times, while DaemonSet runs only once and terminate. Because of that, these tests are not testing a case where some of the traces are being dropped, because the data is no longer limited.

swiatekm-sumo: If you don't want the Deployment to restart the container, you can use the same trick we used for log generation here: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/887e2b312e99b34c0e7f7e76a96ae2e5eac7295c/tests/integration/internal/logsgenerator/logsgenerator.go#L85, which is basically just adding a sleep command.

swiatekm-sumo (approved):

swiatekm-sumo:
By the way, I don't think this needs a changelog entry, as it's not a user-facing feature. I'm also wondering if this test shouldn't go into the `default` tests, but that can happen in a separate PR.

PR #2538, opened 2022-09-21
docs: parsing json content by fluentd

pdelewski:
##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

### DIFF
--- a/deploy/docs/Best_Practices.md
+++ b/deploy/docs/Best_Practices.md
@@ -45,6 +45,7 @@
   - [Using NodeSelectors](#using-nodeselectors)
     - [Binding pods to linux nodes](#binding-pods-to-linux-nodes)
 - [Disable Thanos](#disable-thanos)
+- [Parsing log content as json](#parsing-log-content-as-json)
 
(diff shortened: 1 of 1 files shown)


deploy/docs/Best_Practices.md
```diff
@@ -1448,3 +1449,23 @@ kube-prometheus-stack:
     prometheusSpec:
       thanos: null
 ```
+
+## Parsing log content as json
+
+In order to parse and store log content as json following configuration has to be applied.
+This is needed due to correct metadata enrichment otherwise everything that is part of
+metadata like `kubernetes` tag will be visible toghether with log content instead of part
+of fields
```
sumo-drosiek, line 1458: I would skip this. It is incomplete or you have to have context to understand the motivation

pdelewski: > I would skip this. It is incomplete or you have to have context to understand the motivation

ok, just wanted to add at least some context

sumo-drosiek (approved):

Issue #1822, opened 2021-10-18
Update documentation around additionalRemoteWrite for kube-prometheus-stack


perk-sumo:

perk-sumo:
We cannot bump the kube-prometheus-stack in v2.2 because that would be a breaking change: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/1832
This task needs to be done whenever we can update the stack again.

rnishtala-sumo:
@perk-sumo  @sumo-drosiek  The documentation update was merged in the PR below
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2549
Could we close this issue?

PR #2549, opened 2022-10-03
docs: update documentation for additionalRemoteWrite

rnishtala-sumo:
##### Description

Update documentation around additionalRemoteWrite for kube-prometheus-stack

### DIFF
--- a/deploy/docs/additional_prometheus_configuration.md
+++ b/deploy/docs/additional_prometheus_configuration.md
@@ -4,6 +4,8 @@
 
 Prometheus configuration is located in `values.yaml` under `kube-prometheus-stack` key for helm installation.
 
+__Note__: It is best practice to add custom configuration to a user-supplied values file and then use it like so (```helm install -f my_values.yaml -n sumologic```)
+
 If the `kube-prometheus-stack` has been installed directly, a `prometheus-overrides.yaml` should be generated
 using `docker`/`kubectl` and [sumologic-kubernetes-tools](https://github.com/sumologic/sumologic-kubernetes-tools#template-dependency-configuration):
 
@@ -30,14 +32,10 @@ The configuration contains a section like the following for each of the Kubernet
 If you would like to collect other metrics that are not listed in configuration, you can add a new section to the file.
 
 ```yaml
-kube-prometheus-stack:  # For values.yaml
-    # ...
+kube-prometheus-stack:  # Add to a user-supplied values.yaml
     prometheus:
-      # ...
       prometheusSpec:
-        # ...
-        remoteWrite:
-          # ...
+        additionalRemoteWrite:
           - url: http://$(FLUENTD_METRICS_SVC).$(NAMESPACE).svc.cluster.local.:9888/prometheus.metrics.<some_label>
             writeRelabelConfigs:
             - action: keep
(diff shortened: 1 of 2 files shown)


deploy/docs/additional_prometheus_configuration.md
```diff
@@ -36,7 +36,7 @@ kube-prometheus-stack:  # For values.yaml
       # ...
       prometheusSpec:
         # ...
-        remoteWrite:
+        additionalRemoteWrite:
           # ...
```
sumo-drosiek, line 40: no need for `# ...` it was indication that there was something to copy

also, we should search documentation for all `remoteWrite` mentions, as I'm not sure but we probably mention in multiple places that you need to copy it from our `values.yaml` which is no longer true

sumo-drosiek (commented):
I would like to ask you also to go through the process of adding additional prometheus metrics, so we can have additional input on that, and make it less complicated (/more understable) :sweat:

deploy/docs/additional_prometheus_configuration.md
rnishtala-sumo: ok got it, will look for other references to remoteWrite. Also, I was able to send some metrics from the k8s-collector to a sumo instance. Will try adding additional prometheus metrics.

rnishtala-sumo: get this error when I try to send Prometheus metrics to Sumo (long) via Fluentd.
config error file="/fluentd/etc/fluent.conf" error_class=Fluent::ConfigError error="Invalid SumoLogic endpoint url: "

it's not clear where in endpoint url gets configured in fluentd, configured the following in values.yaml
endpoint: "https://long-api......."

rnishtala-sumo: Also @sumo-drosiek there doesn't seem to be any other mention of copying remoteWrite from values.yaml in the documentation

rnishtala-sumo: Was able to send additional Prometheus metrics from a new metrics source. Also, removed copy references from the readme and added a note like below

NOTE: It is best practice to add custom configuration to a user-supplied values file and then use it like so (```helm install -f my_values.yaml -n sumologic```)

deploy/docs/additional_prometheus_configuration.md
```diff
@@ -271,15 +268,13 @@ To send custom metrics to Sumo Logic you need to update it to include a rule to
 Make sure you include the same tag you created in your Fluentd configmap in the previous step.
 Here is an example addition to the configuration file that will forward metrics to Sumo:
 
+NOTE: Please add the below to a user-supplied values file  (```helm install -f```)
```
sumo-drosiek, line 271: incomplete command

sumo-drosiek:
Generally I'm not sure about copying the same note in multiple places in one document
Also for all our docs Note's are usually **bold**

cc: @astencel-sumo

deploy/docs/additional_prometheus_configuration.md
rnishtala-sumo: ok makes sense, I could keep the one comment at the beginning and remove the rest of them. Would that work?

sumo-drosiek (approved):
Thanks!

swiatekm-sumo:
@rnishtala-sumo the CI should be fine if you rebase on `main`.

rnishtala-sumo:
Fixes issue: #1822

PR #2570, opened 2022-10-14
ci: add PR job for reporting integration test status

swiatekm-sumo:
We want to make integration tests a required check, but this is annoying to do given their number. Add a job that only reports their aggregate success for this purpose.


rnishtala-sumo (approved):
Seems like a good idea to expect for all the integration tests to pass before merge 👍

PR #2605, opened 2022-11-11
Removing all references of the PodSecurityPolicy for fluentd

rnishtala-sumo:
##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


sumo-drosiek (commented):
I would consider backward compatibility for k8s < 1.25

swiatekm-sumo:
> I would consider backward compatibility for k8s < 1.25

You mean like retaining the PodSecurityPolicies for k8s < 1.25? Does that actually serve any practical purpose?

I think we should just mark this as a breaking change and be done with it.

rnishtala-sumo:
@sumo-drosiek @swiatekm-sumo have we decide that this will be a breaking change and can be removed, or do we need to think about backward compatibility?

sumo-drosiek:
> @sumo-drosiek @swiatekm-sumo have we decided that this will be a breaking change and can be removed, or do we need to think about backward compatibility?

Let's make it breaking change. We can always re-add it

sumo-drosiek (approved):

PR #2609, opened 2022-11-14
docs: add information about creation of additional Prometheus instance

kkujawa-sumo:
docs: add information about creation of additional Prometheus instance
docs: add section about OOM failures for Prometheus Pod to troubleshooting collection doc


deploy/docs/additional_prometheus_configuration.md
```diff
@@ -300,3 +300,75 @@ invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]s
 ```
 
 If all goes well, you should now have your custom metrics piping into Sumo Logic.
+
+## Additional Prometheus instance
+
+When there is a need to collect huge amount of metrics then single Prometheus Pod may not be enough.
+One of the solutions that can be used is creating additional Prometheus instance for capturing only selected metrics.
+
```
swiatekm-sumo, line 308:
````suggestion
Metrics can be assigned to different Prometheus instances by giving them separate sets of ServiceMonitors to handle. In the Prometheus resource, the settings responsible for this are:
```yaml
  serviceMonitorNamespaceSelector:
  serviceMonitorSelector:
```
Figuring out what ServiceMonitors exist in your cluster, the metrics each of them selects, and their data volume, are beyond the scope of this document.
````

swiatekm-sumo (commented):
I think we should be more explicit about how this is supposed to work. Looks good otherwise.

deploy/docs/additional_prometheus_configuration.md
swiatekm-sumo: I screwed up the diff, this was supposed to be an addition, not a replacement. 😒

kkujawa-sumo: @swiatekm-sumo I added this information, please take another look ;)

swiatekm-sumo (approved):

PR #2610, opened 2022-11-14
fix(logs): Changing the service name to metadata-logs to be provider independent

rnishtala-sumo:
##### Description

- Changing the service name to be independent of the logs provider
- Changes to integration tests to use the new service name
---

##### Checklist

- [x] Changelog updated

###### Testing performed

- [x] Redeploy metadata-logs and fluentbit
- [x] Confirm logs and metrics are coming in


deploy/helm/sumologic/templates/chart-configmap.yaml
```diff
@@ -3,6 +3,6 @@ kind: ConfigMap
 metadata:
   name: sumologic-configmap
 data:
-  fluentdLogs: {{ template "sumologic.metadata.name.logs.service" . }}
+  otelcolLogs: {{ template "sumologic.metadata.name.logs.service" . }}
```
sumo-drosiek, line 6: could we name it `logs`, `metadataLogs` or `logsMetadata`?

I always felt that using specific component name was bad decision

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -412,15 +412,11 @@ helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
 {{- end -}}
 
 {{- define "sumologic.metadata.name.logs" -}}
-{{- if eq .Values.sumologic.logs.metadata.provider "fluentd" -}}
-{{ template "sumologic.metadata.name.fluentd" . }}-logs
-{{- else if eq .Values.sumologic.logs.metadata.provider "otelcol" -}}
 {{ template "sumologic.metadata.name.otelcol" . }}-logs
```
sumo-drosiek, line 415: We should leave statefulset names as they are. AFAIK, After this change we can end with fluentd statefulset named otelcol

deploy/helm/sumologic/templates/chart-configmap.yaml
rnishtala-sumo: yeah, maybe I'll go for `metadataLogs`

deploy/helm/sumologic/templates/_helpers.tpl
rnishtala-sumo: what I was trying to do here was to use `otelcol` regardless of  the provider.

rnishtala-sumo: actually, I see what you mean, we can't change the name of a StatefulSet on upgrade. I'll change the default provider for logs and metrics to otelcol.

deploy/helm/sumologic/values.yaml
```diff
@@ -1061,7 +1061,8 @@ fluent-bit:
   podLabels: {}
 
   ## Add custom pod annotations to fluent-bit daemonset pods
-  podAnnotations: {}
+  podAnnotations:
+    sumologic.com/otel: "true"
```
swiatekm-sumo, line 1065: This is a bit misleading, since it's not really about otel, but the name of the service. Maybe something like `sumologic.com/major-version: 3`.

We should also add a comment explaining why this exists.

tests/integration/helm_default_installation_test.go
```diff
@@ -27,10 +27,10 @@ import (
 	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/stepfuncs"
 )
 
-func Test_Helm_Default_FluentD_Metadata(t *testing.T) {
+func Test_Helm_Default_Otel_Metadata(t *testing.T) {
```
swiatekm-sumo, line 30: If you're going to change this, you should modify the existing otel metadata test to use fluentd instead as well.

rnishtala-sumo: I changed this because the default metadata provider was changed to otel. ok I'll ensure we have tests for both fluentd and otc metadata installation.

deploy/helm/sumologic/values.yaml
```diff
@@ -235,7 +235,7 @@ sumologic:
     enabled: true
     metadata:
       ## Set provider service (either fluentd or otelcol).
-      provider: fluentd
+      provider: otelcol
```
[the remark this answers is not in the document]
swiatekm-sumo: What I mean is that there are two changes in this PR:
* the service name
* the default provider
There isn't really any reason for these to be tied together.

deploy/helm/sumologic/values.yaml
```diff
@@ -1061,7 +1061,9 @@ fluent-bit:
   podLabels: {}
 
   ## Add custom pod annotations to fluent-bit daemonset pods
-  podAnnotations: {}
+  podAnnotations:
+    ## This annotation triggers a restart of the fluent-bit pod on upgrade to helm chart v3
+    sumologic.com/major-version: "3"
```
swiatekm-sumo, line 1066: I'm wondering if this is necessary now that we've changed the key in the configmap. In principle it should restart the Fluent-Bit Pods with just that.

rnishtala-sumo: yes, it looks like the annotation is not necessary, the fluent-bit pod restarted on upgrade,

rnishtala-sumo: this change is removed.

swiatekm-sumo (approved):

Issue #1480, opened 2021-03-05
A specific parameter for additional remote writes


abjrcode:
Why is it not possible to add additional `remoteWrites` just like it is the case for [`additionalServiceMonitors`](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/values.yaml#L1278)?

The documentation on scraping additional services now points to the prometheus-overrides.yaml file, but that is error prone as I need to _(possibly)_ re-generate it on chart updates?

Basically it would be nice to add additional monitors and remote writes without touching the ones provided by Sumo _(just augmenting them)_.

abjrcode:
Something to consider, I and probably most others implement this with GitOps and want to avoid running any commands manually if possible.

perk-sumo:
Hi, the reason for that is because the upstream chart we use doesn't provide such possibility - there is only one section for remoteWrite to be used:
https://github.com/prometheus-community/helm-charts/blob/kube-prometheus-stack-12.3.0/charts/kube-prometheus-stack/values.yaml#L1943

We know it's not ideal.

abjrcode:
Thanks for the clarification. Wouldn't be possible that you expose this as a parameter on your own chart and then merge the result when passing it to the subchart?

sumo-drosiek:
> Wouldn't be possible that you expose this as a parameter on your own chart and then merge the result when passing it to the subchart?

We are not aware of possibility to do that and keep everything clean and simple

We created [PR](https://github.com/prometheus-community/helm-charts/pull/747) in the kube-prometheus-stack helm chart, so maybe we will be able to do it in a proper way :crossed_fingers:

sumo-drosiek:
labeling as `breaking-change`, because it requires updating the Prometheus CRDs

PR #2611, opened 2022-11-15
feat: add parameter to configure additional Prometheus remote writes

kkujawa-sumo:
Fixes https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/1480

I've found that `additionalRemoteWrite` parametr is already mentioned in our documentation:
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/docs/additional_prometheus_configuration.md#filter-metrics

`remoteWrite` parameter is mentioned in following docs and I think that it doesn't need to be changed to `additionalRemoteWrite`:
- https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/docs/Performance.md#recommendations
- https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/docs/standAlonePrometheus.md#update-existing-prometheus


swiatekm-sumo (commented):
Don't we have any documentation to update alongside this? I tried searching, but it seems like a lot of it is already using this key, which is a bit strange. Does this PR just add it to the parameter list?

PR #2616, opened 2022-11-18
chore(vagrant): simplify jmx metric setup

swiatekm-sumo:
##### Description

The setup for jmx metrics in vagrant required building custom docker containers. This was annoying to do, especially since we use these manifests in E2E tests where we run in a more constrained environment. Fortunately, the docker containers were really simple and could be replaced with `initContainers`.

In essence, we download the necessary java agent jar in an init container, and then launch the main container with the right options. This PR also fixes some quality of life stuff, like using images from public ECR everywhere instead of dockerhub.


sumo-drosiek (approved):
:rocket:

PR #2618, opened 2022-11-18
feat(logs)!: switch to Otelcol as logs metadata provider

astencel-sumo:


tests/integration/helm_traces_enabled_test.go
```diff
@@ -42,144 +41,6 @@ func Test_Helm_Traces_Enabled(t *testing.T) {
 				require.Len(t, secret.Data, 11, "Secret has incorrect number of endpoints")
 				return ctx
 			}).
-		Assess("fluentd logs statefulset is ready",
```
swiatekm-sumo, line 45: If you're deleting these checks, then we should also adjust the values.yaml for this test to disable logs in the first place.

astencel-sumo: Hm I think this test was meant to test that all things can run at the same time. Maybe I should restore these checks after all. :thinking:

astencel-sumo:
Closing in favor of #2621

PR #2621, opened 2022-11-18
feat(logs): Changing the default logs metadata provider to otel

rnishtala-sumo:
##### Description

Changing the default logs metadata provider to otel
This PR shoud be merged after: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2610

---

##### Checklist

- [ ] Changelog updated

###### Testing performed

- [ ] Redeploy metadata-logs and fluentbit
- [ ] Confirm logs and metrics are coming in


tests/integration/helm_default_installation_test.go
```diff
@@ -300,22 +300,19 @@ func Test_Helm_Default_FluentD_Metadata(t *testing.T) {
...
-				"namespace_id":    "",
-				"pod":             "",
-				"pod_id":          "",
-				"container_id":    "",
-				"host":            "",
-				"master_url":      "",
-				"node":            "",
-				"_sourceName":     "",
-				"_sourceCategory": "",
-				"_sourceHost":     "",
```
sumo-drosiek, line 318: Why do we removed some of the expected metadata?

tests/integration/values/values_helm_fluentd_logs.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - hostPath:
+        path: /var/lib/fluent-bit
+        type: DirectoryOrCreate
+      name: tail-db
+    # kind doesn't enable journald persistence, and the journal resides at /run/log/journal
+    # instead of /var/log/journal
+    - hostPath:
+        path: /run/log/journal
+        type: DirectoryOrCreate
+      name: run-log-journal
```
sumo-drosiek, line 27:
```suggestion
      name: run-log-journal

```

tests/integration/helm_fluentd_logs_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+					return len(daemonsets) == 1
+				}, waitDuration, tickDuration)
+
+				require.EqualValues(t, 0, daemonsets[0].Status.NumberUnavailable)
+				return ctx
+			}).
+		Feature()
+
+	featMetrics := features.New("metrics").
```
astencel-sumo, line 187: If this is a logs test (as name implies), then why do we test the metrics?

swiatekm-sumo (commented):
I thought a bit more about the test changes, and I think this will be much simpler if we only changes the values.yaml for the tests for now. We can reorganize the tests after we'd changes all the defaults.

tests/integration/helm_default_installation_test.go
rnishtala-sumo: the expected metadata changes for otelcol, it has an additional entry and a few other changes:
`"_collector":       "kubernetes"`

tests/integration/helm_fluentd_logs_test.go
rnishtala-sumo: that's a good point I'll remove the metrics tests and only keep the logs.

tests/integration/helm_default_installation_test.go
sumo-drosiek: I'm curious what happened with `_source*`, these are pretty crucial fields and we should have it in sumo :grimacing:

rnishtala-sumo:
@swiatekm-sumo it looks like we need changes like these in the tests (outside values.yaml for the tests) when we change the default to otelcol, for the tests to pass.

tests/integration/helm_default_installation_test.go
rnishtala-sumo: got it, I have added those `_source*`, fields to the tests and pushed the changes.

swiatekm-sumo:
> @swiatekm-sumo it looks like we need changes like these in the tests (outside values.yaml for the tests) when we change the logs default to otelcol, for the tests to pass. https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2621/files#diff-5bdd3491954e96c5de4930d0097c8ee092aec3c7dd6184e3bf3a4c02d372bf0bR47

What I meant was that you should instead change this config file: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/values/values_helm_default_fluentd_metadata.yaml to use fluentd as the provider instead.

PR #2653, opened 2022-11-28
fix(setup): add permission to modify secrets

swiatekm-sumo:
##### Description

The setup Job needs permission to modify Secrets, not just create them.

---

##### Checklist

- [X] Changelog updated


sumo-drosiek (approved):

swiatekm-sumo:
This doesn't actually need a backport, the change which caused it was not applied to v2 because it was effectively breaking.

PR #2656, opened 2022-11-28
ci: fix integration test status job

swiatekm-sumo:
The status job aggregates the results of integration test jobs, so it's easy to collective use them for PR checks. However, by default, the job only runs if all the tests complete, whereas the integration tests are skipped if some of them fail. In that event, the status job would be skipped as well, and PR checks would pass when they shouldn't.

As a fix, force the status job to always run.


astencel-sumo (approved):
:shrug:

PR #2669, opened 2022-12-05
test(helm): run template tests using Go and terratest

swiatekm-sumo:
##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


tests/helm/utils.go
```diff
@@ -0,0 +1,173 @@
...
+		Version string `yaml:"version"`
+	}
+	err = yaml.Unmarshal(output, &chartInfo)
+	if err != nil {
+		return "", err
+	}
+	return chartInfo.Version, nil
+}
+
+// The functions below are copied from terratest for the sole reason of being able to skip the helm dependency update
```
astencel-sumo, line 35: Here you write you want to be able to skip helm dependency update, but above in the `setupDependencies` function you call the `helm dependency update` command. I'm probably not understanding something, can you explain?

astencel-sumo: This makes sense, thanks for the explanation. Yeah, the tests should totally support a "skip dependency update" flag to speed up local testing, but this can be a separate PR :+1:

astencel-sumo (approved):
This is a great improvement :clap: :bow:

PR #2670, opened 2022-12-06
test(helm): add builtin label test

swiatekm-sumo:
Add a test for basic builtin labels that covers all manifests generated by the chart. This isn't too exciting by itself, but it shows how you can write a test like this, which we can leverage to greater effect in the future.

The test caught a couple resources with missing labels, added those.

A decent amount of effort in this test is spent filtering out resources from subcharts, which don't and shouldn't have the labels. I'm not sure if what I've done is the best way of achieving this, would appreciate feedback.


astencel-sumo (approved):
As long as it works :crossed_fingers:

PR #2672, opened 2022-12-08
test(helm): refactor template tests

swiatekm-sumo:
Refactor the template tests to stop relying on `config.sh` files from the old test framework. These used to be necessary to know which template we are testing, but we can instead just render all the objects and then pick the one we want based on its name and kind. This will allow us to reorganize the template test files easily.

I also fixed some newline issues with tracing configmaps.

### DIFF
--- a/tests/helm/template_test.go
+++ b/tests/helm/template_test.go
@@ -1,15 +1,12 @@
 package helm
 
 import (
-	"errors"
-	"fmt"
 	"os"
-	"path"
-	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
 
+	doublestar "github.com/bmatcuk/doublestar/v4"
 	"github.com/gruntwork-io/terratest/modules/helm"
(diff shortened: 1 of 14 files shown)


tests/helm/template_test.go
```diff
@@ -86,31 +59,28 @@ func runTemplateTest(t *testing.T, templatePath string, valuesFileName string, o
 	renderedYamlString = fixupRenderedYaml(renderedYamlString, chartVersion)
 	expectedYamlBytes, err := os.ReadFile(outputFileName)
 	expectedYamlString := string(expectedYamlBytes)
-	expectedYamlString = fixupExpectedYaml(expectedYamlString)
+	// expectedYamlString = fixupExpectedYaml(expectedYamlString)
```
sumo-drosiek, line 62:
```suggestion
```

sumo-drosiek (approved):

PR #2673, opened 2022-12-08
test(helm): remove old test framework

swiatekm-sumo:
Remove the old test framework for template tests. I've also renamed the file-based tests to golden file tests.


Makefile
```diff
@@ -5,9 +5,6 @@ shellcheck:
 
 test: test-templates test-templates-bash
```
sumo-drosiek, line 6:
```suggestion
test: test-templates
```

PR #2691, opened 2022-12-13
test(integration): use docker image archives

swiatekm-sumo:
Add infrastructure for using docker image archives in integration tests.

A non-trivial chunk of an integration test's runtime is fetching docker images from repositories. Usually the same images are fetched for every test in any given run. This is wasteful and makes the tests annoying to run locally, especially on slower internet connections.

We add a Makefile target for generating an image archive. This is done by generating all the manifests for the test values files, extracting the image names using `sed`, pulling them, and saving to an archive using `docker save`.

We also modify the tests to make use of the archive by default. I'd like to make the CI use the cache in a separate PR.

In total, this reduces test runtime by around 40%.


tests/integration/.gitignore
```diff
@@ -0,0 +1 @@
+*.tar
```
sumo-drosiek, line 1:
```suggestion
*.tar

```

tests/integration/README.md
```diff
@@ -88,6 +88,30 @@ USE_KUBECONFIG=1 KUBECONFIG=/tmp/xm1r7h7s7 make test
...
+
+Running the above will create an `images.tar` file in this directory, which the tests will automatically make use of.
+
+### Using a custom image archive file
+
+It's possible to change the image archive file name:
+
+```bash
+make create-image-archive IMAGE_ARCHIVE=/my/file.tar
+IMAGE_ARCHIVE=/my/file.tar make test
```
sumo-drosiek, line 112: Could it be?
```
IMAGE_ARCHIVE=/my/file.tar make create-image-archive test
```

sumo-drosiek (approved):

astencel-sumo (approved):
The `create-image-archive.sh` has some issues with shellcheck v0.9.0 (released yesterday), but this should probably be fixed in a separate change.

Thanks for this contribution, this is a great improvement! :+1:

astencel-sumo:
The `xargs` command does not seem to work on MacOS :dizzy_face:

PR #2692, opened 2022-12-13
chore: upgrade otelcol for otellogs to 0.66.0-sumo-0

swiatekm:
In addition to the upgrade, I moved some tests to Go.


sumo-drosiek:
Do we need to perform any migration in configuration?

sumo-drosiek (approved):

PR #2694, opened 2022-12-14
chore: upgrade Fluent Bit to v2.0.6

astencel-sumo:
The image for v1.6.10 has vulnerabilities that we'd rather not care about fixing.
Let's upgrade to the latest and see how it works.


swiatekm-sumo (approved):

swiatekm-sumo:
Maybe we should redo the multipart and multiline configs in-line with what is recommended in 2.x? I'd love to get rid of all the Lua scripts.

PR #2696, opened 2022-12-14
test(integration): reorganize tests

swiatekm-sumo:
Reorganize the integration tests after all the changes to the defaults in v3. The idea was to have one test checking all the features for the default config, and other tests for features not enabled by default. As cluster creation and chart installation are the most expensive steps of an integration test, this should save us time in the long run.

This PR makes the following changes:
* move more configuration to values_common.yaml
* rename the `Default_OT_Metadata` test to just `Default_OT`
* move the events and traces tests to this default test
* remove the OT logs test, and move the multiline portion of it to the default test

We should eventually look at defining the test features separately, so we can more easily run the same tests under different configurations without too much code duplication.


tests/integration/values/values_helm_fluentbit_containerd_multiline_logs.yaml
```diff
@@ -126,3 +126,9 @@ sumologic:
     collector:
       otelcol:
         enabled: false
+  metrics:
+    enabled: false
+  traces:
+    enabled: false
+  events:
+    enabled: false
```
sumo-drosiek, line 134:
```suggestion
    enabled: false

```

tests/integration/helm_otelcol_events_test.go
```diff
@@ -1,95 +0,0 @@
...
-						})
-					if !assert.NoError(t, err) {
-						return false
-					}
-
-					return err == nil && len(pvcs.Items) == 1
-				}, waitDuration, tickDuration)
-				return ctx
-			}).
-		Feature()
```
sumo-drosiek, line 76: there is no need to check it anymore?

swiatekm-sumo: This check is present in the default test.

sumo-drosiek (approved):

PR #2707, opened 2022-12-15
test(integration): add otel and fluent bit logs test

swiatekm-sumo:
We were missing this configuration in tests, and we do support it.


perk-sumo (approved):
LGTM 🚀

PR #2714, opened 2022-12-19
ci: only check links in changed md files in PRs

swiatekm-sumo:
In PRs, only check links in modified Markdown files. We still check all of them on pushes to main in case a link expires.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -111,9 +111,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: gaurav-nelson/github-action-markdown-link-check@v1
-        if: needs.docs-changed.outputs.any_changed == 'true'
         with:
           config-file: '.markdown_link_check.json'
+          use-quiet-mode: yes
+          check-modified-files-only: yes
+          base-branch: main
 
   md-links-lint:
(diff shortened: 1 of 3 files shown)


rnishtala-sumo (approved):

.github/workflows/pull_requests.yml
```diff
@@ -111,9 +111,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: gaurav-nelson/github-action-markdown-link-check@v1
-        if: needs.docs-changed.outputs.any_changed == 'true'
         with:
           config-file: '.markdown_link_check.json'
+          use-quiet-mode: yes
+          check-modified-files-only: yes
+          base-branch: main
```
astencel-sumo, line 118: What will happen on pull requests against other branches, e.g. `release-v2`?

swiatekm-sumo: Not what we want. I'll change it to [github.base_ref.](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) Good catch!

swiatekm-sumo: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2720

PR #2716, opened 2022-12-20
feat(metrics): simplify custom application metrics 

sumo-drosiek:
##### Description

- auto-generate list of endpoints for telegraf configuration
- use `sumologic.metrics.serviceMonitors` to keep our serviceMonitors configuration
- left `additionalServiceMonitors` empty, so customer doesn't need to copy anything

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


docs/collecting-application-metrics.md
```diff
@@ -0,0 +1,170 @@
...
+          - action: keep
+            regex: <metric1>|<metric2>|...
+            sourceLabels: [__name__]
+```
+
+**Note:** We recommend to use regex validator, for example [https://regex101.com/].
+
+### Application metrics are exposed (multiple enpoints scenario)
+
+If in the pod there is more than one endpoint you want to scrape metrics from,
```
swiatekm-sumo, line 40:
```suggestion
If you want to scrape metrics from multiple endpoints in a single Pod,
```

docs/collecting-application-metrics.md
```diff
@@ -0,0 +1,170 @@
...
+    prometheusSpec:
+      additionalRemoteWrite:
+        - url: http://$(METADATA_METRICS_SVC).$(NAMESPACE).svc.cluster.local.:9888/prometheus.metrics.<custom endpoint name>
+          writeRelabelConfigs:
+          - action: keep
+            regex: <metric1>|<metric2>|...
+            sourceLabels: [__name__]
+```
+
+**Note:** We recommend to use regex validator, for example [https://regex101.com/]
```
swiatekm-sumo, line 78:
```suggestion
**Note:** We recommend using a regex validator, for example [https://regex101.com/]
```

swiatekm-sumo (commented):
Can you also add one or two template tests to check if `metric.remotewrites` works correctly with both our ServiceMonitors and the additional ones?

sumo-drosiek:
> Can you also add one or two template tests to check if `metric.remotewrites` works correctly with both our ServiceMonitors and the additional ones?

@swiatekm-sumo I assume you were meaning `metrics.serviceMonitors`

docs/collecting-application-metrics.md
```diff
@@ -0,0 +1,201 @@
+# Collecting application metrics
+
+This document is going to cover multiple different use cases related to scraping custom application metrics.
```
astencel-sumo, line 3:
```suggestion
This document covers multiple different use cases related to scraping custom application metrics.
```

tests/helm/kube-prometheus-stack/servicemonitors/basic-additional.input.yaml
```diff
@@ -0,0 +1,15 @@
+kube-prometheus-stack:
```
swiatekm-sumo, line 1: It would be nicer if this was a Go test instead, we could avoid the output templates and just check if the servicemonitor names in the list are right.

docs/collecting-application-metrics.md
```diff
@@ -0,0 +1,211 @@
+# Collecting application metrics
+
+This document covers multiple different use cases related to scraping custom application metrics.
```
astencel-sumo, line 3:
```suggestion
This document covers multiple different use cases related to scraping custom application metrics exposed in Prometheus format.
```

sumo-drosiek: It's not exactly true, but I will make few more iterations on that doc, so we can improve it later

PR #2722, opened 2022-12-21
ci: fix markdown link check in dev builds

swiatekm-sumo:
#2714 broke the dev builds by adding an incorrect configuration to the workflow. This fixes it.

### DIFF
--- a/.github/workflows/dev_builds.yml
+++ b/.github/workflows/dev_builds.yml
@@ -12,7 +12,6 @@ jobs:
 
   markdown-link-check:
     runs-on: ubuntu-20.04
-    needs: [docs-changed]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/dev_builds.yml
```diff
@@ -12,7 +12,6 @@ jobs:
 
   markdown-link-check:
     runs-on: ubuntu-20.04
-    needs: [docs-changed]
```
rnishtala-sumo, line 15: why did this break the dev builds?

swiatekm-sumo: Because the `docs-changed` job doesn't exist there.

rnishtala-sumo (approved):

PR #2724, opened 2022-12-22
feat: add metadata.metrics.config.extraProcessors

sumo-drosiek:

##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


deploy/helm/sumologic/values.yaml
```diff
@@ -3764,6 +3764,18 @@ metadata:
...
+      ## It can be used for filtering metrics, renaming, changing metadata and so on.
+      ## This is list of objects, for example:
+      ## extraProcessors:
+      ##   - filterprocessor:
+      ##       exclude:
+      ##         match_type: strict
+      ##         metric_names:
+      ##           - hello_world
+      ##           - hello/world
+      extraProcessors: []
```
kkujawa-sumo, line 3777: please add information about this new parameter to https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/README.md

tests/helm/metrics_test.go
```diff
@@ -53,3 +53,155 @@ metadata:
...
+	renameMetadatatatements := []string{
+		`set(attributes["k8s.pod.pod_name_new"], attributes["k8s.pod.pod_name"])`,
+		`delete_key(attributes, "k8s.pod.pod_name")`,
+		`set(attributes["my.static.value"], "<static_value>")`,
+	}
+	require.Equal(t, "resource", otelConfig.Processors.RenameMetadata.MetricStatements[0].Context)
+	require.Equal(t, renameMetadatatatements, otelConfig.Processors.RenameMetadata.MetricStatements[0].Statements)
+}
+
+func TestMetadataMetricsOtelConfigExtraProcessorsPipeline(t *testing.T) {
```
swiatekm-sumo, line 141: These two tests use the same configuration, could we merge them into one? The actual assertions in this test aren't very complex, adding them to the first test should be ok.

I'd also consider moving the valuesYaml to a file in `testdata/`, it's a bit large for an inline value imo.

deploy/helm/sumologic/values.yaml
swiatekm-sumo, line 3777: I wonder if this shouldn't live under `sumologic.metrics.metadata` instead. :thinking: It involves OT config internals, but I imagine it's also going to be one of the main ways of customizing the configuration for users.

docs/collecting-application-metrics.md
```diff
@@ -209,3 +220,127 @@ To scrape and forward exposed metrics to Sumo Logic, please follow one of the fo
...
+
+This section coverts the following metrics modifications:
+
+- [Filtering metrics](#filtering-metrics)
+- [Renaming metric](#renaming-metric)
+- [Adding or renaming metadata](#adding-or-renaming-metadata)
+
+### Filtering metrics
+
+In order to filter in or out the metrics, you can add [filterprocessor] to metric's extraProcessors.
```
swiatekm-sumo, line 234: Missing the link for `[filteprocessor]`.

docs/collecting-application-metrics.md
```diff
@@ -209,3 +220,127 @@ To scrape and forward exposed metrics to Sumo Logic, please follow one of the fo
 
 - [Application metrics are exposed (one endpoint scenario)](#application-metrics-are-exposed-one-endpoint-scenario)
 - [Application metrics are exposed (multiple enpoints scenario)](#application-metrics-are-exposed-multiple-enpoints-scenario)
+
+## Metrics modifications
```
swiatekm-sumo, line 224: Can you also update https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/docs/best-practices.md#excluding-metrics and other metrics sections in the best practices doc with this new info?

sumo-drosiek: Should I write more information there or linking to this document will be sufficient?

swiatekm-sumo: I'd provide some examples. It's structurally not the same thing either, as you may want to filter and modify non-application metrics.

deploy/helm/sumologic/values.yaml
sumo-drosiek: I was considering `metadata.metrics` as Otel specific configuration and extraProcessors is definitely Otel specific

swiatekm-sumo: I guess it's ok until we expose simpler higher-level ways of doing this, like namespace regexes and so forth.

tests/helm/metrics_test.go
sumo-drosiek: done

swiatekm-sumo (approved):

PR #2744, opened 2023-01-02
feat: adjust average utilization for metadata autoscaling

swiatekm-sumo:
##### Description

Adjust average utilization for metadata autoscaling down to 80, the current value of 100 is too aggressive.

Update the documentation to remove any remaining mentions of FluentD autoscaling as well.

---

##### Checklist

- [x] Changelog updated


sumo-drosiek (approved):

kkujawa-sumo:
Please test this change on OpenShift, I remember that for metics-server enabled on OpenShift we must disable creation of apiService, without
```
  metrics-server:
    enabled: true
    apiService:
      create: false
```

swiatekm-sumo:
It's not clear if we want to do this at all right now, there are arguments in favor and against, mostly having to do with the pain of ensuring metrics-server is installed.

swiatekm-sumo:
As per https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/2751, I'm going to switch autoscaling back to disabled by default, and only adjust the average utilization and update documentation.

rnishtala-sumo (approved):

PR #2750, opened 2023-01-04
feat: add otellogs.additionalDaemonSets configuration

sumo-drosiek:
##### Description

---

##### Checklist

- [ ] Changelog updated

###### Testing performed


deploy/helm/sumologic/values.yaml
```diff
@@ -4302,6 +4302,34 @@ otellogs:
     #     mountPath: /certs
     #     readOnly: true
 
+  ## additionalDaemonSets allows to set daemonsets with affinity, nodeSelector and resources
+  ## different than the main DaemonSet
+  ## Be careful and set nodeAffinity for the main DaemonSet,
+  ## as we do not support multiple pods of otellogs on the same node
```
sumo-drosiek, line 4308: @SumoLogic/open-source-collection-team 

Do you think it is safe assumption that if customer needs to set it, thew will know how to do it properly?
Or could we limit to `nodeSelector` and figure out `nodeAffinities` by ourselves?

astencel-sumo:
If we want to add this feature, we need to document it and provide examples. With proper documentation and the examples in place, I think we should be safe.

deploy/helm/sumologic/values.yaml
```diff
@@ -4302,6 +4302,34 @@ otellogs:
...
+  ##   affinity:
+  ##     nodeAffinity:
+  ##       requiredDuringSchedulingIgnoredDuringExecution:
+  ##         nodeSelectorTerms:
+  ##         - matchExpressions:
+  ##           - key: kubernetes.io/os
+  ##             operator: NotIn
+  ##             values:
+  ##             - linux
+  additionalDaemonSets: []
```
astencel-sumo, line 4331: How about a map of named daemonsets? This might make it easier to use this feature, having a daemonset named e.g. `otelcol-logs-collector-hpc` instead of `otelcol-logs-collector-1`. Just a suggestion, it has its ins and outs (like making the names too long).

sumo-drosiek: this is actually a good idea. I will work on it

deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml
```diff
@@ -1,59 +1,70 @@
...
+{{- if not (eq $name "") }}
+{{- $instance = (printf "-%s" $name ) }}
+{{- end }}
+{{- $resources := deepCopy $.Values.otellogs.daemonset.resources | merge .resources }}
+---
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: {{ template "sumologic.metadata.name.logs.collector.daemonset" . }}
+  name: {{ printf "%s%s" (include "sumologic.metadata.name.logs.collector.daemonset" $ctx) $instance }}
```
swiatekm-sumo, line 15: You should truncate this to 63 chars, since it's derived from user input. Same with the labels.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: Let them figure it out imo, they should know what they're doing if they need to use this.

docs/best-practices.md
```diff
@@ -632,6 +633,46 @@ fluent-bit:
     kubernetes.io/os: linux
 ```
 
+#### Setting different resources on different nodes for logs collector
+
+This section covers scenario, when you would like to set different resources
+for OpenTelemetry logs collector depending on the node labels.
```
swiatekm-sumo, line 639:
```suggestion
All of the log collector Pods have the same resource settings, being members of the same DaemonSet. However, sometimes it's necessary to vary this based on the Node - for example when some large Nodes can contain applications generating a lot of log data.

It' possible to set different resource quotas for different Nodes based on labels, resulting in a different log collector DaemonSet for these Nodes.
```

swiatekm-sumo (commented):
I'm not actually sure if this should be in best-practices, or maybe in troubleshooting instead?

sumo-drosiek:
> I'm not actually sure if this should be in best-practices, or maybe in troubleshooting instead?

I think it could be in both

swiatekm-sumo (approved):

sumo-drosiek:
false positive markdown link check (checked manually)

PR #2753, opened 2023-01-04
docs: remove remaining Fluentd mentions from main docs

swiatekm-sumo:
This PR does the following:
* rename `FLUENTD_METRICS_SVC` to `METADATA_METRICS_SVC` as per #2627
* remove the additional Prometheus configuration doc, as it's superseded by #2716 and https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/docs/collecting-application-metrics.md
* update a bunch of random references to FluentD, where it should be OT now


docs/installing-behind-proxy.md
```diff
@@ -5,9 +5,10 @@ If your Kubernetes cluster requires outbound traffic to go through a proxy, you
 ```
 sumologic.httpProxy
 sumologic.httpsProxy
-fluentd.proxyUri
 ```
 
+If you're using FluentD as the metadata provider, you should also set `fluentd.proxyUri`.
```
astencel-sumo, line 10:
```suggestion
If you're using Fluentd as the metadata provider, you should also set `fluentd.proxyUri`.
```

docs/security-best-practices.md
```diff
@@ -19,8 +19,8 @@ Logically, the pipeline for a data type is composed of the following components
 Collector -> Metadata Enrichment -> Sumologic Backend
 ```
 
-Different applications may serve the same role here - for example, FluentD is currently used for metadata enrichment
-for both logs and metrics, but the plan is to replace it with the Opentelemetry Collector in the near future. Nonetheless,
+Different applications may serve the same role here - for example, OpenTelemetry Collector is currently used for metadata enrichment
+for both logs and metrics, but it used to be FluentD in v2 of the Chart. Nonetheless,
```
astencel-sumo, line 23:
```suggestion
for both logs and metrics, but it used to be Fluentd in v2 of the Chart. Nonetheless,
```

astencel-sumo (approved):

PR #2761, opened 2023-01-05
fix(logs): remove unnecessary metadata

swiatekm-sumo:
The specific changes are:
* Remove the `fluent.tag` record attribute from container logs sent via OTLP
   We don't use it in that pipeline, so it's not necessary to set it in the first place.
* Remove `k8s.pod.uid` and `run_id` record attributes
   These aren't that useful, and we eventually want to send the logs as unstructured.
* Remove `pod_id`, `container_id` and `replicaset` resource attributes
   We don't have fields set for these, and they aren't very useful in any case.
* Remove `k8s.pod.pod_name` resource attribute
   This is just an artificial field set by sourceprocessor. Metrics already drop it, the fact that logs didn't was a bug.

I also made the text diffs produced by template tests for configmaps nicer.


deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
```diff
@@ -188,15 +184,10 @@ receivers:
...
         to: attributes["k8s.pod.name"]
-      - type: move
-        from: attributes.run_id
-        to: attributes["run_id"]
-      - type: move
-        from: attributes.uid
-        to: attributes["k8s.pod.uid"]
-      - type: add
-        field: attributes["fluent.tag"]
-        value: EXPR("containers." + attributes["k8s.container.name"])
```
sumo-drosiek, line 199: Just to not forget about `- filter/include_containers`

swiatekm-sumo: That checks the container name:
```yaml
  filter/include_containers:
    logs:
      include:
        match_type: regexp
        record_attributes:
          - key: k8s.container.name
            value: .+
```

sumo-drosiek (approved):

PR #2773, opened 2023-01-10
feat(logs): parse JSON logs

swiatekm-sumo:
Parse JSON logs bodies into OTLP data structures. As a result, Sumo renders the JSON correctly. Without this change, JSON log bodies are just escaped strings.

This change is somewhat dangerous, and I'm wondering if we shouldn't make it possible to disable. The problem is that currently, if the JSON parsing returns an error, it'll result in a 500 returned to the collector, and potentially block the whole batch from being processed indefinitely. Discussion is ongoing in [otel upstream](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16519
) about how the errors should be handled, but right now there doesn't seem to be a way to just ignore parse errors.

I've switched to logstransform processor, and this actually lets the malformed data through while logging an error. It seems to be buggy in `0.68.0` though, and fine in `0.69.0`, so this PR shouldn't be merged until we upgrade to the latter.

### DIFF
--- a/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
@@ -205,6 +205,14 @@ processors:
     container_annotations:
       enabled: {{ .Values.sumologic.logs.container.perContainerAnnotationsEnabled }}
       prefixes: {{ toJson .Values.sumologic.logs.container.perContainerAnnotationPrefixes }}
+  ## logstransformprocessor is experimental, but the same functionality in transformprocessor doesn't offer robust enough error handling
+  ## TODO: use transformprocessor here after https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16519 is closed
+  logstransform/containers_parse_json:
+    operators:
+      - type: json_parser
+        parse_from: body
+        parse_to: body
+        if: body matches "^{.*"
 {{ end }}
 {{ if .Values.sumologic.logs.systemd.enabled }}
   ## Systemd related processors
@@ -421,6 +429,7 @@ service:
         - k8s_tagger
         - resource/add_cluster
         - source/containers
+        - logstransform/containers_parse_json
         - resource/remove_k8s_pod_pod_name
         - resource/drop_annotations
         - resource/containers_copy_node_to_host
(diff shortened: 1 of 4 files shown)


sumo-drosiek:
> Discussion is ongoing in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16519 about how the errors should be handled, but right now there doesn't seem to be a way to just ignore parse errors.

If so, we should make it configurable or move to separate processor eventually

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -205,6 +205,14 @@ processors:
...
       enabled: {{ .Values.sumologic.logs.container.perContainerAnnotationsEnabled }}
       prefixes: {{ toJson .Values.sumologic.logs.container.perContainerAnnotationPrefixes }}
+  ## logstransformprocessor is experimental, but the same functionality in transformprocessor doesn't offer robust enough error handling
+  ## TODO: use transformprocessor here after https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16519 is closed
+  logstransform/containers_parse_json:
+    operators:
+      - type: json_parser
+        parse_from: body
+        parse_to: body
+        if: body matches "^{.*"
```
sumo-drosiek, line 215: Does it produce any logs in case of parsing error? How many?

swiatekm-sumo: It generates a strack trace.

sumo-drosiek: Hmm, log collector's logs are not json so this is going to be logs generator at some point

sumo-drosiek: but this is in metadata logs, so we just need to ensure we filter it out before, especially because we don't treat it as error

swiatekm-sumo: I don't understand this point, this only applies to logs which start with a curly brace.

sumo-drosiek: oh, that's right 🤦

sumo-drosiek: LGTM then

swiatekm-sumo:
Depends on https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2791

sumo-drosiek (approved):

PR #2779, opened 2023-01-11
docs: remove non-existent link

astencel-sumo:
We don't have any docs on removing metrics in Prometheus.

This non-existent link managed to get through the markdown-link-check because we now only check modified files,
so when a file is deleted or renamed,
links to it in other files are not checked.

### DIFF
--- a/deploy/docs/additional_prometheus_configuration.md
+++ b/deploy/docs/additional_prometheus_configuration.md
@@ -1,3 +1,3 @@
 # Additional Prometheus Configuration
 
-Moved to [/docs/additional-prometheus-configuration.md](/docs/additional-prometheus-configuration.md)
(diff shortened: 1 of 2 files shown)


deploy/docs/additional_prometheus_configuration.md
```diff
@@ -1,3 +1,3 @@
 # Additional Prometheus Configuration
 
-Moved to [/docs/additional-prometheus-configuration.md](/docs/additional-prometheus-configuration.md)
+Moved to [/docs/collecting-application-metrics.md](/docs/collecting-application-metrics.md)
```
sumo-drosiek, line 3: Why `additional-prometheus-configuration.md` has been removed?

sumo-drosiek: ok, I see it was planned

sumo-drosiek (approved):

PR #2794, opened 2023-01-13
feat(logs): add format setting

swiatekm-sumo:
Move the `logFormat` setting for Fluentd to `sumologic.logs.container.format`, and make both Fluentd and otel use it.

This requires a migration, which I will add shortly.


docs/collecting-container-logs.md
```diff
@@ -57,6 +57,73 @@ This feature is enabled by default and the default regex will catch logs startin
...
+```text
+2007-03-01T13:00:00Z I am a log line
+```
+
+```javascript
+{
+  log_property: "value",
+  text: "I am a json log"
+}
+```
```
sumo-drosiek, line 125: Could you add example without any property as well?

sumo-drosiek:
There is one issue related to the text format: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/docs/fluent/troubleshoot-collection.md#using-text-format

I don't know why it's only in fluentd troubleshooting though

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -1576,3 +1576,18 @@ Generate list of remoteWrite endpoints for telegraf configuration
 {{- $endpoints := sortAlpha $endpoints -}}
 {{ $endpoints | join ",\n" }}
 {{- end -}}
+
+{{/*
+Return the log format for the Sumologic exporter for container logs
+
+'{{ include "metric.endpoints" . }}'
```
astencel-sumo, line 1583:
```suggestion
'{{ include "logs.otelcol.container.exporter.format" . }}'
```

deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -1576,3 +1576,18 @@ Generate list of remoteWrite endpoints for telegraf configuration
...
+*/}}
+{{- define "logs.otelcol.container.exporter.format" -}}
+{{- if eq .Values.sumologic.logs.container.format "json" -}}
+{{- "json" -}}
+{{- else if eq .Values.sumologic.logs.container.format "text" -}}
+{{- "text" -}}
+{{- else -}}
+{{- fail "`sumologic.logs.container.format` can only be `json` or `text`" -}}
```
astencel-sumo, line 1591: Why not allow users to set `otlp`?

astencel-sumo: Oh in that case I think we should come up with a better name for it.

Still I believe we need to make it possible for users to send data in OTLP sources (after they have them enabled for their org), but it's a different feature than this one (although backed by the same setting in sumologic exporter).

swiatekm-sumo: There's actually no difference in terms of wire format between `text` and `json` in the Sumo exporter. `text` simply ignores record attributes and just sends the body, while `json` sends a json document with keys containing record attributes and body. But fundamentally you're still sending a block of text with records separated by newlines.

docs/collecting-container-logs.md
swiatekm-sumo: Do you think that would be helpful? I've already got four examples, fitting in two more makes the whole section less legible in general.

deploy/helm/sumologic/templates/_helpers.tpl
astencel-sumo: I still think we should allow `otlp` :wink:

astencel-sumo (approved):

docs/collecting-container-logs.md
swiatekm-sumo: I rewrote this whole section while adding `json_merge`.

swiatekm-sumo:
Migration: https://github.com/SumoLogic/sumologic-kubernetes-tools/pull/432

astencel-sumo (approved):

PR #2796, opened 2023-01-16
feat: add pvc-cleaner to the chart

### LINKED ISSUE #2602: Provide information on using the pvc cleaner script
Kubernetes cannot delete persistent volumes for statefulsets yet.
It's available as [alpha feature since 1.23 and should be available as beta since 1.26](https://github.com/kubernetes/enhancements/issues/1847)

This creates problems in EKS when new instance of the pod is created in different AZ than initial one as PVs cannot be mounted across different AZs.

Until that time we need troubleshooting on how to use the [pvc cleaner script](https://github.com/SumoLogic/sumologic-kubernetes-tools/tree/main/src/examples) updated to be used with OTC.

aboguszewski-sumo:
This PR adds pvc-cleaner cronjobs to the helm chart. It's based on what @sumo-drosiek created on [this branch](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/drosiek-pvc-cleaner).

Things done:
- added additional config to `values.yaml`
- added helm templates for cronjobs and service account
- created helm template test
- manually tested that it works
- add docs (readme for `values.yaml`, description in `best-practices.md`)
- add integration test

Resolves #2602


deploy/helm/sumologic/templates/pvc-cleaner/cron-job.yaml
```diff
@@ -0,0 +1,121 @@
+{{- if eq (dig "pvc-cleaner" "logs" "enabled" false (.Values | merge (dict))) true }}
```
aboguszewski-sumo, line 1: The original solution provided by Dominik uses some `dig` function in many places, imo it should be replaced by `.Values.pvc-cleaner.*`, but to be honest I have no idea what's the best option.

deploy/helm/sumologic/templates/pvc-cleaner/cron-job.yaml
```diff
@@ -0,0 +1,121 @@
...
+kind: CronJob
+metadata:
+  name: {{ template "sumologic.metadata.name.pvc-cleaner.logs" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "sumologic.labels.app.pvc-cleaner.logs" . }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+spec:
+  schedule: "*/15 * * * *"
+  jobTemplate:
```
aboguszewski-sumo, line 12: The schedule should probably be configurable, this would be especially useful for potential integration tests.

deploy/helm/sumologic/templates/pvc-cleaner/cron-job.yaml
sumo-drosiek: `.Values.pvc-cleaner.*` doesn't work as `-` is not recognizable properly by helm
we can either use `get` and assing `pvc-cleaner` to value or use `dig` as above

deploy/helm/sumologic/values.yaml
```diff
@@ -4503,3 +4503,49 @@ opentelemetry-operator:
 
   manager:
     env:
+
+pvc-cleaner:
+  metrics:
+    enabled: true
+  logs:
+    enabled: true
```
perk-sumo, line 4511: It should be disabled by default (per this internal discussion: https://docs.google.com/document/d/1qC471RG2YTKw3PjDaqtQ6Rt9U8UaR3yeFwkqV0ZuplA/edit#heading=h.e6ctt1cf9j1s)

deploy/helm/sumologic/templates/pvc-cleaner/serviceaccount.yaml
```diff
@@ -0,0 +1,58 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ template "sumologic.metadata.name.pvc-cleaner.roles.serviceaccount" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "sumologic.metadata.name.pvc-cleaner.roles.serviceaccount" . }}
+subjects:
```
aboguszewski-sumo, line 55: This was based on internal setup. Strangely enough, this wasn't necessary when tested manually, but integration tests didn't work without this role.

aboguszewski-sumo:
This is ready to review now, apart from one thing:
API for CronJobs can't be determined now. k8s versions earlier than `v1.21` use `batch/v1beta1`, which is deprecated from `v1.21` and removed in `v1.25`. Versions from `v1.21` should use `batch/v1` instead. My suggestion is to use something similar to what bitnami does: https://github.com/bitnami/charts/blob/cce758c7135d488c8b3eebde7be8211d6b9a2a8a/bitnami/common/templates/_capabilities.tpl#L73-L94

swiatekm-sumo:
> 

Yeah, we already do the same thing for PDBs: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/templates/_api_versions_helpers.tpl

aboguszewski-sumo:
Integration tests are failing because of insufficient CPU, I'll try to merge the new test into one of the older ones.

aboguszewski-sumo:
Integration tests work after lowering the cpu requests in `values.yaml` for tests.

deploy/helm/sumologic/values.yaml
```diff
@@ -4503,3 +4503,43 @@ opentelemetry-operator:
 
   manager:
     env:
+
+pvc-cleaner:
```
swiatekm-sumo, line 4507: As per Helm's best practices, this should be camelcase. This will also solve the problem of referring to these values elsewhere. See: https://helm.sh/docs/chart_best_practices/values/#naming-conventions

I'm not sure why helm lint doesn't catch this.

tests/helm/pvc-cleaner/config.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+export TEST_TEMPLATE="templates/pvc-cleaner/job.yaml"
```
swiatekm-sumo, line 3: You don't need this file with the new test framework.

aboguszewski-sumo:
I applied changes after review.

swiatekm-sumo (approved):
LGTM

PR #2797, opened 2023-01-16
docs(logs): add section on adding fields

astencel-sumo:


docs/collecting-container-logs.md
```diff
@@ -205,6 +207,40 @@ sumologic:
...
+      otelcol:
+        extraProcessors:
+        - resource/add-k8s_app-field:
+            attributes:
+            - action: insert
+              key: k8s_app
+              from_attribute: pod_labels_app.kubernetes.io/name
+```
+
+**NOTE** Make sure the field is [added in Sumo Logic][sumo_add_fields].
```
sumo-drosiek, line 242: Will that work if we add it to our configuration for field creation?

astencel-sumo: Yes it will, you just need to duplicate the whole list of fields...

```yaml
sumologic:
  logs:
    fields:
    - cluster
    - container
    - daemonset
    - deployment
    - host
    - namespace
    - node
    - pod
    - service
    - statefulset
    - k8s_app
```

sumo-drosiek: I would consider adding additional fields in the future 🤔

swiatekm-sumo: There's a chance we won't have to add the basic k8s metadata fields at all, in which case the problem will solve itself to an extent.

swiatekm-sumo (approved):

Issue #2767, opened 2023-01-09
Document that in case of fluentd metadata, source configuration has to be changed


sumo-drosiek:
Default source configuration: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v3.0.0-beta.1/deploy/helm/sumologic/values.yaml#L256-L265

is incompatible with Fluentd. We need to document it

rnishtala-sumo:
Let's also document that if kube-state-metrics or any other pod/service that we pull metadata from crashloops, we will run into source naming issues.

perk-sumo:
We need to check those potentially problematic cases:
* custom fluentd config in v2 -> copied over to OTC in v3
* defaults for v3 with OTC -> switching to Fluentd

sumo-drosiek:
fluentd plugin related code: https://github.com/SumoLogic/sumologic-kubernetes-fluentd/blob/3a5a2e29e3776107a1bf0c9435de5cc2b14fc7f6/fluent-plugin-kubernetes-sumologic/lib/fluent/plugin/filter_kubernetes_sumologic.rb#L111-L118

PR #2800, opened 2023-01-17
test(integration): check log metadata values

swiatekm-sumo:
Check that the log metadata values are correct in integration tests. For FluentD, this shows a bug which I'm going to fix in a separate change, related to https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/2767.

I've also enabled more linters on the integration tests while at it.


tests/integration/helm_fluentbit_fluentd_test.go
```diff
@@ -327,9 +325,9 @@ func Test_Helm_FluentBit_Fluentd(t *testing.T) {
 			10, // we don't really control this, just want to check if the logs show up
 			map[string]string{
 				"cluster":         "kubernetes",
-				"_sourceName":     "",
+				"_sourceName":     "(?!undefined$).*",
```
aboguszewski-sumo, line 328: I guess that since we started adding constants for regexes, this one could get moved to a constant, too.

tests/integration/helm_fluentbit_fluentd_test.go
```diff
@@ -300,22 +297,23 @@ func Test_Helm_FluentBit_Fluentd(t *testing.T) {
...
+				"container":      internal.LogsGeneratorName,
+				"deployment":     internal.LogsGeneratorName,
+				"replicaset":     fmt.Sprintf("%s%s", internal.LogsGeneratorName, "-[a-z0-9]{9,10}"),
+				"namespace_id":   ".*",
+				"pod":            fmt.Sprintf("%s%s", internal.LogsGeneratorName, internal.PodDeploymentSuffixRegex),
+				"pod_id":         ".*",
+				"container_id":   ".*",
+				"host":           internal.NodeNameRegex,
+				"master_url":     ".*",
+				"node":           internal.NodeNameRegex,
```
aboguszewski-sumo, line 312: Does changing from `""` to `".*"` change anything?
IIRC, the receiver mock still supports empty strings as "anything" - is it a preparation for removing this?
If not and you wanted nonempty strings, then `".+"` should be used.

aboguszewski-sumo (approved):
This looks fine, one question though:
`Test_Helm_Default_OT_FIPS_Metadata` also checks for some logs metadata, why wasn't it changed?

tests/integration/helm_fluentbit_fluentd_test.go
swiatekm-sumo: Doesn't change anything, but it's clearer this way. But you're right, it should be `.+`, I'll change that.

swiatekm-sumo:
> Test_Helm_Default_OT_FIPS_Metadata also checks for some logs metadata, why wasn't it changed?

Because I forgot, will fix. Thought this is an argument in favor of deduplicating it.

aboguszewski-sumo (approved):

Issue #2803, opened 2023-01-18
Multiline logs have newlines stripped


astencel-sumo:
Version: v3.0.0-beta.1
Container runtime: containerd

**Describe the bug**

When running the chart v3 with the default configuration in a cluster with containerd container runtime, multiline logs end up in Sumo without newlines.

**Steps to reproduce:**

Run the collection with default settings in a k8s cluster with containerd container runtime.

values.yaml:

```yaml
sumologic:
  accessId: xxx
  accessKey: yyy
  clusterName: zzz
```

**Expected output:**

Multiline logs from containers should contain newlines:

```
{
    log:"2001-02-03 04:05:06 logger 1 log 2055 first line
      logger 1 log 2055 second line
      logger 1 log 2055 third line",
    stream:"stdout",
    timestamp:1674041362246
}
```

**Actual output:**

Logs from containers do not have newlines:

```
{
    log:"2001-02-03 04:05:06 logger 1 log 1252 first line  logger 1 log 1252 second line  logger 1 log 1252 third line",
    stream:"stdout",
    timestamp:1674041362246
}
```

![image](https://user-images.githubusercontent.com/70892616/213165219-a0e95de1-7953-41d9-ad15-659359004741.png)

PR #2805, opened 2023-01-18
fix(logs): correctly handle newlines

### LINKED ISSUE #2802: Logs from Docker container runtime have a trailing newline
Version: v3.0.0-beta.1

**Describe the bug**

When running the chart v3 with the default configuration in a cluster with Docker container runtime, logs end up in Sumo with a newline at the end of the `log` key.

**Steps to reproduce:**

Run the collection with default settings in a k8s cluster with Docker container runtime.

values.yaml:

```yaml
sumologic:
  accessId: xxx
  accessKey: yyy
  clusterName: zzz
```

**Expected output:**

Logs from containers should not have a trailing newline

```
{
    log:"some log",
    stream:"stdout",
    timestamp:1674039746682
}
```

**Actual output:**

Logs from containers should not have a trailing newline

```
{
    log:"some log
    ",
    stream:"stdout",
    timestamp:1674039746682
}
```

image

astencel-sumo:
Fixes #2802, #2803

### DIFF
--- a/tests/helm/logs_otc/static/basic.output.yaml
+++ b/tests/helm/logs_otc/static/basic.output.yaml
@@ -94,7 +94,7 @@ data:
           combine_with: ""
           id: merge-docker-lines
           is_last_entry: body.log matches "\n$"
-          output: merge-multiline-logs
+          output: strip-trailing-newline
           source_identifier: attributes["log.file.path"]
           type: recombine
         - combine_field: body.log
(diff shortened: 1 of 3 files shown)


tests/helm/logs_otc/static/basic.output.yaml
```diff
@@ -105,8 +105,17 @@ data:
...
+          parse_from: body.log
+          parse_to: body
+          regex: |-
+            ^(?P<log>.*)
+            $
+          type: regex_parser
         - combine_field: body.log
-          combine_with: ""
+          combine_with: |2+
+
```
swiatekm-sumo, line 118: This looks pretty weird, why does it get interpreted as a special character?

swiatekm-sumo (approved):

tests/helm/logs_otc/static/basic.output.yaml
astencel-sumo: This is just regular Yaml :wink: 

I looked it up and it seems to be a combination of the ["keep" chomping indicator](https://yaml.org/spec/1.2.2/#8112-block-chomping-indicator) `+` and the ["explicit" indentation indicator](https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator) `2`.

PR #2806, opened 2023-01-18
chore: upgrade yq to 4.30.8

swiatekm-sumo:
There are some instructions in the documentation that still use `yq` version 3, but those also explicitly say to install that version, so they're not blocking this change.

### DIFF
--- a/vagrant/scripts/yq/avalanche-remote-write.yaml
+++ b/vagrant/scripts/yq/avalanche-remote-write.yaml
@@ -1,0 +1,9 @@
+kube-prometheus-stack:
+  prometheus:
+    prometheusSpec:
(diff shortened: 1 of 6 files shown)


vagrant/scripts/yq/avalanche-remote-write.yaml
```diff
@@ -0,0 +1,9 @@
+kube-prometheus-stack:
```
astencel-sumo, line 1: Do these two avalanche files belong in this PR?

astencel-sumo: Oh I guess they do

astencel-sumo (approved):

PR #2808, opened 2023-01-19
fix(docs): add missing changelog entry for EKS 1.20 support removal

aboguszewski-sumo:
I did not realize that missing changelog does not block automerging.


perk-sumo (approved):
Thank you!

PR #2810, opened 2023-01-19
chore(tests/integration): update supported k8s versions

aboguszewski-sumo:
1.19 and 1.20 are no longer supported, so they have been removed from tests. 
1.23 and 1.24 are now supported, so they have been added to the tests.


swiatekm-sumo:
Can you add `kind_images.json` to the list of files checked for changes before tests are run? They should be run for this PR.

astencel-sumo (approved):

PR #2826, opened 2023-01-24
docs(troublehooting): minor changes to prometheus snippets

sumo-drosiek:
Add step to figure out the prometheus name


swiatekm-sumo (approved):

swiatekm-sumo:
Can't we use the label directly in `kubectl logs` instead?

PR #2827, opened 2023-01-24
chore: use prettier for formatting and linting yaml

swiatekm-sumo:
Use [prettier](https://prettier.io/docs/en/configuration.html) for formatting and linting yaml. The basic idea here is to consolidate some of our tooling and make it more accessible. Prettier is a standard tool heavily used in the js world, and is therefore excellently supported in general. I'm hoping to use it for Markdown formatting as well.

The formatting rules are opinionated and not configurable, so I was hoping to get some feedback for the yaml changes:
* quoting is now consistently using double quotes
* lists are indented
* comments have consistent indentation

I've also set it to run on *all* the yaml files in the repo. Not completely sure it's the right choice, but I'd like to try and cover as much as I can, and scale back if needed. Let me know what you think.

### Checklist


deploy/helm/sumologic/values.yaml
```diff
@@ -1183,7 +1183,8 @@ fluent-bit:
     pullPolicy: IfNotPresent
 
   ## Resource limits for Fluent Bit
-  resources: {}
+  resources:
+    {}
```
sumo-drosiek, line 1187: not really like it

examples/kube_prometheus_stack/values-prometheus.yaml
```diff
@@ -497,7 +499,8 @@ prometheus-node-exporter:
   podLabels: {}
   ## Additional annotations for pods in the DaemonSet
   podAnnotations: {}
-  resources: {}
+  resources:
+    {}
```
sumo-drosiek, line 503: why is it formatted differently?
because of comments below?

swiatekm-sumo: Yeah, I think so.

deploy/helm/sumologic/conf/logs/fluentd/logs.kubernetes.sumologic.filter.conf
```diff
@@ -8,4 +8,4 @@ exclude_pod_regex {{ .Values.sumologic.logs.container.excludePodRegex | quote }}
 exclude_container_regex {{ .Values.sumologic.logs.container.excludeContainerRegex | quote }}
 exclude_host_regex {{ .Values.sumologic.logs.container.excludeHostRegex | quote }}
 per_container_annotations_enabled {{ .Values.sumologic.logs.container.perContainerAnnotationsEnabled }}
-per_container_annotation_prefixes {{ join "," .Values.sumologic.logs.container.perContainerAnnotationPrefixes }}
+per_container_annotation_prefixes {{ join "," .Values.sumologic.logs.container.perContainerAnnotationPrefixes | default "[]" }}
```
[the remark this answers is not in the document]
swiatekm-sumo: Indirectly. By default, we have an empty space here, so the output is a text block ending with: `per_container_annotation_prefixes `. Prettier wants to remove what it sees as a trailing space from it. Instead of excluding the file from formatting, I changed the default format - FluentD accepts lists both comma-delimited and json - and I think it's better this way.

astencel-sumo:
This change replaces `yamllint` Python tool with `prettier` NodeJS tool. Does it make things much better?

I suppose the upside is that the `prettier` tool can not only lint yaml, but also format it?

swiatekm-sumo:
> This change replaces `yamllint` Python tool with `prettier` NodeJS tool. Does it make things much better?
> 
> I suppose the upside is that the `prettier` tool can not only lint yaml, but also format it?

Yes, and it also has mature VS Code and IntelliJ (and any other popular code editor, though I haven't checked all of them) integration.

It can also format Markdown, which will allow us to replace two more tools. Markdown formatting has historically been quite a pain, at least for me.

examples/multiline_logs/pod_multiline_single_long_line.yaml
```diff
@@ -4,22 +4,23 @@ metadata:
...
+      args:
+        - /bin/sh
+        - -c
+        - >
+          while true; do
+            LONG_STRING="$(cat /dev/urandom | tr -dc ''a-z0-9'' | head -c 30000)"
+            echo "2022-02-08T20:53:28.621+0000 1st single line..."
+            echo "2022-02-08T20:53:29.622+0000 2nd single line..."
+            echo "2022-02-08T20:53:30.623+0000 Exception in thread "main" java.lang.RuntimeException: Something has gone wrong, aborting!
+          ${LONG_STRING} end of the 1st long line
```
sumo-drosiek, line 18: Is it correct?

swiatekm-sumo: It's a correct transformation, because this yaml block uses `>`, which turns all newlines into spaces: https://yaml.org/spec/1.2.2/#813-folded-style. It should use `|` instead - then prettier leaves it alone.

sumo-drosiek (approved):

astencel-sumo (approved):

PR #2828, opened 2023-01-24
chore: use prettier for formatting and linting Markdown

swiatekm-sumo:
Use prettier to format and link Markdown. This a sister PR to #2827, the same tool applied to different document types. I set a conservative max line length of 140, but I'm open to changing that.

I ended up keeping `markdownlint`, as it can check certain things prettier doesn't, but I'm not sure if the added value is enough. Either way, markdown-table-formatter was removed.

One interesting feature of prettier is that it formats code in Markdown code blocks, so now all of our yaml snippets are formatted the same as in #2827.

### Checklist

- [X] Documentation updated


CHANGELOG.md
```diff
@@ -399,9 +399,9 @@ This change affects you only if you have enabled traces with `sumologic.traces.e
 
 ### Released 2022-07-06
 
-This release changes the OpenTelemetry Collector `Traces` service endpoint `CHART_NAME-sumologic-otelcol.NAMESPACE`
-to `deprecated`. Service will still work and point to `CHART_NAME-sumologic-otelagent.NAMESPACE`.
-This change affects you only if you have enabled traces with `sumologic.traces.enabled: true`.
+This release changes the OpenTelemetry Collector `Traces` service endpoint `CHART_NAME-sumologic-otelcol.NAMESPACE` to `deprecated`. Service
```
sumo-drosiek, line 402: `. Service\n` hurts my eyes :dagger:

sumo-drosiek:
TBH, I'm not happy about MDL formatting, its counterintuitive for me

sumo-drosiek:
> (or reading the source anywhere where you don't have a line wrapping editor),

Personally I prefer to read such documents formatted the way I pointed out

I looked through the internet and there is no such rule for English language, but it is recognized as good practice by part of society

sumo-drosiek:
> > > (or reading the source anywhere where you don't have a line wrapping editor),
> > > Personally I prefer to read such documents formatted the way I pointed out
> 
> Yeah, it would be better, but I'm not aware of a formatter that does this. And I think the state in this PR is still better than the inconsistent formatting we have right now. Did you find a lot of offending lines?

I think I would have to leave with it.

kkujawa-sumo (approved):

astencel-sumo:
> > > > (or reading the source anywhere where you don't have a line wrapping editor),
> > > > Personally I prefer to read such documents formatted the way I pointed out
> > 
> > 
> > Yeah, it would be better, but I'm not aware of a formatter that does this. And I think the state in this PR is still better than the inconsistent formatting we have right now. Did you find a lot of offending lines?
> 
> I think I will have to live with it.

The best thing about using `prettier` is that we can just stop having discussions like this :smile:

PR #2842, opened 2023-01-31
docs(troubleshoot): add solution for dns issue in otellogs

sumo-drosiek:

### Checklist

- [ ] Documentation updated

### DIFF
--- a/docs/security-best-practices.md
+++ b/docs/security-best-practices.md
@@ -434,3 +434,9 @@ otelgateway:
 ### FIPS compliant images for Fluent Bit and Fluentd
 
 This Helm Chart does not provide FIPS compliant binaries for **Fluent Bit** nor **Fluentd**.
+
+### Troubleshooting
+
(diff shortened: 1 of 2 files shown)


swiatekm-sumo:
Under what circumstances does this happen exactly? Should we recommend setting this by default when using FIPS images, or maybe set it in the FIPS build?

sumo-drosiek:
I would set it in the FIPS build, but until that we should have document it.

It happens for `alpine` with statically linked go libraries

sumo-drosiek:
I saw this error every time using `0.69.0-fips` (vagrant, customer environment)

swiatekm-sumo:
I'm wondering if we should add this to troubleshooting docs, given that we're going to enable this option for all FIPS builds.

sumo-drosiek:
Documentation is going to be much faster available for customers than release and helm chart change

swiatekm-sumo (approved):

PR #2846, opened 2023-02-06
chore: upgrade shellcheck to 0.9.0

swiatekm-sumo:
Upgrading shellcheck, really messy because of how we handle the setup scripts. I'm not 100% sure yet if adding error checking (`set -euo pipefail`) is correct for all of these scripts, I'm keeping this as draft until I verify.

### Checklist

- [X] Changelog updated or skip changelog label added

PR #2849, opened 2023-02-07
fix(build): Helm Chart push script

swiatekm-sumo:
Fixing an error that crept into #2846 

### Checklist

- [X]  Changelog updated or skip changelog label added


sumo-drosiek (approved):

sumo-drosiek:
should be `$?` to be in line with shellcheck check?

PR #2850, opened 2023-02-08
chore: use Towncrier for changelog generation

swiatekm-sumo:
Use [Towncrier](https://towncrier.readthedocs.io/en/latest/index.html) for automatic changelog generation from entry files kept in `.changelog`. The specifics are described in the documentation updates.

For the current changelog, this is the block added to CHANGELOG.md after running `make update-changelog VERSION=3.1.0`:

```markdown
## [v3.1.0]

### Released 2023-02-08

### Changed

- chore: add support for EKS 1.24 [#2831]
- chore: add support for GKE 1.24 [#2832]
- feat(otelcol): add max size in batch processors [#2839]
- chore: add support for GKE 1.25 [#2848]
- chore: remove support for KOPS 1.21 [#2848]
- chore: add support for AKS 1.25 [#2848]
- chore: add support for Openshift 4.11 [#2848]
- chore: add support for KOPS 1.25 [#2848]

[#2831]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pulls/2831
[#2832]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pulls/2832
[#2839]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pulls/2839
[#2848]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pulls/2848
[v3.1.0]: https://github.com/SumoLogic/sumologic-kubernetes-collection/releases/v3.1.0
```

One difference between this generated changelog and our current manual process, is that I couldn't get Towncrier to output a github compare link, as it has no concept of a "previous version". We could fix this separately, but I'm not sure it's worth the trouble. Thoughts?

### Checklist

- [x] Documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,14 +94,10 @@ make test-integration:
 # Changelog management
 ## We use Towncrier (https://towncrier.readthedocs.io) for changelog management
 
-ENTRY_TEXT ?= 'Use [Semantic Versioning](https://semver.org/spec/v2.0.0.html) for your changelog entry'
-CHANGE_ID ?= 9999
-CHANGE_TYPE ?= changed
-
-## Usage: make add-changelog-entry CHANGE_TYPE=changed
+## Usage: make add-changelog-entry
 .PHONY: add-changelog-entry
 add-changelog-entry:
-	towncrier create -c "$(ENTRY_TEXT)" "(CHANGE_ID).(CHANGE_TYPE).txt"
+	./ci/add-changelog-entry.sh
 
 ## Consume the files in .changelog and update CHANGELOG.md
 ## We also format it afterwards to make sure it's consistent with our style
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,35 @@ make test-integration:
 	make -C ./tests/integration test
 
 
+# Changelog management
+## We use Towncrier (https://towncrier.readthedocs.io) for changelog management
+
+ENTRY_TEXT ?= 'Use [Semantic Versioning](https://semver.org/spec/v2.0.0.html) for your changelog entry'
+CHANGE_ID ?= 9999
+CHANGE_TYPE ?= changed
+
+## Usage: make add-changelog-entry CHANGE_TYPE=changed
+.PHONY: add-changelog-entry
+add-changelog-entry:
+	towncrier create -c "$(ENTRY_TEXT)" "(CHANGE_ID).(CHANGE_TYPE).txt"
+
+## Consume the files in .changelog and update CHANGELOG.md
+## We also format it afterwards to make sure it's consistent with our style
+## Usage: make update-changelog VERSION=x.x.x
+.PHONY: update-changelog
+update-changelog:
+ifndef VERSION
+	$(error Usage: make update-changelog VERSION=x.x.x)
+endif
+	towncrier build --yes --version $(VERSION)
+	prettier -w CHANGELOG.md
+	git add CHANGELOG.md
+
+## Check if the branch relative to main adds a changelog entry
+.PHONY: check-changelog
+check-changelog:
+	towncrier check
+
 # Various utilities
 .PHONY: push-helm-chart
 push-helm-chart:
--- a/ci/add-changelog-entry.sh
+++ b/ci/add-changelog-entry.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+# This is a simple interactive script for adding changelog entries.
+
+set -euo pipefail
+
+function get_default_change_type() {
+    local text
+    readonly text="${1}"
+    readonly commit_type="${text%[(:]*}"
+
+    local change_type
+    case "${commit_type}" in
+    fix )
+        change_type="fixed"
+        ;;
+    feat )
+        change_type="added"
+        ;;
+    * )
+        change_type="changed"
+    esac
+    echo "${change_type}"
+}
+
+function get_default_pr_number() {
+    local existing_pr_number
+    local commit_hash
+    commit_hash=$(git show -s --format=%h)
+    existing_pr_number=$(curl -s "https://api.github.com/repos/SumoLogic/sumologic-kubernetes-collection/commits/${commit_hash}/pulls?per_page=1" | jq '.[0].number' 2>/dev/null)
+    if [[ -n "${existing_pr_number}" ]]; then
+        echo "${existing_pr_number}"
+        return
+    fi
+
+    local latest_pr_number
+    latest_pr_number=$(curl -s "https://api.github.com/repos/SumoLogic/sumologic-kubernetes-collection/issues?per_page=1" | jq '.[0].number')
+    echo "$((latest_pr_number + 1))"
+}
+
+DEFAULT_TEXT="$(git show -s --format=%s || true)"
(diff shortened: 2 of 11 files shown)


towncrier.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[[tool.towncrier.type]]
+directory = "changed"
+name = "Changed"
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "fixed"
+name = "Fixed"
+showcontent = true
```
sumo-drosiek, line 28:
```suggestion
showcontent = true

```

Makefile
```diff
@@ -92,6 +92,35 @@ make test-integration:
...
+## We use Towncrier (https://towncrier.readthedocs.io) for changelog management
+
+ENTRY_TEXT ?= 'Use [Semantic Versioning](https://semver.org/spec/v2.0.0.html) for your changelog entry'
+CHANGE_ID ?= 9999
+CHANGE_TYPE ?= changed
+
+## Usage: make add-changelog-entry CHANGE_TYPE=changed
+.PHONY: add-changelog-entry
+add-changelog-entry:
+	towncrier create -c "$(ENTRY_TEXT)" "(CHANGE_ID).(CHANGE_TYPE).txt"
```
sumo-drosiek, line 105: Any documentation for that? Especially using with `CHANGELOG_ID`?

Makefile
```diff
@@ -92,6 +92,35 @@ make test-integration:
 	make -C ./tests/integration test
 
 
+# Changelog management
+## We use Towncrier (https://towncrier.readthedocs.io) for changelog management
+
+ENTRY_TEXT ?= 'Use [Semantic Versioning](https://semver.org/spec/v2.0.0.html) for your changelog entry'
+CHANGE_ID ?= 9999
```
sumo-drosiek, line 99: There shouldn't be default for `CHANGE_ID`

Makefile
```diff
@@ -92,6 +92,35 @@ make test-integration:
...
+CHANGE_TYPE ?= changed
```
sumo-drosiek, line 100: What are other options? Can we validate correctness of the type?

Makefile
swiatekm-sumo: Why not?

Makefile
swiatekm-sumo: The types are: `breaking`, `added`, `changed` and `fixed`. They're listed in `towncrier.toml`, and we do validate correctness.

Makefile
swiatekm-sumo: My intent here was for this to be simple. If you call this without any arguments, it should generate an example that is easy to manually edit. I can document the options, but using them is awkward imo.

Makefile
sumo-drosiek: because its obligatory field and different for every change

Makefile
sumo-drosiek: I hoped this will be complex solution, and now it seems does it resolves only conflicting issue, but instead you need to perfom multiple manual steps:

```
1. make 
2. edit filename
3. edit file content
```

I would appreciate it more if we run
```
make add-changelog-entry
```
and it will ask me for PR, type and entry text

swiatekm-sumo: Do you think that's an improvement? I honestly find it simpler to edit the file vs having to put all the arguments in a single command invocation. You can do `make add-changelog-entry CHANGE_ID=2564 CHANGE_TYPE=fixed ENTRY_TEXT="fix: something important", but I don't really think that's any better.

I could make the default text be the title of the most recent commit, but I don't want to make this excessively magical.

sumo-drosiek: I proposed rather interactive script

swiatekm-sumo: Allright, I can add a script for this. Maybe I can even be clever about default values?

sumo-drosiek: I would be great, but even dumb script would be nice feature for me :)

swiatekm-sumo: I think it's easy enough to put the title of HEAD as the default text. PR number and change type need to be put in by hand though (though I can probably infer something about change type from the title).

swiatekm-sumo: Ok, added it, can you check it out and let me know if this works for you?

kkujawa-sumo:
@swiatekm-sumo Could document how to prepare pull request for this repository with the changes that you introduced? Probably it will be good to have it documented in https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/CONTRIBUTING.md

swiatekm-sumo:
> @swiatekm-sumo Could document how to prepare pull request for this repository with the changes that you introduced? Probably it will be good to have it documented in https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/CONTRIBUTING.md

I think it'd also be good to add it to the PR template, now that I think about it.

ci/add-changelog-entry.sh
```diff
@@ -0,0 +1,43 @@
...
+    local CHANGE_TYPE 
+    case "${commit_type}" in
+    fix )
+        CHANGE_TYPE="fixed"
+        ;;
+    feat )
+        CHANGE_TYPE="added"
+        ;;
+    * )
+        CHANGE_TYPE="changed"
```
astencel-sumo, line 21: What about breaking changes? Shouldn't there be a separate category for it?

ci/add-changelog-entry.sh
```diff
@@ -0,0 +1,43 @@
...
+
+DEFAULT_TEXT="$(git show -s --format=%s || true)"
+DEFAULT_CHANGE_TYPE="$(get_default_change_type "${DEFAULT_TEXT}")"
+
+echo -e "Generating changelog entry"
+
+read -rp "Enter the ID of your Pull Request " PR_ID
+PR_NUMBER="${PR_ID###}"  # this removes a leading # if present
+
+echo -e "Enter the type of your change. Available types are: breaking, changed, added, fixed"
```
astencel-sumo, line 34: Oh this says there is a "breaking" type

ci/add-changelog-entry.sh
swiatekm-sumo: This just tries to infer the change type from commit title and sets that as the default. You can still set the change as breaking. Though I suppose I could check if the commit prefix has an exclamation mark and set based on that.

docs/dev.md
```diff
@@ -2,6 +2,50 @@
...
+following files in [.changelog]:
+
+```text
+123.fixed.0.txt
+123.fixed.1.txt
+123.fixed.2.txt
+```
+
+Unfortunately, the `towncrier create` command doesn't have support for this [yet](https://github.com/twisted/towncrier/issues/474), so you
+need to create the files manually.
```
sumo-drosiek, line 43: Should we allow to have multiple `README.md` entries for one PR?

swiatekm-sumo: Why not? It doesn't come up often, but for PRs that change the support matrix it's often more readable to have multiple change entries for a single PR. See Changelog for 3.1.0 for an example.

I guess we could also have a multiline entry in that case, but I leave that up to the PR author.

sumo-drosiek: I was rather thinking that it's not good to have multiple changes in one PR :D

astencel-sumo (approved):
Let's test it in production :tm: :muscle:

PR #2852, opened 2023-02-09
fix(logs): fix excluding logs by container, namespace, node, pod

### LINKED ISSUE #2851: Logs on 3.0.x no longer being ignored

On the 2.x branch we were able to ignore all logs by default by setting the following value in our helm values.yaml file:

`sumologic.logs.container.excludeNamespaceRegex: '.*'`

We would then have services, deployments, etc. explicitly opt into log collection by adding the following annotations:

```
sumologic.com/include: "true"
```

This no longer appears to work, even though the documentation says it should.

Also, we tried to set `otelcolInstrumentation.sourceMetadata.excludeNamespaceRegex` to `'.*'` as well, but this also had no impact.

**Logs** Logs which can be helpful in investigating the issue.

N/A

**Command used to install/upgrade Collection** Direct command which you used to install/upgrade collection.

ArgoCD upgraded the chart.

https://gist.github.com/poblahblahblah/952bf52aa4fe6ea4d6a422a122127f7d

Try to ignore logs, it should fail

_most_ logs on our cluster should have been ignored.

**Environment (please complete the following information):**

- **Collection version** (e.g. `helm ls -n sumologic`): 3.0.0
- **Kubernetes version** (e.g. `kubectl version`): 1.23
- **Cloud provider**: AWS/EKS
- **Others**: N/A

**Anything else do we need to know** Add any other context about the problem here, e.g. issue happens only occasionally.

We're currently flooding Sumologic with nginx access logs.

astencel-sumo:
Fixes #2851

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - chore: add support for GKE 1.25 [#2848]
 - chore: add support for Openshift 4.11 [#2848]
 - feat(otelcol): add max size in batch processors [#2839]
+- fix(logs): fix excluding logs by container, namespace, node, pod regex
 
 [#2831]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2831
 [#2832]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2832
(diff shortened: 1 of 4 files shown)


swiatekm-sumo (approved):

swiatekm-sumo:
Should we backport this to 3.0?

PR #2857, opened 2023-02-09
chore: bump sumologic-kubernetes-setup image to 3.6.0 and kube-state-metrics image to 2.7.0

aboguszewski-sumo:
The images had vulnerabilities found out by `Trivy`.


deploy/helm/sumologic/values.yaml
```diff
@@ -1875,6 +1875,8 @@ kube-prometheus-stack:
       # requests:
       #   cpu: 10m
       #   memory: 32Mi
+
```
swiatekm-sumo, line 1878: Can you add a comment explaining why we're changing this default?

swiatekm-sumo (approved):

PR #2863, opened 2023-02-10
fix(logs): fix metadata for pods with more than one container

### LINKED ISSUE #2862: Logs from two different containers of one pod show up in Sumo as coming from one of the containers
Version: `v3.0.1`
Kubernetes: v1.25.3
Container runtime: containerd

I have a pod that has two containers in it. I'm seeing the same `_sourceName` value and `container` value for logs from both containers, but I would expect them to be different.

### Steps to reproduce

- Create a k8s cluster with e.g. `minikube start --kubernetes-version=1.25.3 --container-runtime=containerd`
- Create a pod with two containers

    ```yaml
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: logger
    spec:
      selector:
        matchLabels:
          app: logger
      replicas: 1
      template:
        metadata:
          # annotations:
          #   sumologic.com/singleline.sourceCategory: single-line-sc
          labels:
            app: logger
        spec:
          containers:
          - name: singleline
            image: busybox
            args: [/bin/sh, -c, 'i=0; while true; do echo "singleline-$i"; i=$((i+1)); sleep 1; done']
            resources: {}

          - name: multiline
            image: busybox
            args: [/bin/sh, -c, 'i=0; while true; do echo "2001-02-03T04:05:06 multiline 1 log $i first line"; echo "  multiline 1 log $i second line"; echo "  multiline 1 log $i third line"; i=$((i+1)); sleep 1; done']
            resources: {}
    ```

- Install collection chart with `helm install -n sumo --create-namespace col sumologic/sumologic --version=3.0.1 --values user-values.yaml` with the following values file:

    ```yaml
    sumologic:
      accessId: xxx
      accessKey: yyy

      clusterName: two-containers

      events:
        enabled: false
      logs:
        enabled: true
      metrics:
        enabled: false
      traces:
        enabled: false
    ```

### Expected output

Logs from `multiline` container should be logged with `container=multiline`, logs from `singleline` container should be logged with `container=singleline`

### Actual output

Logs from both containers are logged with `container=singleline`

image

astencel-sumo:
Fixes #2862
The `containerName` metadata extraction gives incorrect results for pods with more than one containers.
We don't need this extraction anyway, because container logs already have the `k8s.container.name` attribute from log file name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- fix(logs): fix container attribute [#2863]
+
+[#2863]: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2863
+[Unreleased]: https://github.com/SumoLogic/sumologic-kubernetes-collection/compare/v3.1.0...main
+
 ## [v3.1.0]
 
 ### Released 2023-02-09
--- a/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
@@ -181,7 +181,6 @@ processors:
     extract:
       metadata:
         ## extract the following well-known metadata fields
-        - containerName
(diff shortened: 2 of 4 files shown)


sumo-drosiek (approved):

sumo-drosiek:
Could we add some tests for it?

astencel-sumo:
I'll do this separately. I don't want this fix to be slowed down by adding a new integration test for this - which isn't trivial.

PR #2864, opened 2023-02-10
chore(dev): recommend prettier vscode extension

swiatekm-sumo:


.vscode/extensions.json
```diff
@@ -1,5 +1,6 @@
 {
     "recommendations": [
         "davidanson.vscode-markdownlint",
```
astencel-sumo, line 3: Do we still need the `markdownlint` extension? Isn't it now superseded with Prettier?

astencel-sumo (approved):

.vscode/extensions.json
swiatekm-sumo: Not completely. We still keep it around because it does some checks prettier doesn't. I'm not sure it's worth it, but for now it stays.

PR #2875, opened 2023-02-15
feat(instrumentation): scrape label added and updated

mat-rumian:
PR adds scrape labels to service related to `otelcol-instrumentation`. Label value changed from `traces` -> `instrumentation`.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/deploy/helm/sumologic/templates/_helpers.tpl
+++ b/deploy/helm/sumologic/templates/_helpers.tpl
@@ -683,8 +683,8 @@ sumologic.com/app: otelcol-events
 sumologic.com/component: events
 {{- end -}}
 
-{{- define "sumologic.labels.traces.component" -}}
-sumologic.com/component: traces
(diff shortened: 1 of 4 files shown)


deploy/helm/sumologic/templates/_helpers.tpl
```diff
@@ -683,8 +683,8 @@ sumologic.com/app: otelcol-events
 sumologic.com/component: events
 {{- end -}}
 
-{{- define "sumologic.labels.traces.component" -}}
-sumologic.com/component: traces
```
swiatekm-sumo, line 687: I guess this wasn't used anywhere?

swiatekm-sumo: Nevermind, missed it below.

swiatekm-sumo (approved):
It would be nice to have a test for this, but I'm ok with it as-is.

PR #2876, opened 2023-02-15
ci: prune dev Helm Chart releases

swiatekm-sumo:
Prune dev Helm Chart releases older than 1 month. We don't really need the older ones, and our index was growing excessively large.

### Checklist

### DIFF
--- a/ci/_build_functions.sh
+++ b/ci/_build_functions.sh
@@ -67,6 +67,48 @@ function push_helm_chart() {
   set +ex
 }
 
+function prune_helm_releases() {
+  local chart_dir="$1"
+  local max_age_timestamp="$2"
+  local remote="origin"
+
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):
nit: I would wait some random time before retrying in while loop

swiatekm-sumo:
> nit: I would wait some random time before retrying in while loop
> 
> I don't feel comfortable with bare while loop sweat_smile

I just copied the current push chart loop xD. So I should probably add that delay in both of them - and in that case, maybe in a separate PR?

sumo-drosiek:
> > nit: I would wait some random time before retrying in while loop
> > I don't feel comfortable with bare while loop sweat_smile
> 
> I just copied the current push chart loop xD. So I should probably add that delay in both of them - and in that case, maybe in a separate PR?

Sure :)

PR #2879, opened 2023-02-16
chore: use nix for local environment

swiatekm-sumo:
Set up the local environment with Nix. 

### Checklist

- [X] Documentation updated


sumo-drosiek (commented):
Do we want to install vagrant by nix? :thinking:

swiatekm-sumo:
> Do we want to install vagrant by nix? thinking

I tried, but it doesn't really help. You need the vagrant version to be relatively tightly synced with your VM provider, and you can't install virtualbox using nix on Mac, for example. You'd also need to manage plugins manually anyway.

Same thing with docker. You can install the client using nix easily, but not the daemon. So may as well have users do it manually.

sumo-drosiek (approved):

CONTRIBUTING.md
```diff
@@ -1,16 +1,29 @@
 # Contributing Guide
 
-To contribute you will need to ensure you have the following setup:
+## Setting up a development environment
 
-- You have a working [Docker environment](https://docs.docker.com/engine).
-- You have a working [Ruby environment](https://ruby-doc.org).
+### Using nix
```
astencel-sumo, line 5: I don't think `Using nix` should be the first option listed in `Setting up a development environment`. It implies that it's the recommended and supported one, while I believe currently we still recommend and support the Vagrant option?

CONTRIBUTING.md
```diff
@@ -1,16 +1,29 @@
...
-```text
-git clone https://github.com/SumoLogic/sumologic-kubernetes-collection.git
-./ci/build.sh
-```
+   ```
+   sh <(curl -L https://nixos.org/nix/install) --daemon
+   ```
+
+1. Run `nix-shell` in the project root and wait for dependencies to be installed
+1. In order to run integration tests, you'll need the Docker daemon installed separately
```
astencel-sumo, line 14: Not sure what this means. Can you provide an example?

swiatekm-sumo: What is unclear about it? Integration tests require Docker, and Nix can't install the daemon part of it, so the user needs to do it themselves.

CONTRIBUTING.md
swiatekm-sumo: Sure, I can move it.

astencel-sumo (approved):

PR #2882, opened 2023-02-16
chore: bump setup image to 3.6.0

swiatekm-sumo:
This also includes a small change to field imports. Strictly speaking our current code was incorrect, but Terraform accepted it up until 1.3.0 where they made some changed that (probably accidentally) broke it.

We are now doing this import [correctly](https://developer.hashicorp.com/terraform/cli/commands/import#example-import-into-resource-configured-with-count).

Also removed an unnecessary changelog section from #2875 

### Checklist


tests/helm/terraform/static/all_fields.output.yaml
```diff
@@ -641,7 +641,7 @@ data:
 
             terraform import \
                 -var="create_fields=1" \
-                sumologic_field."${FIELD}" "${FIELD_ID}"
+                sumologic_field."${FIELD}"[0] "${FIELD_ID}"
```
rnishtala-sumo, line 644: why do we need the first index?

rnishtala-sumo (approved):

tests/helm/terraform/static/all_fields.output.yaml
swiatekm-sumo: Because this is a resource configured with count: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/ca05b7ff08182b781193e295955e2372d571856e/deploy/helm/sumologic/conf/setup/fields.tf#L3, and this syntax is how you're supposed to import them.

astencel-sumo: Nice find :clap:

PR #2892, opened 2023-02-22
feat(servicemonitor): fix instrumentation scraping, add tests

mat-rumian:
PR fixes metrics scraping from `instrumentation` related collectors and adds integration tests for Sumo defined ServiceMonitors and checks if desired services contain scraping labels.

### Checklist

- [ ] Documentation updated


tests/integration/helm_default_service_monitors_test.go
```diff
@@ -0,0 +1,157 @@
...
+		eventsAppLabel = "otelcol-events"
+	)
+
+	if err := promv1.AddToScheme(scheme.Scheme); err != nil {
+		require.Fail(t, "failed to register scheme: %v", err)
+	}
+
+	expectedDefaultServiceMonitors := internal.DefaultExpectedSumoServiceMonitors
+
+	featServiceMonitors := features.New("sumo service monitors number").
```
swiatekm-sumo, line 38: All of these servicemonitor tests should be template tests instead in my opinion. You don't need a cluster to check this, just the rendered manifests.

tests/integration/helm_default_service_monitors_test.go
```diff
@@ -0,0 +1,157 @@
...
+			stepfuncs.WaitUntilServiceMonitorPresent("collection-sumologic-otelcol-traces",
+				features.Labels{componentLabel: "instrumentation", scrapeLabel: "true"},
+				waitDuration, tickDuration)).
+		Assess("collection-sumologic-prometheus",
+			stepfuncs.WaitUntilServiceMonitorPresent("collection-sumologic-prometheus",
+				features.Labels{"operated-prometheus": "true"},
+				waitDuration, tickDuration)).
+		Feature()
+
+	featInstrumentationServiceMonitorsLabels := features.New("instrumentation services contain scrape labels").
```
swiatekm-sumo, line 85: Same thing as above, we can just check this in rendered templates.

tests/integration/helm_default_service_monitors_test.go
```diff
@@ -0,0 +1,157 @@
...
+		Assess("metadata-logs service scrape labels are present", func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+			name := fmt.Sprintf("%s-sumologic-metadata-logs", ctxopts.HelmRelease(ctx))
+			service := terrak8s.GetService(t, ctxopts.KubectlOptions(ctx), name)
+			require.Equal(t, "fluentd-logs", service.Labels[appLabel], "scrape label to equal true")
+			require.Equal(t, "true", service.Labels[scrapeLabel], "scrape label to be true")
+			return ctx
+		}).
+		Feature()
+
+	testenv.Test(t, featServiceMonitors, featServiceMonitorsSelectors, featInstrumentationServiceMonitorsLabels, featPrometheusOperatedServiceMonitorLabels,
```
swiatekm-sumo, line 155: What I'd do with this is move all of these to template tests, and check in the default metrics test whether the metrics we expect from collection monitoring actually exist. In integration tests, we should check if we get the right data at the end, without worrying about how that is accomplished.

PR #2894, opened 2023-02-23
chore(otoperator): update opentelemetry operator, add instrumentation customization

mat-rumian:
This PR:
* updates OpenTelemetry-Operator to latest `0.24.0`
* updates auto-instrumentation images used by OpenTelemetry-Operator
* add flags to disable/enable metrics and traces export from instrumentation
* increases default resources of OpenTelemetry-Operator pod to avoid OOM

### Checklist

- [ ] Documentation updated
- [x] OpenTelemetry-Operator Python instrumentation test
- [x] OpenTelemetry-Operator NodeJS instrumentation test
- [x] OpenTelemetry-Operator Java instrumentation test
- [x] OpenTelemetry-Operator Dotnet instrumentation test
- [x] OpenTelemetry-Operator Instrumentation Resource creation test


deploy/helm/sumologic/values.yaml
```diff
@@ -4517,6 +4539,13 @@ opentelemetry-operator:
 
   manager:
     env:
+    resources:
+      limits:
+        cpu: 250m
+        memory: 512Mi
+      requests:
+        cpu: 150m
+        memory: 256Mi
```
swiatekm-sumo, line 4548: Are these based on anything specific? I'd always like to keep requests as low as possible for services that don't really need to scale.

deploy/helm/sumologic/templates/opentelemetry-operator/job-instrumentation-cr.yaml
```diff
@@ -6,12 +7,15 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ template "sumologic.metadata.name.opentelemetry.operator.instrumentation.job" . }}
+  name: {{ template "sumologic.metadata.name.opentelemetry.operator.instrumentation.job" . }}-{{ include (print $.Template.BasePath "/opentelemetry-operator/configmap-instrumentation-cr.yaml") $ctx | sha256sum | substr 0 5 }}
```
swiatekm-sumo, line 10: Why is this necessary?

deploy/helm/sumologic/values.yaml
mat-rumian: Yes, we observed on few deployments with opentelemetry-operator enabled that manager pod is crashing becuase of too low resources.

deploy/helm/sumologic/templates/opentelemetry-operator/job-instrumentation-cr.yaml
mat-rumian: It's necessary for a case when customer want's to add additional namespace to be instrumented. For `helm upgrade...` existing job will be not triggered. There're two ways to solve this:
* delete job - I don't want to do that, in case of issues we will lose information about failure reason
* create new job with different name - it makes sure that job will be triggered again and operator will take care of creation of instrumentation resource

swiatekm-sumo: > For helm upgrade... existing job will be not triggered. 

Our Chart setup Job does get triggered every time. So why not this one?

sumo-drosiek: > > For helm upgrade... existing job will be not triggered.
> 
> Our Chart setup Job does get triggered every time. So why not this one?

Because we remove setup job when it finish

swiatekm-sumo: Can we do the same thing here?

deploy/helm/sumologic/Chart.yaml
```diff
@@ -37,6 +37,6 @@ dependencies:
     repository: https://sumologic.github.io/tailing-sidecar
     condition: tailing-sidecar-operator.enabled
   - name: opentelemetry-operator
-    version: 0.18.3
+    version: 0.24.0
```
kkujawa-sumo, line 40: please update Opentelemtery Operator version in support matrix here: https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/main/docs#support-matrix

deploy/helm/sumologic/templates/opentelemetry-operator/job-instrumentation-cr.yaml
mat-rumian: We are going to lose information in case of failed instrumentation resource creation. It will make harder for us to investigate/troubleshoot issues.

swiatekm-sumo: Can't we just use the default hook delete policy from Helm: https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies? Then it's only going to be deleted after you try installing again, but that sounds like it should work if you want logs from failed executions.

The reason I'm asking about this is that adding the hash to the name feels like an antipattern in K8s and I'd rather avoid it.

sumo-drosiek: Hmm. Will that work for non-helm installation?

swiatekm-sumo: Does this work for setup in the non-Helm installation? If so, how?

sumo-drosiek: > Does this work for setup in the non-Helm installation

Good question. I don't know the answer :fearful:

swiatekm-sumo (approved):

PR #2935, opened 2023-03-15
docs: collecting additional logs from the node using OTC

dmolenda-sumo:


docs/best-practices.md
```diff
@@ -290,7 +290,59 @@ sumologic:
...
+        receivers:
+          otlp/extrafiles:
+            protocols:
+              http:
+                endpoint: 0.0.0.0:4319
+        service:
+          pipelines:
+            logs/extrafiles:
+              receivers: [otlp/extrafiles]
+              exporters: [sumologic/containers]
```
rnishtala-sumo, line 340: should this be `sumologic/node` or `sumologic/systemd` instead?

Also should we be recommending using a few processors like these?
```
Batch Processor
Memory Limit Processor
Attributes Processor
Resource Processor
```

docs/best-practices.md
```diff
@@ -290,7 +290,59 @@ sumologic:
...
+
+The following configuration can be used for these purposes:
+
+```yaml
+otellogs:
+  config:
+    merge:
+      receivers:
+        filelog/extrafiles:
+          include: [/var/log/extrafiles/extrafile.log]
```
ballensans, line 307: Can you also provide an example for receiving data from the SystemD Journal?  In particular, this approach looks very close to supporting collection of IPTables "LOG" directives.

astencel-sumo (approved):
This is great :+1:

rnishtala-sumo (approved):

docs/best-practices.md
swiatekm-sumo: systemd logs are already collected by default, you don't need any additional configuration for those.

docs/best-practices.md
swiatekm-sumo: I think we should have the batch processor and memory limiter here, but no need to do anything else by default.

PR #2951, opened 2023-03-22
fix: values.yaml for vagrant

dmolenda-sumo:
Fix values.yaml file for vagrant to make it works with Helm Chart 3.x


sumo-drosiek (approved):
LGTM

PR #2958, opened 2023-03-27
fix(otelllogs): fix disabling otellogs metrics collection

### LINKED ISSUE #2947: otel collector service fails if collector metrics are disabled

otel collector service fails if collector is enabled but collector metrics are disabled

**Logs** Logs which can be helpful in investigating the issue.

```
Error: cannot patch "sumologic-sumologic-otelcol-logs-collector" with kind Service: Service "sumologic-sumologic-otelcol-logs-collector" is invalid: spec.ports: Required value
```

**Command used to install/upgrade Collection** Direct command which you used to install/upgrade collection.

```
otellogs:
  metrics:
    enabled: false
```

deploy helm chart with enabled otel but disable otel  metrics

it should deploy with no errors.

**Environment (please complete the following information):**

- **Collection version** (e.g. `helm ls -n sumologic`): 3.3.0
- **Kubernetes version** (e.g. `kubectl version`): 1.24
- **Cloud provider**: aws
- **Others**:

**Anything else do we need to know** Add any other context about the problem here, e.g. issue happens only occasionally.

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/070d6109042ae9e7e3a52fceadc979b88b9af845/deploy/helm/sumologic/templates/logs/collector/otelcol/service.yaml#L1-L17

if .Values.otellogs.metrics.enabled  == false , ports becomes empty.

kasia-kujawa:
fixes #2947

### Checklist

- [ ] Documentation updated

### DIFF
--- a/tests/helm/otellogs_metrics_test.go
+++ b/tests/helm/otellogs_metrics_test.go
@@ -1,0 +1,20 @@
+package helm
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOtellogsMetricsDisabled(t *testing.T) {
+	t.Parallel()
+	templatePath := "templates/logs/collector/otelcol/service.yaml"
+	valuesYaml := `
+otellogs:
+  metrics:
+    enabled: false
+`
+	_, err := RenderTemplateFromValuesStringE(t, valuesYaml, templatePath)
+	expectedError := "Error: could not find template templates/logs/collector/otelcol/service.yaml in chart"
+	assert.ErrorContains(t, err, expectedError)
+}
--- a/.changelog/2958.fixed.txt
+++ b/.changelog/2958.fixed.txt
@@ -1,0 +1,1 @@
+fix(otelllogs): fix disabling otellogs metrics collection
--- a/deploy/helm/sumologic/templates/logs/collector/otelcol/service.yaml
+++ b/deploy/helm/sumologic/templates/logs/collector/otelcol/service.yaml
@@ -1,4 +1,4 @@
-{{- if eq (include "logs.collector.otelcol.enabled" .) "true" }}
+{{- if and (eq (include "logs.collector.otelcol.enabled" .) "true") (.Values.otellogs.metrics.enabled) }}
 apiVersion: v1
 kind: Service
 metadata:
@@ -14,10 +14,8 @@ spec:
   selector:
     app.kubernetes.io/name: {{ template "sumologic.labels.app.logs.collector.pod" . }}
   ports:
-  {{- if eq .Values.otellogs.metrics.enabled true }}
   - name: metrics
     port: 8888
     targetPort: 8888
     protocol: TCP
-  {{- end }}
 {{- end }}
--- a/tests/helm/testdata/goldenfile/otellogs_metrics/enabled.input.yaml
+++ b/tests/helm/testdata/goldenfile/otellogs_metrics/enabled.input.yaml
@@ -1,0 +1,3 @@
+otellogs:
+  metrics:
+    enabled: true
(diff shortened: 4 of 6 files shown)


swiatekm-sumo (commented):
Can we add a test for this, as it already caused a bug?

sumo-drosiek (approved):

kkujawa-sumo:
> Can we add a test for this, as it already caused a bug?

I added simple test

sumo-drosiek:
Can we check for enabled metrics case as well?

tests/integration/features.go
```diff
@@ -824,3 +824,37 @@ func CheckFluentdEventsInstall(builder *features.FeatureBuilder) *features.Featu
...
+				require.Eventually(t, func() bool {
+					services = terrak8s.ListServices(t, ctxopts.KubectlOptions(ctx), v1.ListOptions{
+						LabelSelector: fmt.Sprintf("app=%s-sumologic-otelcol-logs-collector", releaseName),
+					})
+
+					return len(services) == 1
+				}, waitDuration, tickDuration)
+				return ctx
+			})
+}
```
sumo-drosiek, line 860: Please make it gennric

```
func checkOtellogsMetricsServiceCount(ctx context.Context, t *testing.T, envConf *envconf.Config, count int) context.Context {
				releaseName := ctxopts.HelmRelease(ctx)
				var services []corev1.Service
				require.Eventually(t, func() bool {
					services = terrak8s.ListServices(t, ctxopts.KubectlOptions(ctx), v1.ListOptions{
						LabelSelector: fmt.Sprintf("app=%s-sumologic-otelcol-logs-collector", releaseName),
					})

					return len(services) == count
				}, waitDuration, tickDuration)
				return ctx
			}
}
```

```suggestion
			checkOtellogsMetricsServiceCount(ctx, t, envConf, 1)
}
```

kkujawa-sumo: I changed tests to template tests

sumo-drosiek: Now there is no changes for features.go :fearful:

kkujawa-sumo: To test if service is available or not and if it has proper form we can use template tests instead of integration tests.
Do you think that integration tests are needed also?

sumo-drosiek: Ou, I misunderstood the context of `template`

I think we can use `template` tests, but in current tests form empty files don't test anything

Could you try if the following snippets are correct?
```
---
# Source: sumologic/templates/logs/collector/otelcol/service.yaml
```

kkujawa-sumo:
@sumo-drosiek @swiatekm-sumo please take a look once again

tests/helm/otellogs_metrics_test.go
```diff
@@ -0,0 +1,21 @@
...
+	templatePath := "templates/logs/collector/otelcol/service.yaml"
+	valuesYaml := `
+otellogs:
+  metrics:
+    enabled: false
+`
+	_, err := RenderTemplateFromValuesStringE(t, valuesYaml, templatePath)
+	expectedError := "error while running command: exit status 1; WARNING: Kubernetes configuration file is group-readable. This is insecure. "
+	expectedError += "Location: /var/snap/microk8s/current/credentials/kubelet.config\nError: could not find template templates/logs/collector/otelcol/service.yaml in chart"
+	assert.EqualError(t, err,  expectedError)
```
swiatekm-sumo, line 20: If you don't want to do this, you can just render all the templates, decode them into 
k8s objects, and then assert that the Service isn't among them.

kkujawa-sumo: Isn't this version simpler?
```
	_, err := RenderTemplateFromValuesStringE(t, valuesYaml, templatePath)
	expectedError := "Error: could not find template templates/logs/collector/otelcol/service.yaml in chart"
	assert.ErrorContains(t, err, expectedError)
```

swiatekm-sumo: It is, I didn't like the previous error assertion, but this one works.

swiatekm-sumo (approved):

PR #2967, opened 2023-03-31
docs(troubleshooting): Adding docs for high prometheus memory usage and queue size

rnishtala-sumo:
### Checklist

- [ ] Documentation updated


docs/troubleshoot-collection.md
```diff
@@ -49,6 +52,12 @@ To set your namespace context more permanently, you can run
 kubectl config set-context $(kubectl config current-context) --namespace=sumologic
 ```
 
+## Collection dashboards
+
+Please consult this when diagnosing issues before diving into Kubernetes directly.
+
+[Zaidan Collection](https://stagdata.long.sumologic.net/ui/#/library/folder/20836981)
```
sumo-drosiek, line 59: This is not accessible

docs/troubleshoot-collection.md
```diff
@@ -239,6 +248,63 @@ You [check Prometheus logs](#prometheus-logs) to verify there are no errors duri
...
+prometheus-stag-collection-prometheus-8-control-pl-0   prometheus,config-reloader   20Gi,50Mi
+prometheus-stag-collection-prometheus-9-sumo-0         prometheus,config-reloader   20Gi,50Mi
+prometheus-stag-collection-prometheus-assembly-0       prometheus,config-reloader   20Gi,50Mi
+```
+
+Note that pods can contain more than one container so in this example prometheus pods have 20Gi memory limit and config-reloader containers
+have 50Mi memory limit.
+
+When memory usage is higher than ~`95%` of Prometheus container's memory limit, (in the above case it's not: none of the containers exceeded
+`95% * 20 = 19Gi` of used memory) remove WAL from within the container:
```
sumo-drosiek, line 293: I don't think this is proper solution. I would rather say that customer should increase memory requests/limits for Prometheus or reduce number of metrics being scraped by the Prometheus. Removing WAL is more like temporary solution

docs/troubleshoot-collection.md
```diff
@@ -405,6 +471,32 @@ The duplicated pod deletion command is there to make sure the pod is not stuck i
 If you observe that Prometheus Pod needs more and more resources (out of memory failures - OOM killed Prometheus) and you are not able to
 increase them then you may need to horizontally scale Prometheus. :construction: Add link to Prometheus sharding doc here.
 
+### Otelcol enqueue failures
+
+Enqueue failures happen when otelcol can't write to its persistent queue. They signify data loss, as otelcol is unable to locally buffer the
+data, and is forced to drop it.
```
sumo-drosiek, line 477: Could you add snippet with example error message here?

docs/troubleshoot-collection.md
```diff
@@ -405,6 +471,32 @@ The duplicated pod deletion command is there to make sure the pod is not stuck i
...
+
+#### Queue full
+
+Most of the time, this happens because the queue is full, which is almost always caused by otelcol not being able to send data to Sumo, and
+should be accompanied by other alerts. Resolving the underlying issue will resolve the alert - in this case, this alert is informational.
+
+The queue size can be checked on the Otelcol dashboard, available in [the collection dashboards folder](#collection-dashboards).
+
```
sumo-drosiek, line 485: Shouldn't we have separate section or document for alerts? We do not have such right now, so it can be not clear for customers what they are and how do the work

rnishtala-sumo: Maybe we leave alerts out of this document and instead here I could add here the query and a screenshot for a full queue? (for the customer's reference)

docs/troubleshoot-collection.md
rnishtala-sumo: right! I was thinking about support when I added this, but we should probably only add links here that the customer can access.

sumo-drosiek: Yes, it it's not an internal documentation

docs/troubleshoot-collection.md
sumo-drosiek: yes, especially because it is temporary solution or clean up after/during metrics spike

docs/troubleshoot-collection.md
sumo-drosiek: Yeah, it sounds reasonable

docs/README.md
```diff
@@ -86,7 +86,7 @@ The following table displays the tested Kubernetes and Helm versions.
 
 | Name          | Version                                  |
 | ------------- | ---------------------------------------- |
-| K8s with EKS  | 1.21<br/>1.22<br/>1.23<br/>1.24          |
+| K8s with EKS  | 1.21<br/>1.22<br/>1.23<br/>1.24<br/>1.25 |
```
perk-sumo, line 89: @rnishtala-sumo can we move that one out to a separate commit and/or PR? 🙏

sumo-drosiek:
@rnishtala-sumo Coul you recreate this PR for [SumoLogic docs repo](https://github.com/SumoLogic/sumologic-documentation/)?

Astitva-sumo:
Closing stale PR

PR #2973, opened 2023-04-06
Drosiek pvc cleaner

sumo-drosiek:
### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/2973.changed.txt
+++ b/.changelog/2973.changed.txt
@@ -1,0 +1,1 @@
+feat: stop using sumologic kubernetes tools for pvc cleaner
--- a/deploy/helm/sumologic/conf/pvc-cleaner/pvc-cleaner.sh
+++ b/deploy/helm/sumologic/conf/pvc-cleaner/pvc-cleaner.sh
@@ -1,0 +1,169 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+readonly NAMESPACE="${1:-}"
+readonly PVC_SELECTOR="${2:-}"
+readonly HPA_NAME="${3:-}"
+readonly ARGS_NUMBER="$#"
+readonly MIN_ARGS_NUMBER=2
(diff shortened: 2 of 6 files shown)


deploy/helm/sumologic/README.md
```diff
@@ -573,8 +573,8 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
+| `pvcCleaner.job.image.tag`                                                       | Image tag for pvcCleaner docker containers.                                                                                                                                                                                                                                                                                                                          | `1.26.3`                                                                                                                                                                             |
```
aboguszewski-sumo, line 577: Technically, that sounds like a breaking change (although I doubt that many customers are using it now), do we have any procedures for that?

sumo-drosiek: It is breaking change, as we change `mount` property of the pod. I don't think we have any procedure for that. We can add warning message for helm chart before `3.5.0` to ask customers to remove the current jobs

aboguszewski-sumo: Sounds good, but I guess that it still should be discussed once @perk-sumo is back.

aboguszewski-sumo: If someone changed the image tag, for example. 
I doubt someone is using this with overridden image tag or image name, but still.

sumo-drosiek: I double checked and pod spec in cronjob doesn't seem to be immutable (at least volume configuration is not)

swiatekm-sumo: > If someone changed the image tag, for example.
> I doubt someone is using this with overridden image tag or image name, but still.

If you use a different tools image version, this will still work - the only requirements are bash and kubectl. If someone put something completely custom here, I don't think we support that.

swiatekm-sumo (approved):

PR #2982, opened 2023-04-13
Otel statefulset cloudwatch collector

rnishtala-sumo:
This PR creates a statefulset for collecting logs from cloudwatch. It also documents some of the steps needed to setup log ingestion from cloudwatch. The fargate readme has been updated with some details

- It has initial documentation on how to set this up
   - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/otel-statefulset-collector/docs/fargate.md#logs

- The changes have been tested on a fargate cluster, with logs being ingested into sumo
   - example-1: https://stagdata.long.sumologic.net/ui/#/search/create?id=BKYd1rzPGnOcpDlsB3D1twcoSk6xBYIxl36XrUF8
   - example-2: https://stagdata.long.sumologic.net/ui/#/search/create?id=MIOPkMUGxmQOVyI4tZntY8q6Bgzq8s73HR8wNdKc

- there isn't any processing or metadata enrichment yet

- There's a basic template test, integration tests are pending

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/otel-statefulset-collector/docs/fargate.md#logs

### Checklist

- [x] Documentation updated


deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -582,7 +606,17 @@ service:
...
+## Use this pipeline for cloudwatch logs
+{{- if .Values.sumologic.logs.collector.otelcloudwatch.enabled }}
+    logs/otlp/otelcloudwatch:
+      receivers:
+        - otlp
+      processors:
+        - memory_limiter
+        - batch
+      exporters:
+        - sumologic/systemd
```
aboguszewski-sumo, line 618: I think we will want to send it to `containers` endpoint (and to metadata statefulset - as I understand, you haven't debugged that one yet)

docs/fargate.md
```diff
@@ -115,6 +120,93 @@ fi
...
+
+To configure the service account to use an IAM role use the command below
+
+```bash
+eksctl create iamserviceaccount --name release-sumologic-otelcol-logs-collector \
+                                --namespace default --cluster sumo-fargate \
+                                --role-name sumo-role \
+                                --attach-policy-arn arn:aws:iam::586421305253:policy/sumo-policy \
+                                --override-existing-serviceaccounts --approve
+```
```
aboguszewski-sumo, line 197: Where do we provide this `sumo-policy` and what is it going to be? IMO it should be just a policy that enables reading from cloudwatch

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
rnishtala-sumo: That part works now, the logs are being sent to sumo logic via the metadata statefulset.
https://stagdata.long.sumologic.net/ui/#/search/create?id=EmilVfSu90JlAmXg2EkjawD3nsDpWabmADKiLRWF

Yes I can change it to use the containers endpoint

docs/fargate.md
rnishtala-sumo: Right the sumo-policy would get associated with the IAM service account. I linked the doc for creating the policy and the corresponding role in AWS in the doc
https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
aboguszewski-sumo: Nevermind then, I did not notice that this is the configmap for metadata otelcol.

docs/fargate.md
rnishtala-sumo: ok I can add the role permissions that are needed for the service account

deploy/helm/sumologic/templates/logs/collector/otelcol-cloudwatch/statefulset.yaml
```diff
@@ -0,0 +1,115 @@
...
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: config.yaml
+            path: config.yaml
+          name: {{ template "sumologic.metadata.name.logs.collector.configmap" . }}
+        name: otelcol-config
+      - name: file-storage
+        persistentVolumeClaim:
+          claimName: file-storage-sumo-sumologic-otelcol-logs-0
```
sumo-drosiek, line 63: It should be optional

deploy/helm/sumologic/templates/logs/collector/otelcol-cloudwatch/statefulset.yaml
```diff
@@ -0,0 +1,115 @@
...
+        - name: LOGS_METADATA_SVC
+          valueFrom:
+            configMapKeyRef:
+              name: sumologic-configmap
+              key: metadataLogs
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+{{- end }}
```
sumo-drosiek, line 115:
```suggestion
{{- end }}

```

deploy/helm/sumologic/templates/logs/collector/otelcol/serviceaccount.yaml
```diff
@@ -6,4 +6,10 @@ metadata:
   labels:
     app: {{ template "sumologic.labels.app.logs.collector.serviceaccount" . }}
     {{- include "sumologic.labels.common" . | nindent 4 }}
+  annotations:
+{{- if .Values.sumologic.logs.collector.otelcloudwatch.enabled }}
+{{- if .Values.sumologic.logs.collector.otelcloudwatch.roleArn }}
+    eks.amazonaws.com/role-arn: {{ .Values.sumologic.logs.collector.otelcloudwatch.roleArn }}
+{{- end }}
+{{- end }}
```
sumo-drosiek, line 14: we can use `and` instead of two conditions:

```suggestion
  annotations:
{{- if and .Values.sumologic.logs.collector.otelcloudwatch.enabled .Values.sumologic.logs.collector.otelcloudwatch.roleArn }}
    eks.amazonaws.com/role-arn: {{ .Values.sumologic.logs.collector.otelcloudwatch.roleArn }}
{{- end }}
```

docs/fargate.md
```diff
@@ -115,6 +120,135 @@ fi
...
+                limit: 10
+```
+
+where `my-role` is the name of the role created while setting up [authentication](#authenticate-with-cloudwatch)
+
+For more information on creating the appropriate roles and policies, please refer to
+[service account IAM role](https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html)
+
+Finally, install/upgrade the [collection](#install-or-upgrade-collection)
+
```
aboguszewski-sumo, line 251: I think it's worth mentioning that `fullnameOverride` should be set to `my-service-account`, because this name is essentially the name of the service account used.

docs/fargate.md
```diff
@@ -115,6 +120,135 @@ fi
...
+After these variables have been set and the above trush relationship is defined, the next step is to create an IAM role and attach it to a
+policy.
+
+```bash
+aws iam create-role --role-name my-role --assume-role-policy-document file://trust-relationship.json --description "my-role-description"
+aws iam attach-role-policy --role-name my-role --policy-arn=arn:aws:iam::$account_id:policy/my-policy
+```
+
+#### Enable cloudwatch collection (EKS fargate only)
+
```
aboguszewski-sumo, line 224: Why is it described as `EKS Fargate only`? This doc refers to Fargate anyway.

deploy/helm/sumologic/values.yaml
```diff
@@ -242,7 +242,16 @@ sumologic:
...
+      otelcloudwatch:
+        enabled: false
+        roleArn: ""
+        awscloudwatch:
+          region: us-east-2
+          logs:
+            poll_interval: 1m
+            groups:
+              autodiscover:
+                limit: 10
```
sumo-drosiek, line 254: Do we need to expose full configuration to customers?
I would rather keep:

```
sumologic:
  logs:
    cloudwatch:
      collector:
        enabled: true
      roleArn: ""
      region: ""
      streams:
        autodiscover: true/false
        names: []
        prefixes: []
      ...
```

and `merge` and `override` in addition to that

Also my question is are we able to autodetect region, roleArn, names and/or prefixes of the cluster we are installed in?

docs/fargate.md
```diff
@@ -115,6 +120,135 @@ fi
...
+apiVersion: v1
+metadata:
+  name: aws-observability
+  labels:
+    aws-observability: enabled
+```
+
+The fargate log router manages the `Service` and `Input` sections. One needs to create a `ConfigMap` to enable log routing to cloudwatch, an
+example is shown below:
+
```
aboguszewski-sumo, line 144: We should mention why we're sending the logs to cloudwatch.

docs/fargate.md
```diff
@@ -115,6 +120,135 @@ fi
...
+EOF
+```
+
+After these variables have been set and the above trush relationship is defined, the next step is to create an IAM role and attach it to a
+policy.
+
+```bash
+aws iam create-role --role-name my-role --assume-role-policy-document file://trust-relationship.json --description "my-role-description"
+aws iam attach-role-policy --role-name my-role --policy-arn=arn:aws:iam::$account_id:policy/my-policy
+```
```
aboguszewski-sumo, line 221: It should be mentioned that the policy should at least allow to read cloudwatch.

docs/fargate.md
```diff
@@ -115,6 +120,135 @@ fi
...
+
+- Set your AWS account ID to an environment variable
+- Set your cluster's OIDC identity provider to an environment variable. Replace `my-cluster` with the name of your cluster
+- Set the namespace and the service account name
+
+```bash
+account_id=$(aws sts get-caller-identity --query "Account" --output text)
+oidc_provider=$(aws eks describe-cluster --name my-cluster --region $AWS_REGION --query "cluster.identity.oidc.issuer" --output text | sed -e "s/^https:\/\///")
+export namespace=default
+export service_account=my-service-account
```
sumo-drosiek, line 187: Please use upper case variable names, also is this `namespace` the same like `NAMESPACE` mentioned on the top of this document? We should be consistent regarding the variables within the document

docs/fargate.md
```diff
@@ -115,6 +120,135 @@ fi
...
+        enabled: true
+        roleArn: arn:aws:iam::account-id:role/my-role
+        # cloudwatch receiver configuration: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/awscloudwatchreceiver
+        awscloudwatch:
+          region: us-east-2
+          logs:
+            poll_interval: 1m
+            groups:
+              autodiscover:
+                limit: 10
```
sumo-drosiek, line 242: We should hide the configuration if possible. I don't feel this one is user-friendly and our customers usually are not interested in implementation details (which I consider exposing direct configuration)

deploy/helm/sumologic/values.yaml
aboguszewski-sumo: Group name is determined by fluent-bit config.
Stream names seem to have format `<FLUENT_BIT_PREFIX>-kube.var.log.containers.<POD_NAME>_<NAMESPACE_NAME>_<CONTAINER_NAME>-<LOGFILE_NAME>`
I'm not sure about the rest.

If we are to change the receiver config as you suggest, then we should also mention in the docs that `merge` and `override` can be used.

sumo-drosiek: Shouldn't we actually generate `fluent-bit` config with our helm chart for the most basic cases?

sumo-drosiek: > If we are to change the receiver config as you suggest, then we should also mention in the docs that merge and override can be used.

We will have at some point, but we just get out from exposing full configuration for other components so we shouldn't introduce the same approach again and again :sweat_smile:

aboguszewski-sumo: > Shouldn't we actually generate fluent-bit config with our helm chart for the most basic cases?

I guess we should, you're right.

deploy/helm/sumologic/templates/logs/collector/otelcol-cloudwatch/statefulset.yaml
aboguszewski-sumo: I agree.
Additional comment: currently you don't mention in the docs that these volumes have to be created in the same way as for metrics.

docs/fargate.md
aboguszewski-sumo: Update: the service account name will be rather `<FULLNAME>-otelcol-logs-collector`

deploy/helm/sumologic/values.yaml
rnishtala-sumo: > Do we need to expose full configuration to customers?

I thought it would be useful to do this since there are multiple ways to configure the receiver and the benefit of exposing it this way could be that we wouldn't need to release new versions for different cloudwatch configuration use cases that multiple customers might have, as shown below:

1st use case (using autodiscover)
```
awscloudwatch:
  region: us-west-1
  logs:
    poll_interval: 1m
    groups:
      autodiscover:
        limit: 100
        prefix: /aws/eks/
        streams:
          prefixes: [kube-api-controller]
```

2nd use case: (using named)

```
awscloudwatch:
  region: us-west-1
  logs:
    poll_interval: 5m
    groups:
      named:
        /aws/eks/dev-0/cluster: 
          names: [kube-apiserver-ea9c831555adca1815ae04b87661klasdj]
```

docs/fargate.md
rnishtala-sumo: We could point the customer to the cloudwatch receiver configuration. That way we could just tell the customer that we support cloudwatch receiver configuration as is. Allowing this flexibility might work to our benefit as different customers may have different configuration requirements for cloudwatch receivers.

aboguszewski-sumo:
Some additional thoughts after using this:

- we should probably decide which log groups we're going to observe: with our suggested `fluent-bit` config there's going to be only one group, can it be more groups with different configs?
- current way of providing `awscloudwatch receiver` config is a bit tiresome, for example I tried to use the following config to read log from a specific group:
```yaml
sumologic:
  logs:
    collector:
      otelcloudwatch:
        awscloudwatch:
          logs:
            groups:
              named:
                fluent-bit-cloudwatch:
                  names: ["some-name"]
```
  and it raised an error, because the default is also using `autodiscover`, and both `autodiscover` and `named` can't be used at the same time. So we should probably either have this receiver's config empty as default or don't copy it directly as it was suggested in one of the comments.

- persistency for the collector definitely should be configurable (disabling it manually is a mess) and it should be explained in more detail how to enable it on Fargate
- in general, we should have some common section about how to enable persistency on Fargate that will be common for the collector, metrics and logs

sumo-drosiek:
Also looking again at [Prerequisites](https://docs.aws.amazon.com/eks/latest/userguide/fargate-logging.html) I have to say I don't think I was right, and actually do not feel like generating fluent-bit configuration helps with anything as customers have to perform multiple manual steps anyway:

> **Prerequisites**
> - An existing Fargate profile that specifies an existing Kubernetes namespace that you deploy Fargate pods to. For more information, see [Create a Fargate profile for your cluster](https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-create-profile).
> - An existing Fargate pod execution role. For more information, see [Create a Fargate pod execution role](https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-sg-pod-execution-role).

docs/fargate.md
rnishtala-sumo: yes its the same namespace so, I'll remove this reference

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
sumo-drosiek, line 1: We should create different file with configuration as it doesn't make sense to have other pipelines (`containers`, `syslog`) in fargate

deploy/helm/sumologic/templates/logs/collector/otelcol-cloudwatch/statefulset.yaml
```diff
@@ -0,0 +1,119 @@
...
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ template "sumologic.labels.app.logs.collector.statefulset" . }}
+  serviceName: {{ template "sumologic.labels.app.logs.cloudwatch.service-headless" . }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/logs/collector/otelcol/configmap.yaml") . | sha256sum }}
```
sumo-drosiek, line 17: Please do not reuse collector configuration for fargate usecase

deploy/helm/sumologic/values.yaml
```diff
@@ -242,6 +242,16 @@ sumologic:
...
+      otelcloudwatch:
+        enabled: false
+        roleArn: ""
+        ## Configure persistence for the cloudwatch collector
+        persistence:
+          enabled: true
+        region: us-east-2
+        pollInterval: 1m
+        # A map of log group and stream prefixes
+        logGroups: {}
```
sumo-drosiek, line 254: We should provide some examples. We can also provide defaults which will be working according to our documentation

deploy/helm/sumologic/values.yaml
```diff
@@ -242,6 +242,16 @@ sumologic:
...
         enabled: true
+      otelcloudwatch:
+        enabled: false
+        roleArn: ""
+        ## Configure persistence for the cloudwatch collector
+        persistence:
+          enabled: true
+        region: us-east-2
+        pollInterval: 1m
+        # A map of log group and stream prefixes
```
sumo-drosiek, line 253:
```suggestion
        ## A map of log group and stream prefixes
```

docs/fargate.md
```diff
@@ -115,6 +120,143 @@ fi
...
+        log_group_name my-logs
+        log_stream_prefix from-fluent-bit-
+        log_retention_days 60
+        auto_create_group true
+```
+
+You can stream logs from Fargate directly to Amazon CloudWatch, using the output plugin shown above. Cloudwatch is currently the only
+supported output plugin in fluent-bit running on EKS fargate.
+
+For more information on this refer to [fargate-logging](https://docs.aws.amazon.com/eks/latest/userguide/fargate-logging.html)
```
sumo-drosiek, line 170: We should document or note that the above steps are not enough due to `Prerequisites`

image

We should add some troubleshooting for customers which do not followed the `prerequisites`. I would be also nice to have some verification steps

sumo-drosiek: > We should document or note that the above steps are not enough due to Prerequisites

I see we do it further in this doucument.

tests/helm/testdata/goldenfile/logs_otc_serviceaccount/basic.output.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: RELEASE-NAME-sumologic-otelcol-logs-collector
+  annotations: !!null
```
sumo-drosiek, line 6: it doesn't look correct

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
swiatekm-sumo: Yeah, and I think it makes sense to have a separate config file for this. These don't really share anything substantial.

tests/helm/testdata/goldenfile/logs_otc_serviceaccount/basic.output.yaml
rnishtala-sumo: the helm template tests do pass

deploy/helm/sumologic/values.yaml
rnishtala-sumo:
```
logGroups:
  fluent-bit:
    names: [fluent-bit]
```
this would be an example. I can add it as a comment instead of a default, as the customer will most likely not have the default log group in cloudwatch which will result in error logs in our collector like "Log group not found"

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
rnishtala-sumo: ok I'll work on building a separate configuration file.

docs/fargate.md
rnishtala-sumo: I'll add a prerequisites section in the log router configuration.

rnishtala-sumo: Added additional troubleshooting steps in the last commit

deploy/helm/sumologic/templates/logs/collector/otelcol-cloudwatch/statefulset.yaml
rnishtala-sumo: Added a separate configuration file for the fargate usecase

deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml
rnishtala-sumo: Added a separate config file for this: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/otel-statefulset-collector/deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml

tests/helm/testdata/goldenfile/logs_otc_serviceaccount/basic.output.yaml
sumo-drosiek: Ok, I will elaborate. I don't like how does it look and IMHO we should:

- have annotations, or
- do not `annotations` key at all, or
- have empty dictionary `{}`

Also this change is unrelated to the PR :thinking:

deploy/helm/sumologic/values.yaml
```diff
@@ -242,6 +242,20 @@ sumologic:
     collector:
       otelcol:
         enabled: true
+      otelcloudwatch:
```
sumo-drosiek, line 245: I think we should mark it as experimental for two reasons:
- we are still evaluating the final shape
- we may need to introduce breaking changes in near future

deploy/helm/sumologic/templates/logs/collector/otelcol-cloudwatch/configmap.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ template "sumologic.metadata.name.logs.collector.cloudwatch" . }}
+  labels:
+    app: {{ template "sumologic.labels.app.logs.cloudwatch.configmap" . }}
+    {{- include "sumologic.labels.common" . | nindent 4 }}
+data:
+  config.yaml: |
+{{- $finalConfig | toYaml | nindent 4 }}
+{{- end }}
```
sumo-drosiek, line 21:
```suggestion
{{- end }}

```

docs/fargate.md
```diff
@@ -115,6 +124,153 @@ fi
...
+          "$oidc_provider:sub": "system:serviceaccount:$namespace:$service_account"
+        }
+      }
+    }
+  ]
+}
+EOF
+```
+
+After these variables have been set and the above trush relationship is defined, the next step is to create an IAM role and attach it to a
```
sumo-drosiek, line 235:
```suggestion
After these variables have been set and the above trust relationship is defined, the next step is to create an IAM role and attach it to a
```

sumo-drosiek (commented):
Overally LGTM, some minor minor comments

docs/fargate.md
```diff
@@ -115,6 +124,153 @@ fi
...
+        Match   kube.*
+        region region-code
+        log_group_name my-logs
+        log_stream_prefix from-fluent-bit-
+        log_retention_days 60
+        auto_create_group true
+```
+
+You can stream logs from Fargate directly to Amazon CloudWatch, using the output plugin shown above. Cloudwatch is currently the only
+supported output plugin in fluent-bit running on EKS fargate. Note that the log router setup depends on the prerequisites. Otherwise, one
```
sumo-drosiek, line 182:
```suggestion
supported output plugin in Fluent Bit running on EKS fargate. Note that the log router setup depends on the prerequisites. Otherwise, one
```

rnishtala-sumo:
I've made the suggested changes, will keep this documentation task open
https://jira.kumoroku.com/jira/browse/SUMO-205563
added the comment from this PR to the above ticket.

deploy/helm/sumologic/values.yaml
```diff
@@ -4374,6 +4389,49 @@ otellogs:
...
+          timeoutSeconds: 10
+          failureThreshold: 3
+        readinessProbe:
+          initialDelaySeconds: 5
+          periodSeconds: 10
+          timeoutSeconds: 3
+          failureThreshold: 3
+        startupProbe:
+          periodSeconds: 3
+          failureThreshold: 60
```
sumo-drosiek, line 4434: Please doucment it in README as github actions are failing, also, I think it's not `otellogs` anymore, eventually `otellogs.cloudwatch` or `otelcloudwatch`

sumo-drosiek (approved):
LGTM

PR #3003, opened 2023-04-25
feat(fargate): Fargate (cloudwatch) logs metadata enrichment

rnishtala-sumo:
Fargate (cloudwatch) logs metadata enrichment
Examples can be seen here: https://stagdata.long.sumologic.net/ui/#/search/create?id=t920BQQYEe72Ejb89Q7j7mOJhJHbxGrSZR6oRD9l

### Checklist


deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -1,14 +1,3 @@
-receivers:
```
rnishtala-sumo, line 1: Removed this because its a duplicate definition of the receiver

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -600,18 +589,8 @@ service:
         - fluentforward
 {{ end }}
 {{ end }}
-## Use this pipeline for cloudwatch logs
```
rnishtala-sumo, line 603: removed the cloudwatch logs pipeline from the metadata config. Cloudwatch logs will now re-use the container logs pipeline.

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
```diff
@@ -26,6 +26,33 @@ processors:
...
+    log_statements:
+      - context: resource
+        statements:
+          - set(attributes["k8s.pod.name"], attributes["cloudwatch.log.stream"])
+          - set(attributes["k8s.container.name"], attributes["cloudwatch.log.stream"])
+          - set(attributes["k8s.namespace.name"], attributes["cloudwatch.log.stream"])
+          - replace_pattern(attributes["k8s.pod.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
+          - replace_pattern(attributes["k8s.container.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
+          - replace_pattern(attributes["k8s.namespace.name"], "^.*kube\\.var\\.log\\.containers\\.[0-9a-zA-Z\\-]+\\_([a-zA-Z\\-]*)_.*", "$$1")
+  transform/parse:
```
swiatekm-sumo, line 39: Can you set `error_mode: ignore` here for safety? https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor#config

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
```diff
@@ -26,6 +26,33 @@ processors:
...
+      - context: log
+        statements:
+          - merge_maps(attributes, body, "insert")
+          - set(body, "") where IsMatch(body, "^{") == true
+  transform/move:
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["k8s.container.name"], resource.attributes["k8s.container.name"])
+          - set(attributes["k8s.deployment.name"], resource.attributes["k8s.deployment.name"])
```
swiatekm-sumo, line 55: Couldn't we do this directly in `transform/get_cloudwatch_pod`? And why aren't we moving `k8s.pod.name` as well?

rnishtala-sumo: It was split because of different contexts (resource vs log) but maybe this can be done instead
{code}
transform/cloudwatch_logs:
    error_mode: ignore
    log_statements:
      - context: resource
        statements:
          - set(attributes["k8s.pod.name"], attributes["cloudwatch.log.stream"])
          - set(attributes["k8s.container.name"], attributes["cloudwatch.log.stream"])
          - set(attributes["k8s.namespace.name"], attributes["cloudwatch.log.stream"])
          - replace_pattern(attributes["k8s.pod.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
          - replace_pattern(attributes["k8s.container.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
          - replace_pattern(attributes["k8s.namespace.name"], "^.*kube\\.var\\.log\\.containers\\.[0-9a-zA-Z\\-]+\\_([a-zA-Z\\-]*)_.*", "$$1")
      - context: log
        statements:
          - set(attributes["k8s.container.name"], resource.attributes["k8s.container.name"])
          - set(attributes["k8s.pod.name"], resource.attributes["k8s.pod.name"])
          - set(attributes["k8s.deployment.name"], resource.attributes["k8s.deployment.name"])
  transform/parse:
    log_statements:
      - context: log
        statements:
          - set(body, ParseJSON(body)) where IsMatch(body, "^{") == true
          - merge_maps(attributes, body, "insert")
          - set(body, "") where IsMatch(body, "^{") == true
 {code}

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
```diff
@@ -26,6 +26,29 @@ processors:
...
+  transform/cloudwatch_logs:
+    error_mode: ignore
+    log_statements:
+      - context: resource
+        statements:
+          - set(attributes["k8s.pod.name"], attributes["cloudwatch.log.stream"])
+          - set(attributes["k8s.container.name"], attributes["cloudwatch.log.stream"])
+          - set(attributes["k8s.namespace.name"], attributes["cloudwatch.log.stream"])
+          - replace_pattern(attributes["k8s.pod.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
+          - replace_pattern(attributes["k8s.container.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
```
astencel-sumo, line 38: Why is `k8s.pod.name` and `k8s.container.name` set to the same value? It looks like it's the pod name, which is fine in `k8s.pod.name`, but the `k8s.container.name` should rather contain the container name and not the pod name, right?

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
swiatekm-sumo: I'm actually a bit confused. Why do we even need to set record attributes here? Why is the second group of statements even necessary? These keys are supposed to be resource attributes to begin with.

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
aboguszewski-sumo: Sample stream name: `from-fluent-bit-kube.var.log.containers.logger-server-67fc8884f4-4vwft_demo_main-3dad569bd5d74c70072b461f986b92da15c10d1b745b8686c65d1395a95bb361.log`
I don't remember well, but the container name is either `main` or `main-...`, so yes, the regex definitely should be modified

aboguszewski-sumo:
If you're reusing the `containers` pipeline then you should probably modify the config file, for example here:
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/1564213dd3a5cf27cf2ec4ac146ff9ee02891345/deploy/helm/sumologic/conf/logs/otelcol/config.yaml#L26
you should set the condition as "containers enabled OR cloudwatch enabled"

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
rnishtala-sumo: We need to set the container name, because the container pipeline in the metadata layer, looks for the container name as a record attribute

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
rnishtala-sumo: right `logger-server-67fc8884f4-4vwft` is the pod name and `main` is the container name. I'll change the regex for the container name

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
rnishtala-sumo: yes, I believe the metadata pipeline moves the following record attributes to the resource level

```
groupbyattrs/containers:
    keys:
      - k8s.container.id
      - k8s.container.name
      - k8s.namespace.name
      - k8s.pod.name
      - _collector
  ```

rnishtala-sumo:
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3003#issuecomment-1523191894

I don't think we'll ever disable container logs with the cloudwatch collector, so maybe not having that extra condition is ok?

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
rnishtala-sumo: this change is completed

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
rnishtala-sumo: updated the container name regex in the PR

swiatekm-sumo:
To me, cloudwatch is simply an alternative way to ingest container logs. So if the user disabled container logs, cloudwatch should be disabled as well.

aboguszewski-sumo:
@rnishtala-sumo if you think so, then leave it as it is

swiatekm-sumo (approved):

PR #2901, opened 2023-03-01
test(integration): refactor test features

swiatekm-sumo:
### Checklist

- [ ] Changelog updated or skip changelog label added
- [ ] Documentation updated
- [ ] Template tests added for new features
- [ ] Integration tests added or modified for major features

PR #3012, opened 2023-04-27
ci: fix integration tests

swiatekm-sumo:
The actual fix is the removal of the cgroupdriver workaround we added for MacOS in #2901 and the upgrade to ubuntu 22.04 for some unknown reason. I've also upgraded kind images in the process.

This is probably still broken on MacOS, but it's more important to unblock the CI right now.

### Checklist


mat-rumian (approved):
LGTM

PR #3040, opened 2023-05-09
feat(logs): add ability to use OTLP source

swiatekm-sumo:
Add a setting for switching to the OTLP source for logs. I'm going to add a bit of documentation about this, but I'd rather add this support simultaneously for all signal types.

The source is created for all users with setup enabled, and the setting only controls if we're sending data to it.

### Checklist

- [x] Documentation updated


deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -630,7 +631,7 @@ service:
     ## - added transform/remove_attributes processor
     logs/otlp/kubelet:
       exporters:
-        - sumologic/systemd
+{{ include "logs.otelcol.systemd.exporters" . | indent 8 }}
```
astencel-sumo, line 634: Did you want `kubelet` here, as you did above in line 547?
```suggestion
{{ include "logs.otelcol.kubelet.exporters" . | indent 8 }}
```

docs/otlp-source.md
```diff
@@ -0,0 +1,41 @@
+# Using the OTLP source
+
+Historically, agents used by this Chart sent logs and metrics data to a [generic HTTP Source][http_source] in Sumo Logic. Ever since the
+[version 3][v3] release, the Chart predominantly uses the [OpenTelemetry Collector][otel], and consequently also the [OTLP protocol][otlp].
+The data needed to be converted to the formats the generic HTTP source supports before being sent.
+
+Recently, we've added support for directly sending data using the OTLP protocol to Sumo Logic. This is achieved using the [OTLP
+source][otlp_source]. This document explains how to use this new source with the Helm Chart.
+
+**Note:** OTLP Source is currently available for the following signal types: logs
```
astencel-sumo, line 10: This should be rephrased. The OTLP Source is available for all signal types: logs, metrics, traces. It is our chart that currently only supports sending logs to an OTLP Source.

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
swiatekm-sumo: Yes, good catch. They're the same right now, but could be different in the future.

docs/otlp-source.md
swiatekm-sumo: Good point, will change this to be a statement about support in the Helm Chart.

astencel-sumo (approved):

PR #3049, opened 2023-05-11
feat(fargate-docs): Adding aws policies needed as part of cloudwatch logs collection

rnishtala-sumo:
Added the following aws policies that are needed as part of cloudwatch logs collection
- A policy that provides permissions to EKS fargate to create cloudwatch log groups
- A policy that authenticates and allows reading the cloudwatch log streams

### Checklist
- [x] Documentation updated

### DIFF
--- a/docs/fargate.md
+++ b/docs/fargate.md
@@ -1,6 +1,6 @@
 # Fargate
 
-**NOTE: This is an experimental feature and it's not a subject of breaking changes policy.**
+**NOTE: This is the alpha release of EKS Fargate.**
 
 The following are some limitations of deploying this helm chart on EKS fargate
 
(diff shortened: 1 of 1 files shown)


docs/fargate.md
```diff
@@ -640,7 +665,70 @@ export oidc_provider=$(aws eks describe-cluster --name $CLUSTER --region $AWS_RE
...
+                "logs:DescribeLogGroups",
+                "logs:DescribeLogStreams",
+                "logs:DescribeSubscriptionFilters",
+                "logs:StartQuery",
+                "logs:Unmask",
+                "logs:DescribeMetricFilters",
+                "logs:FilterLogEvents",
+                "logs:GetLogGroupFields",
+                "logs:ListTagsForResource"
+            ],
```
aboguszewski-sumo, line 721: Now that I think of it, do we need all of these permissions?

aboguszewski-sumo (approved):

docs/fargate.md
rnishtala-sumo: I tested it with this policy, maybe I could reduce the permissions

aboguszewski-sumo: Yes, I originally also tested with this policy, but maybe we should limit this.
Unfortunately, I don't see any info on that in the `awscloudwatch` receiver's docs ☹️  https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.77.0/receiver/awscloudwatchreceiver

docs/fargate.md
```diff
@@ -1,6 +1,8 @@
 # Fargate
 
-**NOTE: This is an experimental feature and it's not a subject of breaking changes policy.**
+| Status    |         |
+| --------- | ------- |
+| Stability | [alpha] |
```
sumo-drosiek, line 5: Why do we use Open Telemetry stability here?

I believe Fargate is more complex in terms of support and it is not fully dependent on receiver support

rnishtala-sumo: it was discussed in one if the standups last week that once we were confident with the E2E tests, this would be marked alpha
https://docs.google.com/document/d/1qC471RG2YTKw3PjDaqtQ6Rt9U8UaR3yeFwkqV0ZuplA/edit#heading=h.cz8mbgjlmnay

docs/fargate.md
rnishtala-sumo: The manual test for this seemed to be failing, i.e the logs weren't being collected after  I removed "logs:ListTagsForResource" for example. I can run more tests and raise a future PR if this can be reduced and merge these tested permissions for now, WDYT?

docs/fargate.md
sumo-drosiek: Sure, but we shouldn't use `https://github.com/open-telemetry/opentelemetry-collector#alpha` as reference. Until we don't have stability concept in the repository I would either add it to all components or make it descriptive way (like we did with experimental)

rnishtala-sumo: ok got it, I'll just make it descriptive then like we did with experimental.

rnishtala-sumo: done

sumo-drosiek (approved):

PR #3051, opened 2023-05-12
feat: allow setting otelcol image globally

### LINKED ISSUE #2700: Add single otelcol.image section for all of the components and allow overrides for each component separately
**Is your feature request related to a problem? Please describe.**
Right now if someone wishes to change the `otelcol` version or use fips binaries they need to change it in a bunch of places.

See this PR:
* https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2697

**Describe the solution you'd like**
I'd like to have a single `otelcol.image` section which will be a default for all of the components with possibility to override it in every component separately.

**Describe alternatives you've considered**
Alternative is what we have now - no default section.

**Additional context**
Right now to set fips binaries I need to change it in 6 places:
image

swiatekm-sumo:
Add a new config value for setting otelcol image for all components:

```yaml
sumologic:
  otelcolImage:
      repository: ...
      tag: ...
```

I've also set all the component-specific settings to be empty and the underlying resources will use the default image in this case.

This is implemented with some rather mundane template functions. Most of the added lines are goldenfile tests I've added, that really should've been there to begin with.

Fixes #2700 

### Checklist

- [X] Documentation updated


aboguszewski-sumo:
I guess this will resolve #2700?

sumo-drosiek (approved):

PR #3059, opened 2023-05-18
feat(EKS Fargate): Add multiline support to EKS Fargate

rnishtala-sumo:
Add multiline support to EKS Fargate

https://stagdata.long.sumologic.net/ui/#/search/create?id=HS2q7MasCkcxQkWvkMdVFUEfYJ3Ku7DYbByPc4XL 

- [x] Documentation updated

### DIFF
--- a/deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
@@ -35,13 +35,6 @@ processors:
   groupbyattrs/stream:
     keys:
       - cloudwatch.log.stream
-  ## need to reset the source identifier after grouping
-  transform/reset_source_identifier:
-    error_mode: ignore
-    log_statements:
-      - context: log
-        statements:
-          - set(attributes["cloudwatch.log.stream"], resource.attributes["cloudwatch.log.stream"])
   transform/parsejson:
     error_mode: ignore
     log_statements:
@@ -69,16 +62,14 @@ processors:
         is_last_entry: attributes.logtag == "F"
         output: "merge-multiline-logs"
         overwrite_with: newest
-        source_identifier: attributes["cloudwatch.log.stream"]
+        source_identifier: resource["cloudwatch.log.stream"]
         type: recombine
       - id: merge-multiline-logs
         combine_field: attributes.log
         combine_with: "\n"
         is_first_entry: attributes.log matches {{ .Values.sumologic.logs.multiline.first_line_regex | quote }}
-        source_identifier: attributes["cloudwatch.log.stream"]
+        source_identifier: resource["cloudwatch.log.stream"]
         type: recombine
-      - field: attributes["cloudwatch.log.stream"]
-        type: remove
 receivers:
   awscloudwatch:
     region: {{ .Values.sumologic.logs.collector.otelcloudwatch.region }}
@@ -100,7 +91,6 @@ service:
       processors:
         - transform/set_source_identifier
         - groupbyattrs/stream
-        - transform/reset_source_identifier
         - transform/parsejson
         - logstransform/cloudwatch
         - transform/metadata
--- a/deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
@@ -26,7 +26,31 @@ processors:
     send_batch_max_size: 2000
     send_batch_size: 1000
     timeout: 1s
-  transform/cloudwatch_logs:
+  transform/set_source_identifier:
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["cloudwatch.log.stream"], resource.attributes["cloudwatch.log.stream"])
+  groupbyattrs/stream:
+    keys:
+      - cloudwatch.log.stream
+  ## need to reset the source identifier after grouping
+  transform/reset_source_identifier:
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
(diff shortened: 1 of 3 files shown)


deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
```diff
@@ -26,24 +26,59 @@ processors:
...
-          - set(attributes["k8s.namespace.name"], resource.attributes["cloudwatch.log.stream"])
-          - replace_pattern(attributes["k8s.pod.name"], "^.*kube\\.var\\.log\\.containers\\.([0-9a-zA-Z\\-]+)\\_.*", "$$1")
-          - replace_pattern(attributes["k8s.container.name"], "^.*kube\\.var\\.log\\.containers\\.[0-9a-zA-Z\\-]+\\_[a-zA-Z\\-]*\\_([a-zA-Z]*).*", "$$1")
-          - replace_pattern(attributes["k8s.namespace.name"], "^.*kube\\.var\\.log\\.containers\\.[0-9a-zA-Z\\-]+\\_([a-zA-Z\\-]*)_.*", "$$1")
-  transform/parse:
+          - set(attributes["cloudwatch.log.stream"], resource.attributes["cloudwatch.log.stream"])
+  groupbyattrs/stream:
+    keys:
+      - cloudwatch.log.stream
+  ## need to reset the source identifier after grouping
```
sumo-drosiek, line 38: Why do we need to reset it? We can use resource attribute in logstransform's recombine operator. Now I see we set it to use in logstransform and hen we remove it, which seems like unnecessary overhead

sumo-drosiek (changes requested):
IMHO, no need to move `cloudwatch.log.stream` to record level atributes. We whoulc be able to use resource level attribute

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
aboguszewski-sumo: Yes, I agree. Initially we didn't think about that, but that makes a lot of sense.

rnishtala-sumo: I think we need to set the cloudwatch.log.stream as a record attribute so that we could use it to group all the log records under a resource. If we don't set this as a record attribute, then what would I group by?

this is from the doc for groupby attr
```
If the log record and metric data point has at least one of the specified attributes key, it will be moved to a Resource with the same value for these attributes
```

rnishtala-sumo: yes that makes sense, made the change.

sumo-drosiek (approved):
Thanks!

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
```diff
@@ -63,8 +89,11 @@ service:
       receivers:
         - awscloudwatch
       processors:
-        - transform/parse
-        - transform/cloudwatch_logs
+        - transform/set_source_identifier
+        - groupbyattrs/stream
```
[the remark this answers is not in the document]
rnishtala-sumo: yeah we're grouping all the logs by the cloudwatch log stream. we'll also plan to raise an upstream issue for the cloudwatch receiver so that the logs are placed (correctly) under a log stream resource.

deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml
rnishtala-sumo: @sumo-drosiek @aboguszewski-sumo actually I get the following warning when I try to use the log stream resource attribute as the source identifier for the recombine operator

```2023-05-19T17:00:15.352Z    warn    recombine/recombine.go:235    entry does not contain the source_identifier, so it may be pooled with other sources    {"kind": "processor", "name": "logstransform/cloudwatch", "pipeline │
│ ": "logs/collector/otelcloudwatch", "operator_id": "merge-cri-lines", "operator_type": "recombine"}```

I'll revert to copying the log stream attribute to record level to prevent unexpected issues.

Comment (rnishtala-sumo) [reply deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml:38]: confirmed that doing this removes the above warning. In the interest of avoiding a scenario where we combine logs from multiple sources (as the above warning indicates), I'd like to keep the original changes

Comment (sumo-drosiek) [reply deploy/helm/sumologic/conf/logs/collector/otelcloudwatch/config.yaml:38]: The following is working for me:

configuration:
```
receivers:
  filelog:
    include:
      - ./tmp/logs/multiline.json  
    start_at: beginning
exporters:
  logging:
    verbosity: detailed
processors:
  logstransform/containers_parse_json:
    operators:
      - if: body matches "^{[\\s\\S]+"
        parse_from: body
        parse_to: body
        type: json_parser
      - type: add
        field: resource["cloudwatch.log.stream"]
        value: resource_attribute
  logstransform:
    operators:
      - id: merge-multiline-logs
        combine_field: body.log
        combine_with: "\n"
        is_first_entry: body.log matches "^a"
        source_identifier: resource["cloudwatch.log.stream"]
        # source_identifier: attributes["log.file.name"]
        type: recombine
service:
  pipelines:
    logs:
      receivers:
        - filelog
      processors:
        - logstransform/containers_parse_json
        - logstransform
      exporters:
        - logging
```

file:
```
{"log": "abc"}
{"log": "def"}
{"log": "ghi"}
{"log": "asdc"}
```

rnishtala-sumo: changed the source id to `resource["cloudwatch.log.stream"]` and that seems to work. thanks!

sumo-drosiek (approved):

PR #3061, opened 2023-05-22
test(integration): mark metrics used by Apps

swiatekm-sumo:
I've reviewed all the metrics we test for and marked the ones used in Sumo Apps. Might also be useful to review the application metrics we don't test for, but that can happen in a separate change.

### Checklist

### DIFF
--- a/tests/integration/internal/constants.go
+++ b/tests/integration/internal/constants.go
@@ -82,68 +82,68 @@ const (
 
 // metrics we expect the receiver to get
 // metrics which are expected in current E2E tests but aren't present here, are commented out
-// TODO: figure out why the expected metrics aren't present
+// The comments about App usage are based on our internal content repository
 var (
-	KubeStateMetrics = []string{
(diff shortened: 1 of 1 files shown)


swiatekm-sumo:
> This shows the current state, which is good, but does not give any way to keep this up to date.

That's definitely not going to happen in this PR. I'm open to representing this information differently, but keeping it up to date is necessarily down to our internal tooling.

astencel-sumo:
> > This shows the current state, which is good, but does not give any way to keep this up to date.
> 
> That's definitely not going to happen in this PR. I'm open to representing this information differently, but keeping it up to date is necessarily down to our internal tooling.

It doesn't need to be done in this PR. My point is that if we don't intend to provide an automation to keep this up to date, then I'm not sure if there's much value in putting it there in the first place.

astencel-sumo:
As a minimum, can you somehow document how you got this data (and how anyone can do that again to verify the data)?

sumo-drosiek (approved):

PR #3063, opened 2023-05-22
chore(logs): use transform processor for json parsing

swiatekm-sumo:
Use transform processor for JSON log parsing instead of the logstransform processor. The former is more stable and better supported, while the latter is slated for removal once OTTL has all the necessary features.

### Checklist

### DIFF
--- a/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
@@ -519,6 +519,7 @@ service:
         - resource/containers_copy_node_to_host
         - sumologic_schema
         - source/containers
+        - transform/containers_parse_json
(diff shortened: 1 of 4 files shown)


deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -605,7 +603,7 @@ service:
         - resource/containers_copy_node_to_host
         - sumologic_schema
         - source/containers
-        - logstransform/containers_parse_json
```
sumo-drosiek, line 608: Shouldn't we add it to `logs/fluent/containers` as well?

swiatekm-sumo: We should, will fix.

sumo-drosiek (approved):

PR #3072, opened 2023-05-30
feat(metrics): make deleting metrics source metadata optional

ryan-codaio:
Currently, there is no way to set the source category for metrics, as we are stripping the `_sourceCategory` attribute. This happens after the `extraProcessors` are added, so there is no way to override this behavior. This is pretty  important for us, as we need to disambiguate metrics from different environments.

The [comment for this `delete_source_metadata` processor](https://github.com/ryan-codaio/sumologic-kubernetes-collection/blob/7535536f8fb61bb3395b72745d0d6128019d8b66/deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml#L102-L105) says it's because we need to wait until we can send these attributes as headers, but the [sumologic exporter has been doing this](https://github.com/SumoLogic/sumologic-otel-collector/blob/a46c466/pkg/exporter/sumologicexporter/sender.go#L699-L709) since Oct 2022. In tests on our infra with this processor removed, this has been working fine.

I added a configuration parameter defaulting to `true` to preserve the old behavior by default, to avoid any unexpected surprises with a sudden cardinality increase.

### Checklist

- [x] Documentation updated


sumo-drosiek:
We are working on improving source metadata for metrics, as it doesn't make sense to use default template for some of them (cluster wide). Workaround for this PR is to use the following configuration:

```
metadata:
  metrics:
    config:
      merge:
        processors:
          resource/delete_source_metadata:
            attributes:
              # attributes list cannot be empty so using non-existing key
              - key: _
                action: delete
```

ryan-codaio:
Thank you for the workaround! This seems to have worked well for us.

Astitva-sumo:
Closing stale PR.

PR #3076, opened 2023-06-04
feat(chart): Allow passing extra args to otel logs stateful set

rupeshnemade:
### Checklist

- [x] Documentation updated


deploy/helm/sumologic/README.md
```diff
@@ -569,6 +569,7 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
+| `metadata.logs.statefulset.extraArgs`                                            | Additional arguments to otelcol container.                                                                                                                                                                                                                                                                                                                           | `["--config=/etc/otel/config.yaml"]`                                                                                                                                                 |
```
swiatekm-sumo, line 572: The default argument here is confusing imo, as it's the same as the default, but we're adding to the default instead of replacing it. Can we just make the default empty instead?

rupeshnemade: @swiatekm-sumo If I keep it empty then PR checker fails in comparing readme data with actual received values.

swiatekm-sumo: You need to make it empty in the commented-out section as well.

rupeshnemade: @swiatekm-sumo I have addressed you comment by removing args from comment section. Can you trigger PR checker?

swiatekm-sumo (approved):

swiatekm-sumo:
Please rebase this on `main` and we should be good to go. :+1:

.changelog/3075.changed.txt
```diff
@@ -1 +0,0 @@
-chore: upgrade Fluent Bit to 2.1.4
```
swiatekm-sumo, line 1: Why are you deleting this?

swiatekm-sumo (changes requested):

PR #3078, opened 2023-06-07
Reordering HPA spec.metrics to match HPA ordering

poblahblahblah:
This addresses an issue when using ArgoCD (and maybe other GitOps operators) where Kubernetes reorders the objects under the spec.metrics key thus causing Sync issues with ArgoCD.

Originally reported to the ArgoCD project here:
https://github.com/argoproj/argo-cd/issues/1079

Originally reported to the Kubernetes project here: https://github.com/kubernetes/kubernetes/issues/74099

Other projects and companies have also addressed this by simply reordering the metrics section:

* https://github.com/kubernetes/ingress-nginx/pull/10043
* https://github.com/nginxinc/kubernetes-ingress/pull/3773
* https://github.com/grafana/helm-charts/pull/758
* https://github.com/open-telemetry/opentelemetry-helm-charts/pull/103
* https://github.com/Nextdoor/k8s-charts/pull/102

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3078.changed.txt
+++ b/.changelog/3078.changed.txt
@@ -1,0 +1,1 @@
+Reordering HPA metrics to match HPA ordering
--- a/deploy/helm/sumologic/templates/logs/common/hpa.yaml
+++ b/deploy/helm/sumologic/templates/logs/common/hpa.yaml
@@ -18,12 +18,6 @@ spec:
   minReplicas: {{ .Values.metadata.logs.autoscaling.minReplicas }}
   maxReplicas: {{ .Values.metadata.logs.autoscaling.maxReplicas }}
   metrics:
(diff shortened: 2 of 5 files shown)


sumo-drosiek:
@poblahblahblah Thanks for the PR!

Seems that templates are not generated correctly. Probably there is now newline rendered because of `-` in helm templates

swiatekm-sumo:
It's a bit concerning that template tests don't catch this. We should have a test that unmarshals all the templates into proper K8s objects to check for structural issues like this one. Right now, the tests just use Unstructured, which is just a glorified map with metadata.

swiatekm-sumo (approved):

PR #3098, opened 2023-06-23
chore: Upgrade kubernetes setup to v3.9.0

rnishtala-sumo:
Upgrade kubernetes setup to v3.9.0

### Checklist

- [x] Documentation updated


tests/helm/testdata/goldenfile/setup/setup/basic.output.yaml
```diff
@@ -33,7 +33,7 @@ spec:
             defaultMode: 0777
       containers:
         - name: setup
-          image: public.ecr.aws/sumologic/kubernetes-setup:3.8.0
+          image: public.ecr.aws/sumologic/kubernetes-setup:%CURRENT_CHART_VERSION%
```
astencel-sumo, line 36: I think it's only a conincidence that the current setup image version `3.9.0` is equal to current chart version `3.9.0`, right? Which means we should still use hardcoded value `3.9.0` here instead of the current chart version replacement string.

swiatekm-sumo (approved):

tests/helm/testdata/goldenfile/setup/setup/basic.output.yaml
rnishtala-sumo:
```
-       (string) (len=5) "image": (string) (len=47) "public.ecr.aws/sumologic/kubernetes-setup:3.9.0",
+       (string) (len=5) "image": (string) (len=65) "public.ecr.aws/sumologic/kubernetes-setup:%CURRENT_CHART_VERSION%"
```
when the test runs the actualObject seems to have `public.ecr.aws/sumologic/kubernetes-setup:%CURRENT_CHART_VERSION%` which doesn't match `public.ecr.aws/sumologic/kubernetes-setup:3.9.0`
causing the test to fail

rnishtala-sumo: @astencel-sumo @swiatekm-sumo I regenerated this file using `make regenerate-goldenfiles` and it substituted the version with `%CURRENT_CHART_VERSION%`

swiatekm-sumo: The regenerate command isn't very smart and can get confused at edge cases. You need to manually fix its mistakes at times.

rnishtala-sumo: ok, it looks like the test does fail when the expected is hardcoded to `3.9.0`. Does the `basic.input.yaml` need to be changed?

tests/helm/goldenfile_test.go
```diff
@@ -132,9 +132,10 @@ func runGoldenFileTest(t *testing.T, valuesFileName string, outputFileName strin
 // expected templates
 func fixupRenderedYaml(yaml string, chartVersion string) string {
 	checksumRegex := regexp.MustCompile("checksum/config: [a-z0-9]{64}")
+	chartsRegex := regexp.MustCompile("([^kubernetes-setup:])" + chartVersion)
```
rnishtala-sumo, line 135: not replacing the chart version using a regex

tests/helm/testdata/goldenfile/setup/setup/basic.output.yaml
rnishtala-sumo: not replacing the k8s-setup chart version using a regex, seems to fix this issue

PR #3106, opened 2023-06-28
fix: only add OTLP source env variables if they're enabled

swiatekm-sumo:
We've been adding OTLP sources unconditionally, which can break installations where setup is disabled even if the otlp source is also disabled. This change makes us add the OTLP source url as an env variable to the relevant StatefulSet only if OTLP sources are actually enabled.

Also fixed an issue where some tracing tests weren't run due invalid file names.
### Checklist

- [X] Documentation updated


tests/helm/testdata/goldenfile/traces-gateway-deployment/basic.output.yaml
```diff
@@ -48,11 +48,6 @@ spec:
                 secretKeyRef:
                   name: sumologic
                   key: endpoint-traces
-            - name: SUMO_ENDPOINT_DEFAULT_OTLP_TRACES_SOURCE
```
rnishtala-sumo, line 51: is this removed from the template because the otlp source is disabled by default?

tests/helm/testdata/goldenfile/traces-gateway-deployment/custom.output.yaml
```diff
@@ -41,16 +48,27 @@ spec:
                 secretKeyRef:
                   name: sumologic
                   key: endpoint-traces
+            - name: SUMO_ENDPOINT_DEFAULT_OTLP_TRACES_SOURCE
```
rnishtala-sumo, line 51: this is here because we enabled the otlp source in the input, correct?

deploy/helm/sumologic/values.yaml
```diff
@@ -662,6 +662,8 @@ sumologic:
     ## How many spans per request should be send to receiver
     spans_per_request: 100
 
+    sourceType: http
```
rnishtala-sumo, line 665: shouldn't  this be `otlp` as per readme below?

tests/helm/testdata/goldenfile/traces-gateway-deployment/basic.output.yaml
swiatekm-sumo: Yes.

tests/helm/testdata/goldenfile/traces-gateway-deployment/custom.output.yaml
swiatekm-sumo: Yes.

deploy/helm/sumologic/values.yaml
swiatekm-sumo: As per what README? Traces use a special HTTP traces source by default, which does accept OTLP, but is supposed to be replaced by our proper OTLP source.

rnishtala-sumo (approved):

PR #3120, opened 2023-06-30
docs: add missing entries to changelog

kkujawa-sumo:
### Checklist


swiatekm-sumo (changes requested):
We should instead fix this release. We shouldn't have all of these changes in a patch version.

PR #3121, opened 2023-06-30
feat(metrics/collector): allow disabling cadvisor and kubelet metrics

swiatekm-sumo:
This is to make gradual migrations easier. prometheus-operator maintains special Services which allow the collection of these metrics to happen using ServiceMonitors. Without it, you need a raw scrape config, which the otel metrics collector uses.

However, if you're gradually moving ServiceMonitors between Prometheus and otel using selectors, you can have both otel and prometheus-operator enabled at the same time, and otel responsible for kubelet and cAdvisor ServiceMonitors. In that case, you want to disable the otel scrape configs, which is what this change allows.

### Checklist

- [ ] Documentation updated


deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -67,6 +68,8 @@ receivers:
               - __address__
               target_label: instance
               action: replace
```
sumo-drosiek, line 70: maybe we can template it in helm :thinking:

sumo-drosiek (approved):
LGTM

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
swiatekm-sumo: I don't think it's worth it for just two copies.

PR #3127, opened 2023-06-30
docs: Update the release guide for creating/pushing tags

rnishtala-sumo:
Update the release guide for creating/pushing tags

- [x] Documentation updated

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -12,14 +12,10 @@ Perform the following steps in order to release new verions of helm chart.
      - add link to minor version, if created
      - set "supported until" date for previous minor version to 6 months after today
 
-1. Create and push new tag:
+1. Define a new tag:
 
    ```bash
    export TAG=x.y.z
-   git checkout main
-   git pull
-   git tag -sm "v${TAG}" "v${TAG}"
-   git push origin "v${TAG}"
    ```
 
 1. Prepare release branch:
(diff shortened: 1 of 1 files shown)


sumo-drosiek (changes requested):
We don't want to tag a release branch (with major releases), as `git hash` is going to give us wrong value if it won't be on `main` branch.

rnishtala-sumo:
I didn't think that git hash depended on a branch. Just did a quick search on this and this popped up.

A commit hash depends on
- The commit message
- The file changes
- The commit author (and committer- they can be different)
- The date
- The parent commit hash

As long as we have the correct list of commits in the release branch, it seems like we should be able to push a tag on this branch and create a release, irrespective of major, minor or patch?

sumo-drosiek:
Oh, sorry, I meant `git describe`:
```
> git describe --tags              
v3.9.1-7-gd373d1ca
```

rnishtala-sumo:
yes that's true, so all the patch release tags would be on the release branch and all the major/minor release tags would go on the `main` branch.

sumo-drosiek:
> yes that's true, so all the patch release tags would be on the release branch and all the major/minor release tags would go on the `main` branch.

Yes, exactly

sumo-drosiek:
@rnishtala-sumo can we close it?

PR #3130, opened 2023-07-04
feat(metrics): collect full etcd histogram metrics

swiatekm-sumo:
We currently only collect the `_bucket` parts of some etcd histogram metrics. This is a historical holdover that doesn't make much sense, as the whole histogram is significantly more useful, and we save very little DPM this way.

With this PR, we collect the full histograms.

### Checklist


sumo-drosiek (approved):
~Could you provide example~
nvm, `count` and `sum` just single values

PR #3132, opened 2023-07-05
test(integration): strict checks for collected metrics

swiatekm-sumo:
Check whether we collect exactly the metrics we expect. This means the tests fail if they see a metric not on an explicit allowlist.

Doing this required making the metric selection logic more complicated, and it's probably due for a refactor. But it does work and shows the exact difference between the Otel and Prometheus metric collectors.

### Checklist


sumo-drosiek (approved):
Now they are better than our e2e tests :1st_place_medal:

PR #3140, opened 2023-07-07
docs(prometheus): Added a section on prometheus sharding

rnishtala-sumo:
Added a section on prometheus sharding

- [x] Documentation updated

### DIFF
--- a/.changelog/3140.added.txt
+++ b/.changelog/3140.added.txt
@@ -1,0 +1,1 @@
+docs(prometheus): Added a section on prometheus sharding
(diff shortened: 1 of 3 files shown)


sumo-drosiek (approved):

swiatekm-sumo (changes requested):
We should mention that this will cause recording rule metrics which require global aggregations to stop working - and with them, possibly the Kubernetes App.

kkujawa-sumo (approved):

swiatekm-sumo (approved):

PR #3141, opened 2023-07-10
fix(metrics/collector): drop k8s labels

swiatekm-sumo:
Prometheus receiver adds some attributes based on Prometheus' meta labels. This cannot be turned off explicitly or prevented by deleting the labels in relabel configs, because the labels are extracted from the context.

Drop these labels in a processor, as we don't need them, and their presence makes the pipeline more difficult to understand.

I've also added a test specifically for metadata of kube-state-metrics, which have specific requirements. For example, Prometheus incorrectly assigns Pod metadata to them, and the test checks whether we remove it.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3141.fixed.txt
+++ b/.changelog/3141.fixed.txt
@@ -1,0 +1,1 @@
+fix(metrics/collector): drop k8s labels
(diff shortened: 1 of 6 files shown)


sumo-drosiek (approved):
LGTM

Issue #3154, opened 2023-07-13
Unable to apply sumo logic helm chart after upgrading k8s clusters to 1.26


jlawrienyt:
**Describe the bug**

When upgrading a k8s cluster from version 1.25 to 1.26 where the Sumo Logic helm chart had already been applied, the next time the chart was applied, we encountered the following error:

```
Error: UPGRADE FAILED: unable to build kubernetes objects from current release manifest: resource mapping not found for name: "sumologic-sumologic-otelcol-logs" namespace: "" from "": no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2beta2"
```

**To Reproduce**

1. Run a k8s clusters v1.25 and install the Sumo Logic helm chart.
2. Upgrade the k8s cluster in place to v1.26
3. Upgrade the Sumo Logic helm chart

**Workaround**

Use the [helm mapkubeapis plugin](https://github.com/helm/helm-mapkubeapis) to update the helm release metadata in the cluster before re-applying the Sumo Logic helm chart in the 1.26 k8s cluster.  Note that as of 07/13/2023, the plugin's default configuration does not contain the autoscaling/v2beta2 API, so this will need to be created locally as 

```yaml
mappings:
  - deprecatedAPI: "apiVersion: autoscaling/v2beta2\nkind: HorizontalPodAutoscaler\n"
    newAPI: "apiVersion: autoscaling/v2\nkind: HorizontalPodAutoscaler\n"
    deprecatedInVersion: "v1.23"
    removedInVersion: "v1.26"
```

and then applied using 

```sh
helm mapkubeapis sumologic --mapfile Map.yaml
```

**Expected behavior**

Applying the Sumo Logic helm chart should succeed.  The logic in the helm chart should install the HorizontalPodAutoscaler resources using api version autoscaling/v2 when the autoscaling/v2 API was introduced, i.e. when the k8s cluster is version 1.23, not when the autoscaling/v2beta2 API was removed (v1.26).

PR #3155, opened 2023-07-13
Change the minimum version for using autoscaling/v2 api to 1.23

jlawrienyt:
This addresses issue #3154

### Checklist

### DIFF
--- a/.changelog/3155.fixed.txt
+++ b/.changelog/3155.fixed.txt
@@ -1,1 +1,1 @@
-Change the minimum version for using autoscaling/v2 api to 1.23
+feat(helm): change the minimum version for using autoscaling/v2 api to 1.23
--- a/.changelog/3155.fixed.txt
+++ b/.changelog/3155.fixed.txt
@@ -1,0 +1,1 @@
+Change the minimum version for using autoscaling/v2 api to 1.23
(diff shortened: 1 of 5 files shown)


.changelog/3155.fixed.txt
```diff
@@ -0,0 +1 @@
+Change the minimum version for using autoscaling/v2 api to 1.23
```
sumo-drosiek, line 1:
```suggestion
feat(helm): change the minimum version for using autoscaling/v2 api to 1.23
```

swiatekm-sumo:
From the issue, it sounds like you run into trouble if you have the Chart installed on a lower version (and therefore are using the deprecated api group), and then upgrade to 1.26, where that api group was removed. Kubernetes itself has no issue with this, and in fact will internally convert to the latest supported api group, but Helm itself won't, and reports an issue with the existing manifests - even if the upgrade would, in fact, fix the problem.

sumo-drosiek:
> From the issue, it sounds like you run into trouble if you have the Chart installed on a lower version (and therefore are using the deprecated api group), and then upgrade to 1.26, where that api group was removed. Kubernetes itself has no issue with this, and in fact will internally convert to the latest supported api group, but Helm itself won't, and reports an issue with the existing manifests - even if the upgrade would, in fact, fix the problem.

Then it's better to check for `1.23`, so the issue won't be a problem if he first installation has been done on at least `1.23`

swiatekm-sumo:
> 



> > From the issue, it sounds like you run into trouble if you have the Chart installed on a lower version (and therefore are using the deprecated api group), and then upgrade to 1.26, where that api group was removed. Kubernetes itself has no issue with this, and in fact will internally convert to the latest supported api group, but Helm itself won't, and reports an issue with the existing manifests - even if the upgrade would, in fact, fix the problem.
> 
> Then it's better to check for `1.23`, so the issue won't be a problem if he first installation has been done on at least `1.23`

That's what this PR does, right?

sumo-drosiek (approved):

PR #3157, opened 2023-07-17
feat(metrics): do not filter in default remote write definitions

swiatekm-sumo:
Merge most of the default remote writes into a single one that does (almost) no data transformations. We filter by job name instead of metric name.

* Node metrics still have their own, due to difficulty handling recording rule metrics.
* Collection internal metrics need some additional work before they can be handled this way.
* We need to manually drop Prometheus' scrape metrics.

This doesn't affect users who modified these, as that requires overriding the whole array, and this is therefore not a breaking change.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3157.changed.txt
+++ b/.changelog/3157.changed.txt
@@ -1,0 +1,1 @@
+feat(metrics): do not filter in default remote write definitions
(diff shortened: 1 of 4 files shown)


sumo-drosiek (approved):
Looks fine for me. It would be nice to have some comparison e2e :sweat_smile: 
maybe some day

PR #3159, opened 2023-07-20
feat: add hpa for tracing

sumo-drosiek:
Adds HPA for tracing components we can add it:

- tracesGateway
- otelcolInstrumentation


swiatekm-sumo:
I think the reason we haven't done this yet is that different replica counts require different cascading filter settings. Right @mat-rumian ?

sumo-drosiek:
> I think the reason we haven't done this yet is that different replica counts require different cascading filter settings. Right @mat-rumian ?

This is true for trace sampler, which I'm not touching in this PR

swiatekm-sumo (approved):

PR #3160, opened 2023-07-20
feat(metrics): disable default Prometheus rules

swiatekm-sumo:
kube-prometheus-stack installs a bunch of alerting and recording rules by default, of which we use two for our own recording rules. The rest aren't necessary and make filtering existing metrics more difficult.

We don't support any of these, so it's not a breaking change, but I've called it out in the changelog for users who may have configured kube-prometheus-stack to use these.

### Checklist

- [ ] Documentation updated


tests/integration/internal/constants.go
```diff
@@ -359,6 +359,7 @@ var (
 		"otelcol_otelsvc_k8s_pod_deleted",
 		"otelcol_processor_batch_batch_size_trigger_send",
 		"otelcol_otelsvc_k8s_ip_lookup_miss",
+		"otelcol_otelsvc_k8s_other_deleted",
```
swiatekm-sumo, line 362: This is unrelated, just another flaky metric I found running tests locally.

sumo-drosiek (approved):

PR #3168, opened 2023-07-26
Add sumologic.logs.kubelet.enabled key

tangyfruits:
### Checklist

Changelog + tests to come

- [x] Documentation updated


sumo-drosiek:
Kubelet logs are systemd logs as well.

Please see the following code: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/7c1008b7e8c68f3d27c1b7dd40cbe1b92adf8825/deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml#L265

This should be handled properly, otherwise kubelet logs won't work anyway

tangyfruits:
> Kubelet logs are systemd logs as well.
> 
> Please see the following code:
> 
> https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/7c1008b7e8c68f3d27c1b7dd40cbe1b92adf8825/deploy/helm/sumologic/conf/logs/collector/otelcol/config.yaml#L265
> 
> This should be handled properly, otherwise kubelet logs won't work anyway

Hmm fair call. I did find it quite confusing that the kubelet config was a sibling of systemd. I had to dig into the helmchart to work out that I had to enable systemd. Maybe there is consideration make kubelet log config a child of systemd instead? It was unintuitive to me using the chart to have to enable systemd - not sure if anyone else agrees

sumo-drosiek:
> Hmm fair call. I did find it quite confusing that the kubelet config was a sibling of systemd. I had to dig into the helmchart to work out that I had to enable systemd. Maybe there is consideration make kubelet log config a child of systemd instead? It was unintuitive to me using the chart to have to enable systemd - not sure if anyone else agrees

This is fair ask. I think it should be separated to avoid confusion, but this is not our priority right now.

As a workaround, you may override `sumologic.logs.systemd.units` to use only `kubelet` related, then even if systemd (not kubeelt) pipeline will be enabled it won't get any data to process

Astitva-sumo:
Closing stale PR

PR #3171, opened 2023-07-27
fix(metrics): don't forward duplicate metrics from annotated Pods

swiatekm-sumo:
Don't forward App metrics if they originate from a Pod scraped due to annotations. We tell customers to do this for these metrics (https://help.sumologic.com/docs/integrations/web-servers/nginx/#configure-metrics-collection), but we have separate remote writes for these cases, and so they are forwarded twice, and with different labels.

I ended up making a bunch of additional changes:
* I've added a integration test for nginx metrics using Telegraf operator
* Refactored some receiver-mock related code in integration tests
* Added a utility function to assert metrics are present for a given filter
* Fixed some minor issues

### Checklist

- [ ] Documentation updated

### DIFF
--- a/deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
@@ -38,7 +38,7 @@ processors:
     metrics:
       metric:
         # we let the metrics from annotations ("kubernetes-pods") through as they are
-        - resource.attributes["service.name"] != "kubernetes-pods" and IsMatch(name, "scrape_.*")
+        - resource.attributes["service.name"] != "pod-annotations" and IsMatch(name, "scrape_.*")
 
 receivers:
   prometheus:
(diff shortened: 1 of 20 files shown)


deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -38,7 +38,7 @@ processors:
     metrics:
       metric:
         # we let the metrics from annotations ("kubernetes-pods") through as they are
```
sumo-drosiek, line 40:
```suggestion
        # we let the metrics from annotations ("pod-annotations") through as they are
```

tests/integration/features.go
```diff
@@ -413,9 +436,6 @@ func GetTracesFeature() features.Feature {
 				// "_sourceCategory":    "kubernetes/customer/trace/tester/customer/trace/tester",
 				"_sourceName": fmt.Sprintf("%s.%s.%s", internal.TracesGeneratorNamespace, internal.TracesGeneratorName, internal.TracesGeneratorName),
 			},
-			internal.ReceiverMockNamespace,
-			internal.ReceiverMockServiceName,
-			internal.ReceiverMockServicePort,
```
sumo-drosiek, line 418: We are not going to check receiver-mock metrics anymore?

sumo-drosiek (approved):
Going to merge it in order to put into the release

tests/integration/features.go
swiatekm-sumo: This doesn't change any of the test checks, I just moved this logic to a separate function which gets these values from module constants.

swiatekm-sumo: I do want to get rid of checking for receiver-mock metrics, because they're annoying to work with - receiver-mock only emits some of them based on ingested data, so they're different for each test - but I didn't do that in this PR.

PR #3192, opened 2023-08-07
feat(metrics/collector): support partial histograms

swiatekm-sumo:
Support collecting only the count and sum components of a histogram. Prometheus does this out of the box, but otel collector is stricter about enforcing both the OpenMetrics and Otel specs. This means that if we only collect the count and sum, we also get one bucket which covers the whole histogram for compliance with the spec. This change drops that bucket in a processor to keep compatibility with what Prometheus does.

### Checklist

- [x] Documentation updated


deploy/helm/sumologic/values.yaml
```diff
@@ -527,6 +527,14 @@ sumologic:
         annotatedPods:
           enabled: true
 
+        ## By default, the Helm Chart collects some high-cardinality histogram metrics, as Sumo Apps make use of the sum and count components.
+        ## This setting causes the metrics collector to drop the actual histogram buckets, keeping only the sum and the count.
+        ## This affects the following metrics:
+        ## - apiserver_request_duration_seconds
+        ## - coredns_dns_request_duration_seconds
+        ## - kubelet_runtime_operations_duration_seconds
+        dropHistogramBuckets: true
```
rnishtala-sumo, line 536: nice, looks like this should drop the histogram metrics.

rnishtala-sumo:
LGTM!, thanks

rnishtala-sumo (approved):

PR #3203, opened 2023-08-10
feat(metrics/collector) allow setting global scrape interval

swiatekm-sumo:
I've also upgraded the operator Helm Chart to 0.35.0. This is necessary for setting the scrape interval for ServiceMonitors handled by the otel metrics collector.

### Checklist


tests/integration/values/values_helm_ot_metrics.yaml
```diff
@@ -12,6 +12,7 @@ sumologic:
     collector:
       otelcol:
         enabled: true
+        scrapeInterval: 15s
         resources:
```
aboguszewski-sumo, line 16: I think that for tests we might consider even smaller scrape interval, would it be a performance problem?
It's not a blocking thing, besides that LGTM

aboguszewski-sumo (approved):

tests/integration/values/values_helm_ot_metrics.yaml
swiatekm-sumo: We can probably reduce it further, but right now this test is relatively quick anyway, so there isn't much benefit if we need to wait for slower ones anyway. We should also do it for Prometheus tests, which I think should happen in a separate PR.

PR #3211, opened 2023-08-14
feat(prometheus): Removing prometheus recording rule metrics

rnishtala-sumo:
Removing prometheus recording rule metrics

### Checklist

- [x] Documentation updated


deploy/helm/sumologic/values.yaml
```diff
@@ -2210,48 +2018,21 @@ kube-prometheus-stack:
...
-        ## :node_disk_utilisation:avg_irate
-        ## :node_memory_swap_io_bytes:sum_rate
-        ## :node_memory_utilisation:
-        ## :node_net_saturation:sum_irate
-        ## :node_net_utilisation:sum_irate
         ## cluster_quantile:apiserver_request_duration_seconds:histogram_quantile
         ## cluster_quantile:scheduler_binding_duration_seconds:histogram_quantile
         ## cluster_quantile:scheduler_framework_extension_point_duration_seconds:histogram_quantile
         ## cluster_quantile:scheduler_e2e_scheduling_duration_seconds:histogram_quantile
         ## cluster_quantile:scheduler_scheduling_algorithm_duration_seconds:histogram_quantile
```
swiatekm-sumo, line 2025: Can you also disable the default recording rules under `kube-prometheus-stack.defaultRules`? Then we can get rid of this whole remote write definition.

swiatekm-sumo (commented):
LGTM, aside from one note. We should also create the v4 migration doc and have this PR add a note about recording rule removal there.

deploy/helm/sumologic/values.yaml
```diff
@@ -1520,37 +1520,6 @@ kube-prometheus-stack:
...
-      kubeSchedulerRecording: false
-      kubeStateMetrics: false
-      network: false
-      node: true
-      nodeExporterAlerting: false
-      nodeExporterRecording: false
-      prometheus: false
-      prometheusOperator: false
-      windows: false
-
```
swiatekm-sumo, line 1553: These are all enabled by default, so what you need to do here instead is to set everything to `false`.

deploy/helm/sumologic/values.yaml
```diff
@@ -2016,24 +1985,7 @@ kube-prometheus-stack:
             - action: keep
               regex: node-exporter;(?:node_load1|node_load5|node_load15|node_cpu_seconds_total)
               sourceLabels: [job, __name__]
```
swiatekm-sumo, line 1987: Since we're dropping recording rule metrics, we need to forward node metrics which are used to create them, so users can reproduce this in Sumo. After this change, we should check that Prometheus sends these metrics: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/70032194c87cef8bce5aff3eb23d15946433db8a/tests/integration/internal/constants.go#L183 as well.

rnishtala-sumo: hmm ok, I'll add these to the list of node metrics

deploy/helm/sumologic/values.yaml
rnishtala-sumo: set all of them to `false`, thanks

deploy/helm/sumologic/values.yaml
rnishtala-sumo: Screenshot 2023-08-17 at 11 11 26 AM

Looks like the new node metrics are now being ingested with the latest change

swiatekm-sumo: Ah, you also need to add these to the ServiceMonitor definition under `additionalServiceMonitors`.

rnishtala-sumo: ok will take a look. In the meanwhile, the OT metrics test seems to find these unexpected metrics

[kubelet_runtime_operations_duration_seconds_bucket coredns_dns_request_duration_seconds_bucket apiserver_request_duration_seconds_bucket]

rnishtala-sumo: > Ah, you also need to add these to the ServiceMonitor definition under additionalServiceMonitors

why do we need to scrape these node metrics twice?

swiatekm-sumo: > ok will take a look. In the meanwhile, not sure why this IT is failing, those `bucket` metrics aren't being collected or expected
> 
> ```
> Test_Helm_OT_Metrics/metrics/expected_metrics_are_present 2023-08-17T21:28:13Z retry.go:103: WaitUntilExpectedMetricsPresent() returned an error: found the following unexpected metrics: [kubelet_runtime_operations_duration_seconds_bucket apiserver_request_duration_seconds_bucket coredns_dns_request_duration_seconds_bucket]. Sleeping for 3s and will try again.
>     assess_funcs.go:202: 'WaitUntilExpectedMetricsPresent()' unsuccessful after 40 retries
> ```

It's an unrelated failure that #3192 fixes.

> > Ah, you also need to add these to the ServiceMonitor definition under additionalServiceMonitors
> 
> why do we need to scrape these node metrics twice?

You don't, my mistake. I got confused for a second there. You should see the node metrics after adding them to the node remotewrite.

rnishtala-sumo: yes they've been added to the node remotewrite

docs/v4-migration-doc.md
```diff
@@ -66,4 +71,7 @@ require additional action.
 
 ## Full list of changes
 
-:construction:
+- Drop Prometheus recording rule metrics
+
+  OpenTelemetry can't collect Prometheus recording rule metrics. The new version therefore stops collecting recording rule metrics and
+  updates will be made to the Kubernetes App to not show these metrics.
```
swiatekm-sumo, line 77: I think it would help to provide a list of metrics that went away, and a list of Node metrics we collect instead.

rnishtala-sumo: ok I'll mention this under the full list of changes

swiatekm-sumo (approved):
LGTM

docs/v4-migration-doc.md
rnishtala-sumo: its a long list :)
https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e87dfe0a2f256df80e7677c4cd5500ac1b6fb366/docs/v4-migration-doc.md#full-list-of-changes

PR #3221, opened 2023-08-18
feat(metrics/collector): support remote write proxy

swiatekm-sumo:
The remote write proxy was originally disabled for the otel metrics collector. Until we've added the ability to disable persistent connections in the upstream collector, it's still useful for load balancing.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3221.added.txt
+++ b/.changelog/3221.added.txt
@@ -1,0 +1,1 @@
+feat(metrics/collector): support remote write proxy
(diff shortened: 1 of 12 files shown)


sumo-drosiek (approved):
Should we consider the same for logs?

swiatekm-sumo:
> Should we consider the same for logs?

In that case I think it's better to wait for the upstream setting. Remote write proxy is already here, so using it is very little additional effort. This is also a much smaller problem for logs, where we have a lot more collectors than metadata Pods.

PR #3247, opened 2023-09-06
docs: document otlp format for logs

sumo-drosiek:
### Checklist

- [ ] Documentation updated


docs/collecting-container-logs.md
```diff
@@ -214,9 +214,12 @@ Whereas log line 2 will be displayed as JSON:
 }
 ```
 
-> **Warning** Setting the format to `text` has certain consequences for multiline detection. See [here][troubleshooting_text_format] for
+> **Warning** Setting the format to `text` with HTTP source has certain consequences for multiline detection. See [here][troubleshooting_text_format] for
 > more details.
 
+If you want to send metadata along wih log, you have to use reosurce level attributes, because record level attributes are going to be removed before sending log to Sumo.
```
sumo-drosiek, line 220:
```suggestion
If you want to send metadata along wih log, you have to use reosurce level attributes, because record level attributes are going to be removed before sending log to Sumo Logic.
```

docs/collecting-container-logs.md
```diff
@@ -466,6 +469,13 @@ sumologic:
       enabled: false
 ```
 
+### Using OTLP Source
+
+OTLP source resolves some issues of `text` format, which affects HTTP source:
+
+- Multiline logs are respected, it means that every line of the same log is not treated as separate log
+- Timestamp is set correctly to time set by Kubernetes, because it is send separately, which is not possible in HTTP format
```
swiatekm-sumo, line 477:
```suggestion
- Timestamp is set correctly to time set by Kubernetes, as OTLP contains a dedicated timestamp field.
```

docs/collecting-container-logs.md
```diff
@@ -466,6 +469,13 @@ sumologic:
       enabled: false
 ```
 
+### Using OTLP Source
+
+OTLP source resolves some issues of `text` format, which affects HTTP source:
+
+- Multiline logs are respected, it means that every line of the same log is not treated as separate log
```
swiatekm-sumo, line 476:
```suggestion
- Multiline logs are respected, Sumo will not split logs sent to an OTLP source
```

swiatekm-sumo (approved):

PR #3054, opened 2023-05-15
fix: warn about release name length

swiatekm-sumo:
Fixes #3019 

### Checklist

- [x] Changelog updated or skip changelog label added
- [x] Template tests added for new features

PR #3248, opened 2023-09-07
feat!: truncate fullname after 22 characters

### LINKED ISSUE #3057: Set a limit on the release name
The release name is used in resource name and label values, some of which are limited to 63 characters by [Kubernetes itself](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/). Right now, if a name that is too long is provided, this will result in an error - see #3019 for a practical example.

Unfortunately, in practice the max length is determined by the settings. Therefore, limitting this would be a breaking change, so it needs to wait till the next major release. In the meantime, we can emit a warning about the length.

aboguszewski-sumo:
Fixes #3057

The longest name for a statefulset I found was `*-otelcloudwatch-logs-collector`, the suffix here is `30` characters.

According to https://github.com/kubernetes/kubernetes/issues/64023 the statefulset name should be 52 characters or shorter.

I did not manage to fail rendering the template, I tried using both `fail` and `required` function but without success. This is why I'm only truncating the `sumologic.fullname` name.

### Checklist

- [X] Documentation updated


swiatekm-sumo:
I actually think the limit for this wasn't determined by StatefulSets, but by label values. Can you make sure this test: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/55b7ab04c8b06710723d5e0e15e974b312b23138/tests/helm/common_test.go#L124 also enabled otel cloudwatch?

aboguszewski-sumo:
Changes applies. But, please note that because this does not cause template rendering to fail, the test actually does not check anything more.

deploy/helm/sumologic/templates/_helpers/_common.tpl
```diff
@@ -7,14 +7,15 @@ Expand the name of the chart.
 
 {{/*
 Create a default fully qualified app name.
```
swiatekm-sumo, line 9: I think we want to make the same change for `sumologic.name` above.

swiatekm-sumo:
> Changes applies. But, please note that because this does not cause template rendering to fail, the test actually does not check anything more.

The test checks the actual object names and label values.

deploy/helm/sumologic/templates/_helpers/_common.tpl
aboguszewski-sumo: Is it used anywhere? I grepped for that name and it seems that it's not used anywhere

aboguszewski-sumo:
What about this then? https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/55b7ab04c8b06710723d5e0e15e974b312b23138/tests/helm/common_test.go#L197
What should we check here if we're not failing the rendering?

deploy/helm/sumologic/templates/_helpers/_common.tpl
swiatekm-sumo: Huh, we should delete it then.

swiatekm-sumo:
You should also remove the warning I added in #3054. Incidentally, does putting the `fail` in that file not work?

aboguszewski-sumo:
>  Incidentally, does putting the fail in that file not work?

in `_common.tpl`? No, but I only used `helm template` to test that, maybe an install would fail.

aboguszewski-sumo:
Ok, I tested and `fail` works when put in `NOTES.txt`. But we still prefer to truncate, right?

swiatekm-sumo:
Yeah, seems better all around.

aboguszewski-sumo:
Warning removed and tests fixed.

tests/helm/const.go
```diff
@@ -5,14 +5,14 @@ const (
 	yamlDirectory            = "static"
 	chartDirectory           = "../../deploy/helm/sumologic"
 	chartName                = "sumologic"
-	releaseName              = "collection-test"
+	releaseName              = "col-test"
```
swiatekm-sumo, line 8: Why is this change necessary?

aboguszewski-sumo: Else we get such failures: https://github.com/SumoLogic/sumologic-kubernetes-collection/actions/runs/6109112931/job/16579492782
and I found it easier to modify the constant instead of modifying the tests

swiatekm-sumo: But why does this change cause these failures. `collection-test` is less than 22 characters, so it shouldn't make any difference, right?

deploy/helm/sumologic/templates/_helpers/_common.tpl
aboguszewski-sumo: I'll do it in another pr

swiatekm-sumo (approved):

PR #3254, opened 2023-09-08
chore: drop support for v3.2

aboguszewski-sumo:


astencel-sumo (approved):
Approving this PR, but I think we should change the format of this list so that we don't remove the dates when a version goes out of support.

PR #3261, opened 2023-09-13
Remove prometheus from images in documentation

dmolenda-sumo:


swiatekm-sumo (commented):
The component replacing Prometheus should be called "Metrics Collector" instead of "Metadata enrichment" and we should indicate somehow that it's using the Prometheus pull model.

swiatekm-sumo (approved):

astencel-sumo (approved):

PR #3267, opened 2023-09-13
fix: disable keep-alives for internal traffic

swiatekm-sumo:
Disable HTTP keep-alives for network traffic from collectors to metadata enrichment. This should make the Kubernetes iptables-based load balancing work better at the cost of increased network traffic.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3267.fixed.txt
+++ b/.changelog/3267.fixed.txt
@@ -1,0 +1,1 @@
+fix: disable keep-alives for internal traffic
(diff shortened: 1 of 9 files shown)


astencel-sumo (approved):

astencel-sumo:
Note for future us: This change applies to both `logs-collector` and `metrics-collector`. It does not apply to Prometheus as metrics collector, which still needs the remote write proxy for proper load balancing.

PR #3272, opened 2023-09-14
Troubleshoot guide for error when replacing Prometheus with Opentelem…

dmolenda-sumo:
…etry on Vagrant


sumo-drosiek (approved):

icco:
This is great, a similar section for normal upgrades would be useful as I ran into this error while upgrading to v4.

icco:
Helm via terraform

PR #3281, opened 2023-09-19
fix(logs)!: move JSON parsing after user-defined processors

swiatekm-sumo:
It's easier for users to write processors if they can assume the log body is a string, as opposed to having to work with arbitrary structured data.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3281.fixed.txt
+++ b/.changelog/3281.fixed.txt
@@ -1,0 +1,5 @@
+fix(logs)!: move JSON parsing after user-defined processors
+
+The log body will now always be a string if accessed in extra processors.
+Users who want to access specific fields in their parsed JSON log
+should explicitly call ParseJSON in their processor definition.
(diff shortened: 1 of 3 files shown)


sumo-drosiek:
> It's easier for users to write processors if they can assume the log body is a string

Hmm, it is problematic if they would like to operate on the log structure :thinking:

swiatekm-sumo:
> > It's easier for users to write processors if they can assume the log body is a string
> 
> Hmm, it is problematic if they would like to operate on the log structure 🤔

They can explicitly parse it from json themselves in that case. And going the other way is apparently impossible in OTTL, or at least I haven't found a way. All the body manipulation functions assume the body is a string and do nothing otherwise. If you want to do things like masking PII, working with the serialized JSON is much easier.

astencel-sumo:
Isn't this a breaking change?

swiatekm-sumo:
> Isn't this a breaking change?

It's a fix, as the current behaviour is not intended, and the new behaviour will be better for the majority of users. However, some users may depend on the current behaviour. I think the change is beneficial enough in the aggregate that it should go into v3. I've added an extended note to the changelog and marked the change as breaking.

astencel-sumo (approved):

PR #1786, opened 2021-09-27
fix(otelcol): remove k8s.pod.pod_name for metrics pipeline

pmalek-sumo:

PR #3306, opened 2023-09-26
feat!(metrics): move extra processors after sumologic_schema

### LINKED ISSUE #3282: Discrepancy between `extraProcessors` for logs and metrics
When using `sumologic.logs.container.otelcol.extraProcessors`, the attributes that are available are `pod`, `namespace`, `node`, etc.

But when using `sumologic.metrics.otelcol.extraProcessors`, the attributes that are available are `k8s.pod.name`, `k8s.namespace.name`, `k8s.node.name` etc., even though when I look at the metrics in Sumo, I can see `pod`, `namespace`, `node`, etc.

I would expect the behavior of the metrics `extraProcessors` to match that of logs. My expectation is to be able to operate on the fields that I can then see in Sumo.

aboguszewski-sumo:
Fixes #3282 

### Checklist

- [ ] Documentation updated


deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
```diff
@@ -126,10 +126,10 @@ resource/delete_source_metadata:
       key: _sourceHost
     - action: delete
       key: _sourceName
-resource/remove_k8s_pod_pod_name:
+resource/remove_pod_name:
   attributes:
     - action: delete
-      key: k8s.pod.pod_name
+      key: pod
```
aboguszewski-sumo, line 132: Two things:
1. I don't think this worked before, the actual name set by k8sprocessor is `k8s.pod.name`.
2. After moving this, the name to be deleted changes, because sumologic_schema translates it.

aboguszewski-sumo: Integration tests are failing because of this change. Does anyone know what this processor was supposed to do?

aboguszewski-sumo: It seems to be some antique change: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1786

swiatekm-sumo: I think `k8s.pod.pod_name` is added by source processor. Is there a problem moving this processor to right after source processor in the pipeline?

aboguszewski-sumo: Ok, it makes sense now. 
I'd prefer not moving this directly after source processor to make changes in this PR as little intrusive as possible (if I moved this, then, for example, if someone used key `k8s.pod.pod_name` in extra processors, it would break the config). But we can move it.

swiatekm-sumo:
You should include this change in the full list of changes in the migration doc.

docs/v4-migration-doc.md
```diff
@@ -256,8 +256,14 @@ require additional action.
   Some Kubernetes objects, for example statefulsets, have a tight (63 characters) limit for their names. Because of that, we truncate the
   prefix that is attached to the names. In particular, the value under key `fullnameOverride` will be truncated to 22 characters.
 
+- Moving extra processors in metrics pipeline after sumologic_schema processor
+
+  This has been changed in order to make this coherent with pipeline for logs.
```
swiatekm-sumo, line 261:
```suggestion
  This has been changed in order to make the behaviour consistent with the logs pipeline.
```

swiatekm-sumo (approved):

PR #3315, opened 2023-10-06
use autoscaling/v2 if available on the cluster

montmanu:
Similar to https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3155, this change attempts to address https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3154 which still appears to be an issue as of chart v3.15.0 when targeting a 1.26 cluster (i.e., the chart still renders `autoscaling/v2beta2` which was removed in 1.26).

This implementation uses `Capabilities.APIVersions.Has $version` instead of attempting to compare against `Capabilities.KubeVersion.{Major,Minor}` values. This implementation was based on the approach used for `PodDisruptionBudget`: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/a7e8fe99b614f869f22682907d8d856fa6101ffb/deploy/helm/sumologic/templates/logs/common/pdb.yaml#L3

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3315.fixed.txt
+++ b/.changelog/3315.fixed.txt
@@ -1,0 +1,1 @@
+use autoscaling/v2 if available on the cluster
--- a/deploy/helm/sumologic/templates/_helpers/_api_versions.tpl
+++ b/deploy/helm/sumologic/templates/_helpers/_api_versions.tpl
@@ -12,3 +12,18 @@ policy/v1
 policy/v1beta1
 {{- end -}}
 {{- end -}}
+
+{{/*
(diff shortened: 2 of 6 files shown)


sumo-drosiek:
Does it resolve described issue? https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3155#issuecomment-1635635494

I like this approach more, and should be preferred, but I am afraid it won't help

montmanu:
@sumo-drosiek no, it didn't, so I'm going to close this.

PR #3316, opened 2023-10-10
feat(setup): check if pvcs can be created automatically

aboguszewski-sumo:
This PR adds to setup script a functionality to check if pvcs can be created automatically.

### Checklist

- [ ] Documentation updated

### DIFF
--- a/deploy/helm/sumologic/conf/setup/setup.sh
+++ b/deploy/helm/sumologic/conf/setup/setup.sh
@@ -2,6 +2,58 @@
 
 readonly DEBUG_MODE=${DEBUG_MODE:="false"}
 readonly DEBUG_MODE_ENABLED_FLAG="true"
+readonly TEST_PVC_FILE="test-pvc-ss.yaml"
+readonly TEST_PVC_STATEFULSET="pvc-nginx"
+readonly TEST_PVC_NAMESPACE="test-pvc"
+readonly TEST_PVC_STATEFULSET_CONFIG="apiVersion: v1
+kind: Service
+metadata:
+  name: nginx
+  labels:
+    app: nginx
+spec:
+  ports:
+  - port: 80
+    name: web
+  clusterIP: None
+  selector:
+    app: nginx
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: web
+spec:
+  selector:
+    matchLabels:
+      app: nginx
+  serviceName: \"nginx\"
+  replicas: 1
+  minReadySeconds: 10 # by default is 0
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      terminationGracePeriodSeconds: 10
+      containers:
+      - name: nginx
+        image: registry.k8s.io/nginx-slim:0.8
+        ports:
+        - containerPort: 80
+          name: web
+        volumeMounts:
+        - name: www
+          mountPath: /usr/share/nginx/html
+  volumeClaimTemplates:
+  - metadata:
+      name: www
+    spec:
+      accessModes: [ \"ReadWriteOnce\" ]
+      resources:
+        requests:
+          storage: 1Gi
+"
 
 # Let's compare the variables ignoring the case with help of ${VARIABLE,,} which makes the string lowercased
 # so that we don't have to deal with True vs true vs TRUE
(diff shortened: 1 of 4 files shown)


deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -176,3 +272,7 @@ export SUMOLOGIC_ACCESSKEY=
 export SUMOLOGIC_ACCESSID=
 
 bash /etc/terraform/custom.sh
+
+{{- if .Values.sumologic.setup.dashboards.enabled }}
+can_create_pvc
+{{- end }}
```
aboguszewski-sumo, line 278: Maybe because of waiting, this should be run as a separate process in the background.

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -83,6 +135,52 @@ function should_create_fields() {
...
+function pvc_test_cleanup() {
+    kubectl delete all --all -n ${TEST_PVC_NAMESPACE}
+    kubectl delete pvc --all -n ${TEST_PVC_NAMESPACE}
+    rm ${TEST_PVC_FILE}
+}
+
+function can_create_pvc() {
+    pvc_test_create_statefulset
+    sleep 30
+    pvc_test_check
```
aboguszewski-sumo, line 174: Not sure what constant should be used here.

aboguszewski-sumo:
I'll fix helm template tests once `setup.sh` will be stable after the review.

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -2,6 +2,57 @@
 
 readonly DEBUG_MODE=${DEBUG_MODE:="false"}
 readonly DEBUG_MODE_ENABLED_FLAG="true"
+readonly TEST_PVC_FILE="test-pvc-ss.yaml"
+readonly TEST_PVC_STATEFULSET="pvc-nginx"
+readonly TEST_PVC_STATEFULSET_CONFIG="apiVersion: v1
```
sumo-drosiek, line 7: Is it possible to move it to config map?

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -83,6 +134,59 @@ function should_create_fields() {
...
+    fi
+}
+
+function install_kubectl() {
+    curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+    chmod +x kubectl
+    mkdir -p ~/.local/bin
+    mv ./kubectl ~/.local/bin/kubectl
+    export PATH=$PATH:~/.local/bin
+}
```
sumo-drosiek, line 188: It should be added to setup image

deploy/helm/sumologic/conf/setup/setup.sh
sumo-drosiek: do we need to wait 30 seconds? How do we know pvc cannot be created automatically?

deploy/helm/sumologic/conf/setup/setup.sh
aboguszewski-sumo: Most env variables are in `job.yaml`, wouldn't it be better to move it there?

deploy/helm/sumologic/conf/setup/setup.sh
aboguszewski-sumo: Afaik the only way to do it is to try to create a pvc and see if it gets created - and it doesn't happen automatically. Although it's not perfect, I think it's fine as long as we don't get false negatives (which we might, if creating a pvc takes longer than 30 seconds).
We can lower the impact of waiting by moving this to a separate process, run at the beginning of the script.

deploy/helm/sumologic/conf/setup/setup.sh
sumo-drosiek: I was referring to `TEST_PVC_STATEFULSET_CONFIG` :sweat_smile:

aboguszewski-sumo: I'll have to check how to access it in the setup image.

sumo-drosiek: I don't think we need it in image, we can simply use configmap to add it in runtime: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/templates/setup/configmap.yaml

In dockerfile, we only get files required for `terraform init` to have dependencies already in image

swiatekm-sumo (commented):
Is there a reason we don't try to create a PVC directly instead of the StatefulSet?

aboguszewski-sumo:
When reproducing, I wasn't sure if just creating a pvc will imply whether it can or not be creating automatically - with statefulset I had that guarantee, because it is directly the case we have in the Helm Chart.
In general, creating anything manually seemed fishy for me, because in some cases the pvcs seem to be working fine if created manually.

swiatekm-sumo:
> When reproducing, I wasn't sure if just creating a pvc will imply whether it can or not be creating automatically - with statefulset I had that guarantee, because it is directly the case we have in the Helm Chart. In general, creating anything manually seemed fishy for me, because in some cases the pvcs seem to be working fine if created manually.

You've seen a situation where a plain PVC worked, but not one created by a StatefulSet?

aboguszewski-sumo:
@swiatekm-sumo It worked when I created PV manually. I had some problems with creating it automatically on EKS (even after adding the plugin), so I went for a statefulset.

PR #3323, opened 2023-10-13
fix(metrics): kube-state-metrics pod metadata

swiatekm-sumo:
We always drop `container` and `pod` attributes from kube state metrics if their value is the kube-state-metrics pod/container itself. This is usually wrong, as the attributes are added by the ServiceMonitor configuration to metrics which aren't about pods to begin with, like `kube_deployment_status_replicas`, for example.

However, this means we incorrectly drop these attributes from pod metrics about the `kube-state-metrics` pod itself, like `kube_pod_status_phase`.

This PR fixes this bug by checking for the `uid` attribute, which kube-state-metrics adds to all metrics about pods.

### Checklist

### DIFF
--- a/.changelog/3323.fixed.txt
+++ b/.changelog/3323.fixed.txt
@@ -1,0 +1,1 @@
+fix(metrics): kube-state-metrics pod metadata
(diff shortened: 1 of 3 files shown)


sumo-drosiek (approved):
Nice!

PR #3327, opened 2023-10-16
docs: refactor installation guide

swiatekm-sumo:
Made the installation guide more streamlined, mostly by making the critical path as short and straightforward as possible, and moving less used sections to later parts of the document.

Some more specific changes:
* Compressed the ToC into only the top-level headers, as it was too long
* Deleted the Prometheus-related troubleshooting sections, given that Otel is now the default.
* Move some early sections to Troubleshooting, and Troubleshooting to the end
* Added top-level sections for OpenShift and Non-Helm at the end of the document

### Checklist

- [x] Documentation updated


docs/installation.md
```diff
@@ -388,87 +335,87 @@ Get the logs from that pod:
...
+Kubectl.
 
-```
-helm history my-release
-helm rollback my-release <REVISION-NUMBER>
-```
+> **Warning:** > Before upgrade, please delete the old jobs:
 
-## Uninstalling Sumo Logic Collection
+- `kubectl delete job -n ${NAMESPACE} my-release-sumologic-setup`
```
rnishtala-sumo, line 378: why do we need to delete the job explicitly, isn't the job/pod removed once its complete?

docs/installation.md
```diff
@@ -388,87 +335,87 @@ Get the logs from that pod:
...
+- `kubectl delete job -n ${NAMESPACE} my-release-sumologic-ot-operator-instr` (needed only if `opentelemetry-operator.enabled=true`)
```
rnishtala-sumo, line 379: same question here I guess.

docs/installation.md
```diff
@@ -237,15 +83,40 @@ helm upgrade \
...
+More configuration examples (i.e. OpenShift) can found seen in [this document](/docs/configuration-examples.md).
+
+## Viewing Data In Sumo Logic
+
+Once you have completed installation, you can [install the Kubernetes App and view the dashboards][sumo-k8s-app-dashboards] or [open a new
+Explore tab] in Sumo Logic. If you do not see data in Sumo Logic, you can review our [troubleshooting guide](troubleshoot-collection.md).
+
+![dashboard](/images/sumo_k8s_dashboard.png)
+
+[sumo-k8s-app-dashboards]: https://help.sumologic.com/docs/integrations/containers-orchestration/kubernetes#installing-the-kubernetes-app
```
rnishtala-sumo, line 113: nice looking dashboard!

docs/installation.md
```diff
@@ -388,87 +335,87 @@ Get the logs from that pod:
...
+```yaml
+sumologic:
+  scc:
+    create: true
+otellogs:
+  daemonset:
+    containers:
+      otelcol:
+        securityContext:
+          privileged: true
```
rnishtala-sumo, line 355: is this privilege elevation only required in openshift?

docs/installation.md
swiatekm-sumo: It is after some TTL (https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/), but this isn't guaranteed to have passed at this point.

docs/installation.md
swiatekm-sumo: I believe so, but I'm not an expert on OpenShift. @kkujawa-sumo do you know?

rnishtala-sumo (approved):

PR #3329, opened 2023-10-18
feat!: Enable autoscaling by default

dmolenda-sumo:


deploy/helm/sumologic/values.yaml
```diff
@@ -454,7 +458,7 @@ sumologic:
         ## Option to turn autoscaling on for otelcol and specify params for HPA.
         ## Autoscaling needs metrics-server to access cpu metrics.
         autoscaling:
-          enabled: false
+          # enabled: true
```
sumo-drosiek, line 461: shouldn't be `false` as well?

swiatekm-sumo: To the contrary, shouldn't it be true everywhere? That's the actual default after this change.

sumo-drosiek: What's the point of uncommenting it if it's true anyway?
We are using `#` more like examples in other cases in this file

swiatekm-sumo:
Looks good to me, overall. In general, the `basic.input.yaml` files shouldn't have any changes - we should see changes in the output, because autoscaling is now enabled. We should disable autoscaling in `custom.input.yaml` golden files, because those files exists to check the effects of changing defaults. Right now, you have a mix of these.

deploy/helm/sumologic/templates/_helpers/_common.tpl
```diff
@@ -527,3 +527,19 @@ podAntiAffinity:
...
 {{- end -}}
+
+{{/*
+Returns whether autoscaling is enabled. 
+Returns value of received parameter (.autoscalingEnabled) if defined, 
+otherwise returns value of global flag (.Values.sumologic.autoscaling.enabled).
+
+Example usage:
+{{- template "is.autoscaling.enabled" (dict "autoscalingEnabled" .Values.metadata.logs.autoscaling.enabled) }}
+*/}}
```
swiatekm-sumo, line 538: Looking at actual usage examples, this isn't correct. I think it should be
```suggestion
Example usage:
{{- template "is.autoscaling.enabled" (dict "autoscalingEnabled" .Values.metadata.logs.autoscaling.enabled "Values" .Values) }}
*/}}
```

swiatekm-sumo (approved):
LGTM

PR #3343, opened 2023-10-20
chore: use nix for installing tools in Vagrant

swiatekm-sumo:
Use nix for installing tools in the Vagrant box. This gives us:

* consistent versions between the Nix environment and the VM
* single source of truth for said version
* support for ARM, relevant for M1 Macs

You need to do some slight weirdness to force Nix to pull packages during VM provisioning rather than at startup, but otherwise this is fairly straightforward. There's probably a more idiomatic way to do this (install packages from shell.nix into a user environment), but the current way works fine.

### Checklist


dbado-sumologic:
Below is my vagrant.configure section for arm & qemu, which works just fine in this setup :) 

```
Vagrant.configure('2') do |config|
  config.vm.box = "perk/ubuntu-2204-arm64"
  config.disksize.size = '50GB'
  config.vm.box_check_update = false
  config.vm.host_name = 'sumologic-kubernetes-collection'
  # Vbox 6.1.28+ restricts host-only adapters to 192.168.56.0/21
  # See: https://www.virtualbox.org/manual/ch06.html#network_hostonly
  config.vm.network :private_network, ip: "192.168.56.2"

  config.vm.provider "qemu" do |qe|
    qe.gui = false
    qe.smp = 8
    qe.memory = 16384
    qe.name = 'sumologic-kubernetes-collection'
  end
  
  config.vm.provision 'shell', path: 'vagrant/provision.sh'

  config.vm.synced_folder ".", "/sumologic"
  config.vbguest.auto_update = false
end
```

dbado-sumologic (approved):
tested on ARM and works fine

sumo-drosiek (approved):
Works :)

.gitignore
```diff
@@ -25,3 +25,6 @@ tmp
 # Go workspace files
 go.work
 go.work.sum
+
+# Local values files
+values.local.yaml
```
sumo-drosiek, line 30:
```suggestion
values.local.yaml

```

PR #2502, opened 2022-09-05
feat: add dot suffix to internal dns addresses

sumo-drosiek:
Add dot suffix to internal dns queries

ref: https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html

PR #3362, opened 2023-10-26
feat: allow setting the cluster DNS domain

swiatekm-sumo:
It's possible to change the domain from the default `cluster.local` using KOPS or Kubespray, or just kubeadm directly. In that case, our endpoints won't resolve correctly. There's reasons to use the fully qualified names whenever possible, so I'm adding a setting for the cluster DNS domain for users who actually set a custom one.

There's no straightforward way to replace this for Prometheus remote write urls, so I've just removed the suffix completely. Given that Prometheus is slowly on the way out and not the default in v4, this should be fine.

See for reference on why we bother with this: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2502 https://github.com/kubernetes/kubernetes/issues/56903#issuecomment-397404229

### Checklist


deploy/helm/sumologic/templates/_helpers/_metrics.tpl
```diff
@@ -36,10 +36,10 @@ Generate list of remoteWrite endpoints for telegraf configuration
 {{- $endpoints := list -}}
 {{- $kps := get .Values "kube-prometheus-stack" -}}
 {{- range $remoteWrite := $kps.prometheus.prometheusSpec.remoteWrite }}
-{{- $endpoints = append $endpoints ($remoteWrite.url | trimPrefix "http://$(METADATA_METRICS_SVC).$(NAMESPACE).svc.cluster.local.:9888" | quote) -}}
+{{- $endpoints = append $endpoints ($remoteWrite.url | trimPrefix "http://$(METADATA_METRICS_SVC).$(NAMESPACE):9888" | quote) -}}
 {{- end }}
 {{- range $remoteWrite := $kps.prometheus.prometheusSpec.additionalRemoteWrite }}
-{{- $endpoints = append $endpoints ($remoteWrite.url | trimPrefix "http://$(METADATA_METRICS_SVC).$(NAMESPACE).svc.cluster.local.:9888" | quote) -}}
+{{- $endpoints = append $endpoints ($remoteWrite.url | trimPrefix "http://$(METADATA_METRICS_SVC).$(NAMESPACE):9888" | quote) -}}
 {{- end -}}
```
aboguszewski-sumo, line 43: Why in configs you add `.Values.sumologic.clusterDNSDomain` but here you simply remove the suffix?

swiatekm-sumo: I had to remove it in the remoteWrite configuration in values.yaml, so I thought I'd be consistent about it.

deploy/helm/sumologic/README.md
```diff
@@ -142,6 +142,7 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
 | `sumologic.clusterName`                                                               | An identifier for the Kubernetes cluster. Whitespaces in the cluster name will be replaced with dashes.                                                                                                                                                                                                    | `kubernetes`                                                                                                                                                                         |
+| `sumologic.clusterDNSDomain`                                                          | The DNS domain for the cluster. Change this if you're using a custom domain.                                                                                                                                                                                                                               | `cluster.local`                                                                                                                                                                      |
```
astencel-sumo, line 145: Do we need the `sumologic.` prefix to this property?

astencel-sumo (approved):

sumo-drosiek (approved):

deploy/helm/sumologic/README.md
swiatekm-sumo: It's there for consistency. Similar properties like the proxy settings live in this section as well.

PR #3371, opened 2023-11-02
chore: upgrade otel to 0.88.0-sumo-0

swiatekm-sumo:
### Checklist

- [x] Documentation updated


swiatekm-sumo:
The reason these tests fail is that these metrics aren't present. As far as I can tell, this is somehow caused by https://github.com/open-telemetry/opentelemetry-collector/pull/8674. However, they are present for otlp exporter, so I'm assuming the root cause lies in sumologic exporter. I haven't yet checked any other exporters to verify.

astencel-sumo:
> However, they are present for otlp exporter

From my testing, this is not the case. Since `v0.88.0`, the OTLP exporter also only emits the `otelcol_exporter_enqueue_failed_*` metrics if they have non-zero value.

This was discussed in yesterday's OpenTelemetry Collector SIG meeting and the result is this is the intended behavior. I have adjusted our integration tests to not check for the existence of those metrics.

swiatekm-sumo:
I've set the changelog entry to breaking and added a note explaining the change.

astencel-sumo (approved):

PR #3373, opened 2023-11-02
fix(instrumentation): replace tools image with kubectl

swiatekm-sumo:
The kubernetes-tools image isn't supported for production use. The kubectl image with minimal dependencies suffices for applying default instrumentation CRs.

### Checklist


mat-rumian (approved):
LGTM :)

PR #3404, opened 2023-11-20
chore: upgrade OpenTelemetry Operator chart from `v0.42.0` to `v0.42.3`

astencel-sumo:


swiatekm-sumo:
You should modify the existing changelog entry instead of creating a new one imo.

swiatekm-sumo (approved):

PR #3405, opened 2023-11-20
chore: upgrade `metrics-server` subchart from `v6.6.0` to `v6.6.2`

astencel-sumo:


swiatekm-sumo:
You should modify the existing changelog entry instead of creating a new one imo: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/.changelog/3392.changed.txt

swiatekm-sumo (approved):

PR #3410, opened 2023-11-21
docs: add more detailed information about filtering ingested data

aboguszewski-sumo:
I checked for remaining mentions to filtering in other docs.

Also - removed configs for `filterprocessor` were outdated, just saying.

### Checklist

- [X] Documentation updated

### DIFF
--- a/docs/collecting-application-metrics.md
+++ b/docs/collecting-application-metrics.md
@@ -183,42 +183,7 @@ This section coverts the following metrics modifications:
 
 ### Filtering metrics
 
-In order to filter in or out the metrics, you can add [filterprocessor] to metric's extraProcessors. Please see the following example:
-
-```yaml
-sumologic:
-  metrics:
-    otelcol:
-      extraProcessors:
-        - filter/1:
-            metrics:
-              ## Definition of inclusion
-              include:
-                ## Match type, can be regexp or strict
-                match_type: regexp
-                ## metric names to match for inclusion
-                metric_names:
-                  - prefix/.*
-                  - prefix_.*
-                ## Metadata to match for inclusion
-                resource_attributes:
-                  - key: container.name
-                    value: app_container_1
-              ## Definition of exclusion
-              exclude:
-                ## Match type, can be regexp or strict
-                match_type: strict
-                ## Metric names for exclusion
-                metric_names:
-                  - hello_world
-                  - hello/world
-                ## Metadata to match for exclusion
-                resource_attributes:
-                  - key: container.name
-                    value: app_container_7
-```
-
-[filterprocessor]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor
+Please see [the doc about filtering data](/docs/filtering.md#metrics).
 
 #### Default attributes
 
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -1,0 +1,126 @@
+# Filtering
+
+One of the easiest way to lower your ingest is to filter out data you do not need.
+In this guide you will learn how to do this for logs, metrics and their metadata.
+
+- [OpenTelemetry Collector processors](#opentelemetry-collector-processors)
+  - [Filter processor](#filter-processor)
+  - [Transform processor](#transform-processor)
+- [Metadata](#metadata)
+- [Logs](#logs)
+- [Metrics](#metrics)
+
+## OpenTelemetry Collector processors
+
+Basic way to perform any custom filtering is adding additional processors to the pipeline in the metadata layer.
+
+For logs, you can specify a list of config under the following keys:
+
+- `sumologic.logs.container.otelcol.extraProcessors` for container logs
+- `sumologic.logs.systemd.otelcol.extraProcessors` for systemd logs
+- `sumologic.logs.kubelet.otelcol.extraProcessors` for kubelet logs
+
+For example, with this config you can specify two `filter` processors to filter out container logs:
+
+```yaml
+sumologic:
+  logs:
+    container:
+      otelcol:
+        extraProcessors:
+          - filter/drop_something:
+              ## <filter processor config>
+          - filter/drop_something_else:
+              ## <filter processor config>
+```
+
+For metrics, there is a single key `sumologic.metrics.otelcol.extraProcessors`:
+
+```yaml
+sumologic:
+  metrics:
+    otelcol:
+      extraProcessors:
+        - filter/drop_something:
+            ## <filter processor config>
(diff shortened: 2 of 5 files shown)


aboguszewski-sumo:
I've checked and removed all the remaining mentions to filtering outside this doc.

docs/filtering.md
```diff
@@ -0,0 +1,324 @@
...
+
+## OpenTelemetry Collector processors
+
+Basic way to perform any custom filtering is adding additional processors to the pipeline in the metadata layer.
+
+For logs, you can specify a list of config under the following keys:
+
+- `sumologic.logs.container.otelcol.extraProcessors` for container logs
+- `sumologic.logs.systemd.otelcol.extraProcessors` for systemd logs
+- `sumologic.logs.kubelet.otelcol.extraProcessors` for kubelet logs
```
swiatekm-sumo, line 33: We should also mention the metrics setting here.

docs/filtering.md
```diff
@@ -0,0 +1,324 @@
...
+- [Metrics](#metrics)
+  - [Prometheus](#prometheus)
+    - [Filtering Prometheus metrics by namespace](#filtering-prometheus-metrics-by-namespace)
+  - [Filter out app metrics](#filter-out-app-metrics)
+  - [Custom filtering](#custom-filtering-1)
+    - [Drop unnecessary metrics](#drop-unnecessary-metrics)
+
+## OpenTelemetry Collector processors
+
+Basic way to perform any custom filtering is adding additional processors to the pipeline in the metadata layer.
```
swiatekm-sumo, line 27:
```suggestion
This Helm Chart uses the OpenTelemetry Collector for data collection and processing. Filtering is accomplished via appropriate Otel processors.
```
We should link to filter processor and transform processor here as well. I think mentioning pipelines and the metadata layer here is unnecessary, users shouldn't care about that for most of their filtering needs.

docs/collecting-application-metrics.md
```diff
@@ -183,42 +183,7 @@ This section coverts the following metrics modifications:
 
 ### Filtering metrics
 
-In order to filter in or out the metrics, you can add [filterprocessor] to metric's extraProcessors. Please see the following example:
```
swiatekm-sumo, line 186: As a general rule, I'm wondering if it isn't better to keep these examples here and *also* link to the filtering document. That might make it easier to navigate this documentation, but I'm not entirely sure. @astencel-sumo wdyt?

aboguszewski-sumo: The main counterargument for that is maintainability: most of these examples I removed were already outdated. I'm not particularly for or against it, just pointing out.

astencel-sumo: I agree with Adam that the examples should only exist in one place, not two.

I agree with Mikołaj 😄 that I'd rather see metrics filtering described in the "Collecting metrics" document and logs filtering described in the "Collecting logs" document, rather than have a separate "Filtering" document.

swiatekm-sumo: That's also a valid point, though if we have a lot of examples, it'll make that document quite long.

docs/filtering.md
```diff
@@ -0,0 +1,324 @@
...
+      excludeFacilityRegex: "some-log-facility-*"
+      excludeHostRegex: "some-log-host-*"
+      excludePriorityRegex: "some-log-priority-*"
+      excludeUnitRegex: "some-log-unit-*"
+```
+
+### Custom filtering
+
+As [mentioned before](#opentelemetry-collector-processors), you can add custom filtering using OpenTelemetry Collector's processors. Below
+are few common examples. Here we present only configs for container logs, but it works the same way for systemd and kubelet logs.
```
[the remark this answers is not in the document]
aboguszewski-sumo: Any concrete ideas? "All usable examples" is quite a broad term: what are we lacking here? I can come up only with excluding basing on log body, severity and time. (and maybe things like span_id but I think it's less useful)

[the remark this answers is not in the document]
swiatekm-sumo: I just meant that every example you currently have in this document in other sections should be here. For coming up with more, maybe it's worthwhile to read through our past support questions?

docs/collecting-application-metrics.md
aboguszewski-sumo: Do we have a decision here? I'm rather for having this in a separate file.

dbado-sumologic: OK, I see we have it already [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/08caa788ddab44f2c0e9f7cfcb485274bff2473d/docs/collecting-application-metrics.md#filtering-metrics)

docs/filtering.md
```diff
@@ -0,0 +1,361 @@
...
+    - [Drop metric datapoints with unspecified type](#drop-metric-datapoints-with-unspecified-type-1)
+  - [Prometheus](#prometheus)
+    - [Filtering Prometheus metrics by namespace](#filtering-prometheus-metrics-by-namespace)
+
+## OpenTelemetry Collector processors
+
+This Helm Chart uses the OpenTelemetry Collector for data collection and processing. Filtering is accomplished via appropriate Otel
+processors such as [Filter processor](#filter-processor) and [Transform processor](#transform-processor).
+
+For logs, you can specify a list of config under the following keys:
```
swiatekm-sumo, line 36:
```suggestion
For logs, you can specify a list of processor definitions under the following keys:
```

docs/filtering.md
```diff
@@ -0,0 +1,361 @@
...
+
+```yaml
+transform/truncate:
+  log_statements:
+    - context: log
+      statements:
+        - truncate_all(attributes, 4096)
+```
+
+#### Limit to `32` the number of fields sent to Sumo Logic
```
swiatekm-sumo, line 113:
```suggestion
#### Limit the number of fields sent to Sumo Logic to `32` 
```

docs/filtering.md
```diff
@@ -0,0 +1,361 @@
+# Filtering
+
+One of the easiest way to lower your ingest is to filter out data you do not need. In this guide you will learn how to do this for logs,
```
swiatekm-sumo, line 3:
```suggestion
One of the easiest ways to lower your ingest is to filter out data you do not need. In this guide you will learn how to do this for logs,
```

swiatekm-sumo (approved):
If Damian as a relative outsider thinks it's better to have a separate filtering document, then I'm ok with it as well. Added some minor phrasing fixes, but otherwise LGTM.

PR #3415, opened 2023-11-27
deps: update falco to 3.8.5

dmolenda-sumo:
As per [3384](https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3384)


aboguszewski-sumo:
Two changelog files, please change that.

swiatekm-sumo:
> Two changelog files, please change that.

This is intentional, there's one changelog file we're adding in this PR, and one we're changing from a previous update to the Falco Chart. Otherwise we'd have two entries for this in the release changelog, which we don't want.

aboguszewski-sumo (approved):

swiatekm-sumo (approved):

PR #3422, opened 2023-11-29
docs(troubleshooting): combine troubleshooting docs into one

astencel-sumo:
We currently have two different troubleshooting docs:
- General troubleshooting: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.2.0/docs/troubleshoot-collection.md
- Troubleshooting installation: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.2.0/docs/installation.md#troubleshooting-installation

It's sometimes not obvious which doc I should look through for an issue. I usually search both just to be sure.
I believe it would be much better to have all troubleshooting in one doc - user can just "Cmd+F" trough that single doc in search for their error.
I propose that we move the "installation troubleshooting" to the "general troubleshooting" docs.


sumo-drosiek:
Do we have any opinion of Customer Success, Customer Support and Customers regarding separated troubleshooting documents?

swiatekm-sumo (approved):

sumo-drosiek (approved):

PR #3427, opened 2023-12-05
feat(chart): add global nodeSelector option

aboguszewski-sumo:
Quick notes:
- some options were missing tests for nodeSelector (instrumentation/*/deployment.yaml)
- pvcCleaner has a bug (`nodeSelector:` is printed even when disabled)
- `kube-prometheus-stack` has documented entries to change that, but it's a part of a subchart

The above are not fixed yet.

### Checklist

- [X] Documentation updated


swiatekm-sumo (commented):
LGTM overall. General comments:

* You'll need to update the documentation as well.
* Instead of all the goldenfile tests, could we use a normal Go test and just check all the resources which can have a nodeSelector set? Similar to: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/c1964af9c571c35e6a025c6aa8417fb5f2be2433/tests/helm/common_test.go#L81

aboguszewski-sumo:
@swiatekm-sumo which documentation you mean in particular? `best-practices.md`?

swiatekm-sumo:
> @swiatekm-sumo which documentation you mean in particular? `best-practices.md`?

Yes, and anywhere else we instruct users to set nodeSelectors. I think that's the only place though.

docs/best-practices.md
```diff
@@ -840,20 +840,22 @@ pod will end up on specific node according your requirements like operating syst
...
+| `remoteWriteProxy`       | `sumologic.metrics.remoteWriteProxy.nodeSelector.kubernetes.io/os`              |
+| `otelcolInstrumentation` | `otelcolInstrumentation.statefulset.nodeSelector.kubernetes.io/os`              |
+| `tracesGateway`          | `tracesGateway.deployment.nodeSelector.kubernetes.io/os`                        |
+| `otellogs`               | `otellogs.daemonset.nodeSelector.kubernetes.io/os`                              |
+| `setupJob`               | `sumologic.setup.job.nodeSelector.kubernetes.io/os`                             |
+| `tracesSampler`          | `tracesSampler.deployment.nodeSelector.kubernetes.io/os`                        |
+| `metadata`               | `metadata.metrics.statefulset.nodeSelector.kubernetes.io/os`                    |
+| `metadata`               | `metadata.logs.statefulset.nodeSelector.kubernetes.io/os`                       |
+| `pvcCleaner`             | `pvcCleaner.job.nodeSelector.kubernetes.io/os`                                  |
+| `metrics otelcol`        | `sumologic.metrics.collector.otelcol.nodeSelector.kubernetes.io/os`             |
```
swiatekm-sumo, line 858: I think we should split this into two tables:
1. Just the `sumologic.nodeSelector` and subchart options
2. The component specific options

While we're at it, let's also add missing subcharts here: otel-operator, telegraf-operator and falco.

docs/best-practices.md
```diff
@@ -864,6 +866,29 @@ pvcCleaner:
       kubernetes.io/os: linux
 ```
 
+You can also specify a global nodeSelector option that will be used for all components above except for `otellogs`, `kube-prometheus-stack`,
```
swiatekm-sumo, line 869: Why is `otellogs` excluded here?

docs/best-practices.md
```diff
@@ -864,6 +866,29 @@ pvcCleaner:
...
+      nodeSelector:
+        kubernetes.io/os: metricOS
+```
+
+The above config will set:
+
+- `nodeSelector.kubernetes.io/os` to `metricOS` for metadata metrics statefulset
+- nothing for `otellogs`, `kube-prometheus-stack`, `kube-state-metrics` and `prometheus`
+- `nodeSelector.kubernetes.io/os` to `linux` for other components
+
```
swiatekm-sumo, line 891: This whole section should be above the table imo.

docs/best-practices.md
aboguszewski-sumo: Grep did not find the key for me, and I saw that in readme we had only `otellogs.daemonset` so I thought we just copy it.
I looked deeper into it and `nodeSelector` is copied a bit differently there:

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/c1964af9c571c35e6a025c6aa8417fb5f2be2433/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml#L52

I'm not sure what it means, could you explain?

docs/best-practices.md
aboguszewski-sumo: note, we don't have `telegraf-operator` listed [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/main/deploy/helm/sumologic#configuration), I'll try to drop another pr for that late

swiatekm-sumo (approved):

PR #3428, opened 2023-12-06
fix(metrics): use `sumologic.metrics.excludeNamespaceRegex` instead of `sumologic.logs.container.excludeNamespaceRegex`

astencel-sumo:
The property `sumologic.logs.container.excludeNamespaceRegex` was being used to filter out not only container logs, but also metrics. This was incorrect.

I have introduced a new property `sumologic.metrics.excludeNamespaceRegex` that can now be used to filter out metrics based on namespaces.


sumo-drosiek (approved):
LGTM, missing changelog

deploy/helm/sumologic/templates/_helpers/_metrics.tpl
```diff
@@ -309,3 +309,22 @@ Example Usage:
...
+{{/*
+Returns list of namespaces to exclude
+
+Example:
+
+{{ include "metrics.excludeNamespaces" . }}
+*/}}
+{{- define "metrics.excludeNamespaces" -}}
+{{- $excludeNamespaceRegex := .Values.sumologic.metrics.excludeNamespaceRegex | quote -}}
+{{- if eq .Values.sumologic.collectionMonitoring false -}}
```
rnishtala-sumo, line 322: why do we have this condition on `collectionMonitoring`

astencel-sumo: Yeah this is a good point. We'd need to do this for both logs and metrics.

tests/helm/testdata/goldenfile/metadata_metrics_otc/exclude-namespace.output.yaml
```diff
@@ -0,0 +1,201 @@
+---
```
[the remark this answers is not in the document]
astencel-sumo: Good point, let me change that.

swiatekm-sumo (approved):

PR #3490, opened 2024-01-10
feat: add support for k8s 1.29

aboguszewski-sumo:
Some `coredns_forward` metrics have been deprecated: 
https://github.com/dioad/coredns/blob/1918a715988f40a9d31fb386ecf814832e62cbc2/plugin/forward/README.md?plain=1#L114

so the tests did not pass on k8s 1.29

### Checklist

- [X] Documentation updated
- [X] Integration tests added or modified for major features


tests/integration/internal/constants.go
```diff
@@ -324,6 +328,22 @@ var (
 		"scrape_samples_scraped",
 		"scrape_duration_seconds",
 	}
+
+	// Some metrics might change over k8s versions
+	// It is assumed that
```
swiatekm-sumo, line 333: What is assumed? :smile:

aboguszewski-sumo: This PR is WIP, sorry 😞

tests/integration/internal/constants.go
```diff
@@ -392,3 +412,26 @@ func InitializeConstants() error {
...
+	require.NoError(t, err)
+	return v
+}
+
+func GetVersionDependentMetrics(t *testing.T) []string {
+	res := []string{}
+	currVersion := getKubernetesVersion(t)
+
+	for version, ms := range versionDependentMetrics {
+		if currVersion < version {
```
swiatekm-sumo, line 429: This is a string comparison, right? If we want to compare this way, we should use a more structured data type for this.

swiatekm-sumo: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version looks like a good choice.

swiatekm-sumo (approved):

PR #3523, opened 2024-02-05
fix(logs): fix global tolerations for otellogs daemonset

aboguszewski-sumo:
### Checklist

- [ ] Documentation updated


swiatekm-sumo (approved):

deploy/helm/sumologic/templates/_helpers/_logs.tpl
```diff
@@ -320,6 +320,14 @@ Return the otelcol log collector image
 {{ template "utils.getOtelImage" (dict "overrideImage" .Values.otellogs.image "defaultImage" .Values.sumologic.otelcolImage) }}
 {{- end -}}
 
+{{- define "sumologic.logs.collector.tolerations" -}}
+{{- if .Values.otellogs.daemonset.tolerations  -}}
```
rnishtala-sumo, line 324: For helm chart 4.4.0, what's the best way to set this, through the global toleration option as in this PR?
https://github.com/Sanyaku/collections-gitops/pull/1239/files

rnishtala-sumo: Will the toleration be set in the daemonset, if we add that under `otellogs.daemonset.additionalDaemonSets` in `v4.4.0`?

aboguszewski-sumo: I believe so: if the affinity works then toleration should work as well.

rnishtala-sumo: ok thanks I'll update the other ops PR to remove the global toleration option

rnishtala-sumo: Last question
> so every daemonset will have tolerations set to NoSchedule
will the change have any impact on zaidan on the next upgrade?

rnishtala-sumo: also from here, it currently looks like the daemonset either uses a global or default toleration

https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3459/files#diff-3572015c2b36610beabda893073f8744fdcf5576fe7a0544c22d04b38b456ff5R76

aboguszewski-sumo: If there would be impact, then we already did it (as we have seen recently). From k8s perspective, this change brings everything back to pre-`v4.4.0`

rnishtala-sumo (approved):

PR #3528, opened 2024-02-05
feat(metrics): allow customizing kubelet metrics

swiatekm-sumo:
### Checklist

- [ ] Documentation updated


sumo-drosiek:
Maybe would it be nicer to have list of regexes and conacatenate them with `|`. WDYT?

swiatekm-sumo:
> Maybe would it be nicer to have list of regexes and conacatenate them with `|`. WDYT?

That might be more readable, but a single regex is more explicit. And you can just copy paste a custom regex from your ServiceMonitor config and have it work here.

sumo-drosiek (approved):

PR #3533, opened 2024-02-07
feat(sumologic-mock): use shorter name for serviceaccount

sumo-drosiek:
Add support for setup job by using helm-hooks. This is to run Sumo Logic Mock before setup job, so it can use it as terraform api endpoint

I don't like when I start doing something and then I see that there are some fixes to do. So ended up with 3 changelog entries

### Checklist

- [ ] Documentation updated


deploy/helm/sumologic/templates/_helpers/_common.tpl
```diff
@@ -608,5 +616,5 @@ Example usage:
 {{- end -}}
 
 {{- define "sumologic.sumologic-mock.name.roles.serviceaccount" -}}
-{{- template "sumologic.fullname" . }}
+{{- template "sumologic.fullname" . }}-sumologic-mock
```
astencel-sumo, line 619: I think we already know this is a "Sumo Logic" mock. Also we have had issues with names being too long.
```suggestion
{{- template "sumologic.fullname" . }}-mock
```

astencel-sumo: Oh I see the name `...-sumologic-mock` is already being used in `_sumologic_mock.tpl`, I'm apparently a bit late with this suggestion. Can we still change those names to be shorter?

sumo-drosiek: sure, I will prepare it as separate PR to merge before this one

sumo-drosiek: #3536

deploy/helm/sumologic/README.md
```diff
@@ -502,6 +502,7 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
 | `debug.sumologicMock.image`                                                           | Docker image for the Sumo Logic Mock.                                                                                                                                                                                                                                                                      | `{"repository": "public.ecr.aws/sumologic/sumologic-mock", "tag": "2.22.0-59-g245ae92", "pullPolicy": "IfNotPresent"}`                                                               |
+| `debug.enableLocalMode`                                                               | Use it along with `debug.sumologicMock.enabled` in order to install collection locally without accessId, accessKey                                                                                                                                                                                         | `false`                                                                                                                                                                              |
```
aboguszewski-sumo, line 505: Will we want to add some options to configure the local mode? If so, then I'd make a key `debug.localMode.enable` instead.

sumo-drosiek: I do not plan any for now. For me it is just using mock for registration and collector creation process. Do you have any idea what could be change? Also as experimental and debugging feature we can change it any time :sweat:

astencel-sumo: Ideas to customize in the future:
- only send specific signal to mock (logs/metrics/traces)
- only send to mock vs. send to both mock and Sumo

tests/helm/testdata/goldenfile/cleanup/sumologic-mock-no-secret.output.yaml
```diff
@@ -51,7 +51,7 @@ spec:
                 name: RELEASE-NAME-sumologic-setup
           env:
             - name: SUMOLOGIC_BASE_URL
-              value: http://RELEASE-NAME-sumologic-sumologic-mock.sumologic:3000/terraform/api/
+              value: http://RELEASE-NAME-sumologic-mock.sumologic:3000/terraform/api/
```
sumo-drosiek, line 54: I have no idea what happened here

astencel-sumo (approved):

deploy/helm/sumologic/README.md
sumo-drosiek: > only send to mock vs. send to both mock and Sumo

This is supported right now

> only send specific signal to mock (logs/metrics/traces)

This also (in scenario when both Sumo Logic and Mock are being used)

PR #3539, opened 2024-02-09
fix(metrics): image pull secrets for metrics collector

### LINKED ISSUE #3530: `sumologic.pullSecrets` doesn't work for the Target Allocator
We aren't attaching the pull secrets to the target allocator ServiceAccount. As such, it's impossible to use a private image registry for it.

swiatekm-sumo:
Ensure globally defined pull secrets are added to metrics collector ServiceAccounts. I've also added a test that checks all ServiceAccounts to ensure we catch this problem more easily in the future.

Fixes #3530 

### Checklist

- [ ] Documentation updated

### DIFF
--- a/.changelog/3539.fixed.txt
+++ b/.changelog/3539.fixed.txt
@@ -1,0 +1,1 @@
+fix(metrics): image pull secrets for metrics collector
--- a/deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml
+++ b/deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml
@@ -36,6 +36,7 @@ spec:
     filterStrategy: relabel-config
     prometheusCR:
       enabled: true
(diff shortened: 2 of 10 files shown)


sumo-drosiek (approved):

sumo-drosiek (changes requested):
Do we need another serviceaccount for metrics. Can't we reuse the regular one? At this moment we have three serviceaccount:
* setup
* cleanup
* for installed chart

swiatekm-sumo:
> Do we need another serviceaccount for metrics. Can't we reuse the regular one? At this moment we have three serviceaccount:
> 
>     * setup
> 
>     * cleanup
> 
>     * for installed chart

The metrics collector and target allocator have different required RBAC permissions, so they should have separate ServiceAccounts imo, based on the principle of least privilege. I think the same thing is true about setup and cleanup, which need to modify Secrets, unlike everything else.

sumo-drosiek (approved):

Issue #3503, opened 2024-01-16
use debug exporter in place of deprecated logging exporter


sumo-drosiek:
`logging` exporter is deprecated: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/loggingexporter/README.md

_Originally posted by @aboguszewski-sumo in https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3501#issuecomment-1893876828_

PR #3552, opened 2024-02-16
docs: replace Logging exporter with Debug exporter

astencel-sumo:
Fixes https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3503

### DIFF
--- a/deploy/helm/sumologic/conf/instrumentation/traces.sampler.conf.yaml
+++ b/deploy/helm/sumologic/conf/instrumentation/traces.sampler.conf.yaml
@@ -15,12 +15,6 @@ exporters:
     compression: gzip
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


deploy/helm/sumologic/conf/instrumentation/traces.sampler.conf.yaml
```diff
@@ -15,12 +15,6 @@ exporters:
     compression: gzip
 {{- end }}
 
-  ## Following generates verbose logs with span content, useful to verify what
```
astencel-sumo, line 18: I have removed this comment as it is not accessible to users anymore (it was before when it was in `values.yaml`).

swiatekm-sumo (approved):

PR #3558, opened 2024-02-20
ci: add date to dependencies issue title

swiatekm-sumo:
See: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3559

### DIFF
--- a/.github/workflows/check_dependencies.yaml
+++ b/.github/workflows/check_dependencies.yaml
@@ -22,7 +22,7 @@ jobs:
           python -m pip install -r ci/check_dependencies/requirements.txt
       - name: Get issue title
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/check_dependencies.yaml
```diff
@@ -3,6 +3,7 @@ name: Check dependencies
 on:
   schedule:
     - cron: "0 9 * * 1"
+  workflow_dispatch:
```
astencel-sumo, line 6: This allows us to trigger the workflow manually, correct?

astencel-sumo (approved):

.github/workflows/check_dependencies.yaml
swiatekm-sumo: Yes, it's useful for testing and doesn't really hurt, so I left it in.

PR #3561, opened 2024-02-21
feat(chart): drop kube-prometheus-stack

swiatekm-sumo:
This is a draft showing what getting rid of kube-prometheus-stack would look like in practice. Integration tests should pass, template tests should not, as updating them would pollute the diff too much.

Changes:

1. Add `kube-state-metrics` and `prometheus-node-exporter` charts as direct dependencies. I copied the configuration, as I haven't found anyway to avoid a breaking change here.
2. Add ServiceMonitor and PodMonitor CRDs to the Chart. In the absence of kube-prometheus-stack, we need to ship these ourselves.
3. Move the ServiceMonitors for K8s components from kube-prometheus-stack to our Chart. I've more or less copied the `values.yaml` sections, but I'm not sure we want to make this fully configurable.
4. Drop kube-prometheus-stack.
5. Some integration test adjustments.


deploy/helm/sumologic/templates/_helpers/_common.tpl
```diff
@@ -311,6 +311,7 @@ If dryRun=true, we do not create any chart labels.
 chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
 release: "{{ .Release.Name }}"
 heritage: "{{ .Release.Service }}"
+app.kubernetes.io/instance: "{{ .Release.Name }}"
```
swiatekm-sumo, line 314: This was added because kube-state-metrics and node-exporter charts don't add the `release` label to their ServiceMonitors, and I needed a label to select for. In the near future, this shouldn't be necessary, as otel operator will expose more powerful selectors for this.

swiatekm-sumo:
Closing this, but it can still be used as a reference for the implementation.

PR #3571, opened 2024-02-23
chore(instrumentation): remove deprecated otelcol service

mat-rumian:
PR removes `release_name-sumologic-otelcol` service which was deprecated over year ago.

### Checklist

- [ ] Documentation updated


sumo-drosiek (approved):
> Service will point to otelcolInstrumentation as deprecated and after 2023-01-15 should be removed.

:joy:

swiatekm-sumo (changes requested):
Can't this wait until v5? If there's no hurry, I'd rather make this kind of breaking change in a major release, even if it was deprecated for a long time.

mat-rumian:
> Can't this wait until v5? If there's no hurry, I'd rather make this kind of breaking change in a major release, even if it was deprecated for a long time.

Sorry just merged before your comment. I don't think it will be a pain for customers or us. It's over 1,5y since the service was deprecated. If you think that it will cause a lot of distractions I can revert and hold with this until v5 release.

swiatekm-sumo:
I'd rather revert and wait until v5. In my experience, the fact that it was deprecated won't stop anyone from continuing to use it as long as it works.

PR #3600, opened 2024-03-14
feat: Check if remote write definitions are present when Prometheus i…

dmolenda-sumo:
…s not enabled


sumo-drosiek (approved):

swiatekm-sumo (changes requested):
Can we add a Go test for this as well, checking all the possible config value combinations?

dmolenda-sumo:
> Can we add a Go test for this as well, checking all the possible config value combinations?

If there is any other combination that should be tested, please let me know

swiatekm-sumo (approved):

PR #3612, opened 2024-03-28
chore: upgrade tools

swiatekm-sumo:
Upgrate tools to whatever is in nixpkgs unstable today. Tested all the make targets.

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.pull_request.labels.*.name, 'skip-changelog')"
     env:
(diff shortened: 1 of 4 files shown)


.github/workflows/changelog.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.pull_request.labels.*.name, 'skip-changelog')"
     env:
-      TOWNCRIER_VERSION: 22.12.0
+      TOWNCRIER_VERSION: 23.11.0
```
aboguszewski-sumo, line 17: This tricked my vision, because the minor version changed from 12 to 11, I thought it is a downgrade.

aboguszewski-sumo (approved):

PR #3634, opened 2024-04-03
feat: add pvcCleaner for events

sumo-drosiek:
### Checklist

- [ ] Documentation updated


aboguszewski-sumo (approved):

deploy/helm/sumologic/templates/pvc-cleaner/cron-job-events.yaml
```diff
@@ -0,0 +1,71 @@
...
+{{ toYaml . | indent 12 }}
+{{- end }}
+          containers:
+          - name: pvc-cleaner
+            image: {{ .Values.pvcCleaner.job.image.repository }}:{{ .Values.pvcCleaner.job.image.tag }}
+            command:
+            - "bash"
+            - "/pvc-cleaner/pvc-cleaner.sh"
+            - "{{ template "sumologic.namespace" . }}"
+            - "app={{ template "sumologic.labels.app.events.statefulset" . }}"
```
swiatekm-sumo, line 57: I don't like adding a whole separate CronJob manifest just to change this one parameter. Can we use a tpl function instead?

sumo-drosiek: Actually events differs more, as they do not have hpa

```
diff --git a/deploy/helm/sumologic/templates/pvc-cleaner/cron-job-metrics.yaml b/deploy/helm/sumologic/templates/pvc-cleaner/cron-job-events.yaml
index 4b37d360..6e929967 100644
--- a/deploy/helm/sumologic/templates/pvc-cleaner/cron-job-metrics.yaml
+++ b/deploy/helm/sumologic/templates/pvc-cleaner/cron-job-events.yaml
@@ -1,11 +1,11 @@
-{{- if eq .Values.pvcCleaner.metrics.enabled true }}
+{{- if eq .Values.pvcCleaner.events.enabled true }}
 apiVersion: batch/v1
 kind: CronJob
 metadata:
-  name: {{ template "sumologic.metadata.name.pvcCleaner.metrics" . }}
+  name: {{ template "sumologic.metadata.name.pvcCleaner.events" . }}
   namespace: {{ template "sumologic.namespace"  . }}
   labels:
-    app: {{ template "sumologic.labels.app.pvcCleaner.metrics" . }}
+    app: {{ template "sumologic.labels.app.pvcCleaner.events" . }}
 spec:
   schedule: {{ .Values.pvcCleaner.job.schedule | quote }}
@@ -13,9 +13,9 @@ spec:
     spec:
       template:
         metadata:
-          name: {{ template "sumologic.metadata.name.pvcCleaner.metrics" . }}
+          name: {{ template "sumologic.metadata.name.pvcCleaner.events" . }}
           labels:
-            app: {{ template "sumologic.labels.app.pvcCleaner.metrics" . }}
+            app: {{ template "sumologic.labels.app.pvcCleaner.events" . }}
 {{- include "sumologic.labels.common" . | nindent 12 }}
 {{- with .Values.sumologic.podLabels }}
 {{ toYaml . | indent 12 }}
@@ -54,8 +54,7 @@ spec:
             - "bash"
             - "/pvc-cleaner/pvc-cleaner.sh"
             - "{{ template "sumologic.namespace" . }}"
-            - "app={{ template "sumologic.labels.app.metrics.statefulset" . }}"
-            - "{{ template "sumologic.metadata.name.metrics.hpa" . }}"
+            - "app={{ template "sumologic.labels.app.events.statefulset" . }}"
             imagePullPolicy: {{ .Values.pvcCleaner.job.image.pullPolicy }}
             resources:
               {{- toYaml .Values.pvcCleaner.job.resources | nindent 14 }}
```

aboguszewski-sumo: Worth mentioning that we already have separate manifests for metrics and logs: https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/main/deploy/helm/sumologic/templates/pvc-cleaner

swiatekm-sumo: I know, and I think 3 copies is enough to start making them DRY. It'd be best if we could have just one, with different parameters depending on config, but I'm not sure how realistic that is.

sumo-drosiek: I can create issue for that. Right now I'm on customer support, so I would like to have it done rather quickly with low effort

sumo-drosiek: We will have to do it anyway, as retention policy do not secure us if the pod will be removed and recreated in diffetent AZ (original reason for this feature)

> Bear in mind that these policies only apply when Pods are being removed due to the StatefulSet being deleted or scaled down. For example, if a Pod associated with a StatefulSet fails due to node failure, and the control plane creates a replacement Pod, the StatefulSet retains the existing PVC. The existing volume is unaffected, and the cluster will attach it to the node where the new Pod is about to launch.

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention

sumo-drosiek: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3635

swiatekm-sumo: Is anyone urgently asking for this change? Unlike for logs and metrics, you only ever have one events volume, so this would only come into play if the user disables event collection, but keeps the chart installed. Right?

sumo-drosiek: Not really sure is it the only scenario. It's possible that due to node draining/autoscalling/crash the pod may end in different az than the pvc

swiatekm-sumo: If that happens, the Pod will fail to start in the new AZ, because the PVC name is the same, so the previous PVC would need to be manually deleted anyway.

swiatekm-sumo: pvcCleaner was originally for cleaning up unused PVCs after metadata downscaling. It was definitely not for fixing AZ-related volume mounting problems. And it doesn't run frequently enough by default to be able to do this effectively anyway.

sumo-drosiek: Actually I disagree, the original problem was AZ problem AFAIR, but it was result of non-cleaning unused PVCs. I think we can discuss it internally 😅

sumo-drosiek:
Closing in favor of #3639

deploy/helm/sumologic/templates/pvc-cleaner/cron-job-events.yaml
sumo-drosiek: Outcome after tests is that this PR doesn't resolve the issue. Closed in favor of  #3639

PR #3643, opened 2024-04-11
chore: add instrumentation tests

mat-rumian:
PR adds tests for instrumentation applied by OpenTelemetry-Operator.

### Checklist

- [ ] Documentation updated


tests/integration/internal/stepfuncs/assess_funcs.go
```diff
@@ -89,6 +104,7 @@ func WaitUntilExpectedTracesPresent(
 	tickDuration time.Duration,
 ) features.Func {
 	return func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+		fmt.Printf("%s\n", ctx)
```
kkujawa-sumo, line 107: Is it added intentionally?

mat-rumian: Nah, just for investigation purposes. Need to clean this :)

kkujawa-sumo (approved):

tests/integration/helm_opentelemetry_operator_instr_test.go
```diff
@@ -0,0 +1,144 @@
...
+		CheckTracesWithoutGatewayInstall,
+	}
+
+	featInstall := GetInstallFeature(installChecks)
+
+	// It is required to add v1alpha1 OT Operator Scheme to K8s Scheme
+	// https://github.com/open-telemetry/opentelemetry-operator/issues/772
+	if err := otoperatorappsv1.AddToScheme(scheme.Scheme); err != nil {
+		require.Fail(t, "failed to register scheme: %v", err)
+	}
```
swiatekm-sumo, line 49: Might be better to move this to an `init` function. I think we want all of our tests to be able to use the operator CRDs. We should probably do the same for all of our CRDs, though that can happen in a separate PR.

mat-rumian: I will perform that change in separated PR. Is it fine for you?

swiatekm-sumo (approved):
Overall this looks good to me, thanks for doing it!

One thing I'm wondering is if we shouldn't have self-contained tests per-instrumentation. Right now, you have one feature per instrumented application, and then a single feature that triggers all of them via curl and then checks if the data arrived. It would be more composable if each per-app feature also included the curl trigger and the check. Then it'd be possible to run them individually, and easier to keep track of what data we expect the instrumentation to output.

But that can be done in a followup, if need be. I'll leave it up to you, I'm fine merging this PR as is.

PR #3677, opened 2024-04-24
feat: use Sumo Logic repository for all images 

sumo-drosiek:
### Checklist

- [ ] Documentation updated


.changelog/3677.changed.txt
```diff
@@ -0,0 +1 @@
+feat: use Sumo Logic repository for all images
```
swiatekm-sumo, line 1: We should probably have a call out of this change in the release notes. It's not breaking, but it might be surprising for users who rehost the images, have security scanner exceptions set for specific repositories, and so on.

deploy/helm/sumologic/values.yaml
```diff
@@ -2534,6 +2545,10 @@ opentelemetry-operator:
     image:
       repository: public.ecr.aws/sumologic/kube-rbac-proxy
 
+  testFramework:
+    image:
+      repository: public.ecr.aws/sumologic/busybox
```
swiatekm-sumo, line 2550: What is this?

sumo-drosiek: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/tests/test-service-connection.yaml

swiatekm-sumo: I thought that was only used for tests. Can it actually appear if we install the helm chart normally?

sumo-drosiek: It's generated by `helm template` for the following configuration: https://github.com/SumoLogic/sumologic-openshift-images/blob/0096d8236b323ef3827d2fd91b18c41274560a14/scripts/values.yaml

sumo-drosiek: It shouldn't for `helm install`:
```
  annotations:
    "helm.sh/hook": test
```

swiatekm-sumo: I don't think it really hurts to include it, but it might cause some confusion for users. Can we ignore it?

sumo-drosiek: I would leave it for consistency. Many customers do not uses helm directly (eg. argocd), so I would like to be explicit with all the images

.changelog/3677.changed.txt
sumo-drosiek: I will add it to changelog, as I'm going to relase it 😄

PR #3705, opened 2024-05-14
feat(compression): Make the compression type configurable, default: zstd

rnishtala-sumo:
- Make the compression type configurable, default: zstd
- Regenerated golden files

### Checklist


echlebek (approved):

sumo-drosiek:
Does it need to be configurable? Why customers would need to change it? What are possible values?

deploy/helm/sumologic/README.md
```diff
@@ -432,6 +432,7 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
 | `otellogswindows.additionalDaemonSets`                                                | OpenTelemetry Logs Collector for Windows Nodes Daemonset per node customization options. See [Best Practices](https://help.sumologic.com/docs/send-data/kubernetes/best-practices/#setting-different-resources-on-different-nodes-for-logs-collector).                                                     | `{}`                                                                                                                                                                                 |
+| `metadata.compression`                                                                | Compression for logs, metrics and events                                                                                                                                                                                                                                                                   | `zstd`                                                                                                                                                                               |
```
swiatekm-sumo, line 435: How about putting this at the top-level, just `compression`? Right now, it affects the log collector and the event collector, so not just metadata. But it doesn't affect the metrics collector.

swiatekm-sumo (commented):
Making it configurable makes sense in case there's a problem with it, and we need to switch back. But in that case, it makes more sense to keep the default as gzip, and first dogfood zstd properly.

deploy/helm/sumologic/README.md
rnishtala-sumo: yeah, that's a good point, will make this change.

rnishtala-sumo:
> Does it need to be configurable? Why customers would need to change it? What are possible values?

We want to be able to switch between `gzip` and `zstd`.

sumo-drosiek:
> We want to be able to switch between gzip and zstd.

yeah, but does customer need to know what compression do we use?

rnishtala-sumo:
> We can use merge configuration for development/tests and not sure if we need to expose compression to users

I can see some reasons we may want to expose `compression` as a config option:

- Since we do expose `compression` as a config option in the sumologic exporter, if a user wants to set that, this option in the helm chart would allow it.
- We may want to change it across our dogfooding environments for multiple signals

Looks like reducing egress/ingress costs is driving the push to `zstd`, the customer may want the same to save on cost. 

Is there a negative in exposing compression to users? We could use config `merge` but if we're going to do this it may be error prone and will require multiple changes across logs/metrics and events

deploy/helm/sumologic/README.md
rnishtala-sumo: Moved `compression` to the top level. Also added the config option to the otlphttp exporter of the metrics collector.

deploy/helm/sumologic/values.yaml
```diff
@@ -6,6 +6,9 @@ fullnameOverride: ""
 ## Use the same namespace as namespaceOverride in 'kube-prometheus-stack.namespaceOverride' if Prometheus setup is also enabled
 namespaceOverride: ""
 
+## Compression for logs, metrics and events:
```
swiatekm-sumo, line 9: Could you enumerate what the valid values for this field are, both here and in the README?

sumo-drosiek:
> Since we do expose compression as a config option in the sumologic exporter, if a user wants to set that, this option in the helm chart would allow it.
> Is there a negative in exposing compression to users? 

We don't want to expose everything. This is why we stop exposing full configuration. It is easy to expose something but it's hard to revert that change later. I would be rather restrictive in exposing options which are not features or aren't related to resolve issues. Compression is strictly technical topic, and I would be very careful with exposing such options

> We may want to change it across our dogfooding environments for multiple signals

It's development and as I understand, we want to change it to zstd, so we can test the stability of it and then use as defaults in the feature?

> Looks like reducing egress/ingress costs is driving the push to zstd, the customer may want the same to save on cost.

Yes, so if zstd is stable, we may just change it in helm chart and release it. Customers don't need to know that we had `gzip` and now we have `zstd`.

---

Technically PR looks ok, but I'm not convinced about the change itself 🙈

rnishtala-sumo:
> Compression is strictly technical topic, and I would be very careful with exposing such options. We can always expose it in the future if there will be business need for that

My understanding is that this is driven by business need. Also, other projects like nginx do make compression configurable

> Yes, so if zstd is stable, we may just change it in helm chart and release it. Customers don't need to know that we had gzip and now we have zstd

I think all compression options currently have the same stability.

deploy/helm/sumologic/values.yaml
rnishtala-sumo: I've added these options - `Compression types can be gzip, snappy, zstd or deflate`

rnishtala-sumo: It looks like it could also take `""` or `none` values, but didn't want to call these out, since we always want this enabled.

rnishtala-sumo: this is one thing I'm concerned with exposing compression configuration, it allows the customer to disable it.

echlebek:
I do think Dominik has a good point here, and if we know that zstd is the best overall, then maybe it shouldn't be configurable.

Also, I think something to consider is that a potentially more impactful configuration feature in terms of resource consumption could be the compression level. That's what is going to influence CPU use more than algorithm choice for encoding,  if I'm not mistaken.

rnishtala-sumo:
ok, I'll dogfood `zstd` using config merge on k8s collection as the next step. If there aren't issues, will consider switching over to `zstd`.

rnishtala-sumo:
Raised this PR - https://github.com/Sanyaku/collections-gitops/pull/1330

PR #3711, opened 2024-05-17
feat: add extraExporters to logs

sumo-drosiek:
### Sending Data to Multiple Targets

It is possible to send logs to multiple locations. This section describes the following cases:

- send logs simultaneously to Sumo Logic and other locations
- send all logs to Sumo Logic and part of it to other locations
- send logs selectively to multiple locations

## Send Logs Simultaneously to Sumo Logic and Other Locations

In order to send data to Sumo Logic and other locations, `sumologic.logs.otelcol.extraExporters` should be used.

With the following configuration data will be send to both Sumo Logic and AWS S3:

```yaml
sumologic:
  logs:
    otelcol:
      extraExporters:
        ## define `awss3/production` exporter
        ## as no routing is defined, all data will go to this expoter in addition to the default one
        awss3:
          s3uploader:
              region: 'eu-central-1'
              s3_bucket: 'databucket'
              s3_prefix: 'metric'
              s3_partition: 'minute'
metadata:
  logs:
    statefulset:
      extraEnvVars:
        # Set env variables for awss3 exporter
        - name: AWS_ACCESS_KEY_ID
          value: "< YOUR AWS ACCESS KEY >"
        - name: AWS_SECRET_ACCESS_KEY
          value: "< YOUR AWS SECRET ACCESS KEY >"
```

## Send All Logs to Sumo Logic and Part of it to Other Locations

In case you want to send only part of logs to other location, you should use `sumologic.logs.otelcol.routing.table` in addition to `sumologic.logs.otelcol.extraExporters`.

In the following example, all logs are going to be send to Sumo Logic, and in addition only logs from `production` namespace are going to be send to AWS S3.

```yaml
sumologic:
  logs:
    otelcol:
      extraExporters:
        ## define `awss3/production` exporter
        awss3/production:
          s3uploader:
              region: 'eu-central-1'
              s3_bucket: 'databucket'
              s3_prefix: 'metric'
              s3_partition: 'minute'
      routing:
        table:
          ## send all logs from `production` namespace to `awss3/production` exporter
          ## as `useDefaultExporters` is set to `true` (default), all logs will be send to `sumologic` as well
          - exporter: awss3/production
            statement: route() where resource.attributes["namespace"] == "production"
metadata:
  logs:
    statefulset:
      extraEnvVars:
        # Set env variables for awss3 exporter
        - name: AWS_ACCESS_KEY_ID
          value: "< YOUR AWS ACCESS KEY >"
        - name: AWS_SECRET_ACCESS_KEY
          value: "< YOUR AWS SECRET ACCESS KEY >"
```

`sumologic.logs.otelcol.routing.table` is a list of maps, which consist of two keys `exporter` and `statement`.
Logs are send to `exporter` when `statement` is met. `exporter` should be defined in `sumologic.logs.otelcol.extraExporters`
or default `sumologic` may be used. `statement` has to be written in
[Open Telemetry Transformation Language][ottl]

:::note
Remember to use [attributes after translation to Sumo Logic Schema](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/sumologicprocessor#attribute-translation), for example `namespace` instead of `k8s.namespace.name`
:::

:::note
If log matches multiple statements, it will be send to all corresponding exporters.
For example, the following routing table will send logs from `production` namespace to `sumologic/production` and `awss3/production`:

```yaml
- exporter: awss3/production
  statement: route() where resource.attributes["namespace"] == "production"
- exporter: sumologic/production
  statement: route() where resource.attributes["namespace"] == "production"
```

:::

## Send Logs Selectively to Multiple Locations

This scenario differs from the previous ones by not sending all logs to Sumo Logic. If you want to send some data to Sumo Logic and some (the same or other) data to other locations, you need to disable default exporters. Set `sumologic.logs.otelcol.useDefaultExporters` to `false` and then set `sumologic.logs.otelcol.routing.fallbackExporters` to a list of exporters which are going to be used for the logs which do not match any of the routing table statements.

In the following example, logs from `production` namespace are going to be send to production Sumo Logic exporter and all remaining logs will be send to default Sumo Logic exporter:

```yaml
sumologic:
  logs:
    otelcol:
      ## useDefaultExporters set to false to stop sending all logs to default exporters
      useDefaultExporters: false
      extraExporters:
        ## define `sumologic/production` exporter
        sumologic/production:
          endpoint: http://my-production-sumologic-otlp-source
      routing:
        ## use sumologic as fallbackExporter, which means it will get all data which do not match any statement
        fallbackExporters:
          - sumologic
        table:
          ## send all logs from `production` namespace to `sumologic/production` exporter
          - exporter: sumologic/production
            statement: route() where resource.attributes["namespace"] == "production"
```

[ottl]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/routingprocessor/README.md#tech-preview-opentelemetry-transformation-language-statements-as-routing-conditions


deploy/helm/sumologic/values.yaml
```diff
@@ -353,6 +353,27 @@ sumologic:
...
+          []
+          ## -
+          ##  exporter is an exporter name. For example `exporter/name` for the following extraExporters configuration:
+          ##
+          ##  extraExporters:
+          ##    exporter/name:
+          ##      config_key: config_value
+          ##  exporter: sumologic
+          ##  ## set to true if the exporter should be used as default exporter
+          ##  default: true
```
swiatekm-sumo, line 372: Instead of this, I'd prefer to have a `otelcol.defaultExporter` setting, and make it mutually exclusive with `useDefaultExporters`.

deploy/helm/sumologic/values.yaml
```diff
@@ -463,6 +484,10 @@ sumologic:
 
     sourceType: otlp
 
+    ## additional exporters which may be used in any of the pipeline
+    ## remember add them to containers, systemd or kubelet routing as well
+    extraExporters: {}
```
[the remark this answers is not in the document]
sumo-drosiek: I will look at that. Unification may be not so trivial AFAIR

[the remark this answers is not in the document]
sumo-drosiek: It may be breaking change, as we have configuration per pipeline, especially `extraProcessors` and `sources` configuration, but I can implement one routing processor for all signals 👍

deploy/helm/sumologic/values.yaml
sumo-drosiek: Could you provide a configuration example? I'm not sure what would `defaultExporter` have

swiatekm-sumo:
```yaml
otelcol:
  defaultExporter: sumologic/new
  routing:
     - exporter: sumologic/test
       statement: route() where resource.attributes["exporter"] == "test"
extraExporters:
  sumologic/new:
    endpoint: https://new-endpoint
  sumologic/test:
    endpoint: https://test-endpoint
```

sumo-drosiek: Yes, it is possible, but it will look like the following:
```
otelcol:
  defaultExporters:
    - sumologic/new
    - debug/new
  routing:
     - exporter: sumologic/test
       statement: route() where resource.attributes["exporter"] == "test"
extraExporters:
  sumologic/new:
    endpoint: https://new-endpoint
  debug/new:
    endpoint: https://new-endpoint
  sumologic/test:
    endpoint: https://test-endpoint
```

And that way users needs to maintain 3 configuration keys

Also it requires users to know what our default exporters are, if they would like to send logs to S3 for example

sumo-drosiek: Also this is more complicated for logs, as we may have different exporter per type (systemd, kubelet, container) for http source, so `useDefaultExporters` is easier for us to maintain. We may migrate off it in v5, but for v4 I wouldn't expose default exporters as they depends on other configuration

deploy/helm/sumologic/values.yaml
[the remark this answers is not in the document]
sumo-drosiek: Due to https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3711#discussion_r1606381354 we cannot implement this feature for http source. As we cannot have one default for all three pipelines

sumo-drosiek:
I updated PR description with the documentation from the following PR: https://github.com/SumoLogic/sumologic-documentation/pull/4128

deploy/helm/sumologic/values.yaml
```diff
@@ -442,6 +442,25 @@ sumologic:
       ## Matching unit will be excluded from Sumo. The logs will still be sent to logs metadata provider (otelcol).
       excludeUnitRegex: ""
 
+    otelcol:
+      ## additional exporters which may be used in any of the pipeline
+      ## remember add them to containers, systemd or kubelet routing as well
+      extraExporters: {}
```
rnishtala-sumo, line 448: how does one specify which pipeline this extra exporter would be used in. For ex. containers or systemd?

rnishtala-sumo: > additional exporters which may be used in any of the pipeline
this tells me, that they would consider logs from all the pipelines

> remember add them to containers, systemd or kubelet routing as well
not sure I understand this comment.

sumo-drosiek: > remember add them to containers, systemd or kubelet routing as well

This is leftover from first iteration

They cannot choose the pipeline. This is to simplify the feature and I think we should aim to unify the pipelines in the future

deploy/helm/sumologic/values.yaml
```diff
@@ -442,6 +442,25 @@ sumologic:
       ## Matching unit will be excluded from Sumo. The logs will still be sent to logs metadata provider (otelcol).
       excludeUnitRegex: ""
 
+    otelcol:
+      ## additional exporters which may be used in any of the pipeline
+      ## remember add them to containers, systemd or kubelet routing as well
```
sumo-drosiek, line 447:
```suggestion
```

tests/integration/internal/k8s/pods.go
```diff
@@ -81,6 +81,14 @@ func WaitUntilSumologicMockAvailable(
 	waitDuration time.Duration,
 	tickDuration time.Duration,
 ) {
+	if ctxopts.Namespace(ctx) == internal.AdditionalSumologicMockNamespace {
+		kubectlOpts := *ctxopts.KubectlOptions(ctx)
+		kubectlOpts.Namespace = internal.AdditionalSumologicMockNamespace
```
swiatekm-sumo, line 86: Is this really necessary? The values of `ctxopts.Namespace(ctx)` and `kubectlOpts.Namespace` should be the same.

tests/integration/features.go
```diff
@@ -259,144 +259,319 @@ func addCollectorSpecificMetricLabels(labels sumologicmock.Labels, releaseName s
...
+	}
+
+	return feature.Feature()
+}
+
+func DeployAdditionalSumologicMock() features.Feature {
+	return features.New("create additional sumologic mock").
+		Setup(stepfuncs.KubectlApplyFOpt(internal.YamlPathAdditionalSumologicMock, internal.AdditionalSumologicMockNamespace)).
+		Assess("additional sumologic mock is ready", stepfuncs.WaitUntilAdditionalSumologicMockAvailable(waitDuration, tickDuration)).
+		Feature()
```
swiatekm-sumo, line 478: We should probably create the namespace here if needed and put it in kubectlOptions, and then in the deletion function below we can unset and delete it.

On that note, would it be a problem to make this namespace name depend on the test name? Otherwise it's annoying to run multiple tests in the same cluster, as namespaces take a while to delete.

swiatekm-sumo (commented):
The feature itself looks good to me. I left some comments about tests.

In particular, wouldn't it be better to use Go tests for templates, instead of golden files? It'd be way less text wasted on properties we don't care about.

tests/integration/internal/k8s/pods.go
sumo-drosiek, line 86:
```suggestion
```

sumo-drosiek:
> LGTM! one question do we expect this change to have an impact on our zaidan configuration, more than our customers using it? I haven't seen many customer tickets around custom routing.

Hard to say, I remember some topics around spliting traffic into multiple locations, but I believe it is not so common

rnishtala-sumo (approved):

tests/integration/features.go
sumo-drosiek: I need to test it, but it should be possible

sumo-drosiek: It would require generating configuration for test in-fly. It requires more changes which I want to include in this PR

tests/integration/internal/k8s/pods.go
sumo-drosiek: It doesn't work without this part. We are waiting for the service which is not a part of test

swiatekm-sumo (approved):

PR #3725, opened 2024-05-22
chore: Upgrading tailing-sidecar

rnishtala-sumo:
chore: Upgrading tailing-sidecar

- [x] Documentation updated

### DIFF
--- a/.changelog/3725.changed.txt
+++ b/.changelog/3725.changed.txt
@@ -1,0 +1,1 @@
+chore: Upgrading tailing-sidecar to v0.13.0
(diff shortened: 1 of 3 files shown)


.changelog/3725.changed.txt
```diff
@@ -0,0 +1 @@
+chore: Upgrading tailing-sidecar
```
swiatekm-sumo, line 1: Can you specify which version we're upgrading to?

swiatekm-sumo (approved):

PR #3744, opened 2024-05-31
test(integration): force delete fixed namespaces

swiatekm-sumo:
Namespace deletion can be slow in K8s due to finalizers. This can be a problem for fixed namespaces which are the same for each test that uses them. This change tries to improve this by explicitly disabling finalizers before deleting the namespace.

### DIFF
--- a/tests/integration/features.go
+++ b/tests/integration/features.go
@@ -465,7 +465,7 @@ func GetAllLogsFeature(waitFunction stepfuncs.WaitForLogs, generate bool) featur
 		feature = feature.
 			Teardown(removeLogsDeployment).
 			Teardown(removeLogsDaemonset).
-			Teardown(stepfuncs.KubectlDeleteNamespaceOpt(internal.LogsGeneratorNamespace))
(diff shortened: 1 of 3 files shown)


tests/integration/features.go
```diff
@@ -465,7 +465,7 @@ func GetAllLogsFeature(waitFunction stepfuncs.WaitForLogs, generate bool) featur
 		feature = feature.
 			Teardown(removeLogsDeployment).
 			Teardown(removeLogsDaemonset).
-			Teardown(stepfuncs.KubectlDeleteNamespaceOpt(internal.LogsGeneratorNamespace))
+			Teardown(stepfuncs.KubectlDeleteNamespaceOpt(internal.LogsGeneratorNamespace, true))
```
rnishtala-sumo, line 468: Is the primary reason for this change to speed up namespace termination?

swiatekm-sumo: Yes.

sumo-drosiek (approved):

PR #3762, opened 2024-06-13
chore(setup): store source manifests as json

swiatekm-sumo:
Instead of templating Terraform manifests in HCL, we instead do it in JSON. This is easier and less error-prone, as Helm lets us generate JSON directly from dictionaries in the Helm Chart configuration.

Note that we actually generate more than we need here. The other data will be used in subsequent changes from #3761, of which this change is a part of.


sumo-drosiek:
I need a time to process it :hell_chart:

tests/helm/testdata/goldenfile/terraform/collector_fields.output.yaml
```diff
@@ -753,6 +647,202 @@ data:
...
+                {
+                  "format": "epoch",
+                  "locator": "\\\"timestamp\\\":(\\\\d+)"
+                }
+              ]
+            }
+          },
+          "default_metrics_source": {
+            "config-name": "endpoint-metrics",
+            "name": "(default-metrics)"
```
chan-tim-sumo, line 692: just curious how come default metrics has parenthesis around it? 🤔

swiatekm-sumo: Someone long ago decided it should be this way, changing it is painful, and it doesn't really hurt anyone.

sumo-drosiek (approved):

PR #3766, opened 2024-06-17
chore(setup): pass log fields and collector fields via Terraform variables

swiatekm-sumo:
Pass these as Terraform variables. Define the values in a `terraform.tfvars.json` file, which Terraform loads automatically.

Collector fields require that we know the Helm Chart version, so pass that in as an environment variable.

### Checklist


sumo-drosiek (approved):
Looks OK, Hope you tested it 🙈

PR #3768, opened 2024-06-17
chore(setup): don't use templating for Terraform imports

swiatekm-sumo:
Import using the json file as a reference instead of templating in the bash script. This is the final part of #3761.

Also fixed a bug in cleanup, which was using the wrong Secret name. 

Finally, set an attribute on the generated Secret to align more with Kubernetes defaults.

### DIFF
--- a/deploy/helm/sumologic/conf/setup/setup.sh
+++ b/deploy/helm/sumologic/conf/setup/setup.sh
@@ -131,24 +131,9 @@ fi
 # Sumo Logic Collector and HTTP sources
 # Only import sources when collector exists.
 if terraform import sumologic_collector.collector "${SUMOLOGIC_COLLECTOR_NAME}"; then
-true  # prevent to render empty if; then
-{{- $ctx := .Values -}}
-{{- range $type, $sources := .Values.sumologic.collector.sources }}
-{{- if eq (include "terraform.sources.component_enabled" (dict "Values" $ctx "Type" $type)) "true" }}
-{{- range $key, $source := $sources }}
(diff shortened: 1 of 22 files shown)


deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -131,24 +131,9 @@ fi
...
-{{- else if and (eq $type "metrics") $ctx.sumologic.traces.enabled }}
-{{- /*
-If traces are enabled and metrics are disabled, create default metrics source anyway
-*/}}
-{{- if hasKey $sources "default" }}
-terraform import sumologic_http_source.{{ template "terraform.sources.name" (dict "Name" "default" "Type" $type) }} "${SUMOLOGIC_COLLECTOR_NAME}/{{ $sources.default.name }}"
-{{- end }}
-{{- end }}
-{{- end }}
+    jq -r '.resource[] | to_entries[] | "\(.key) \(.value.name)"' sources.tf.json | while read -r resource_name source_name; do
```
chan-tim-sumo, line 134: would it be better to have a common function for this since I see this is being repeated for all the files?

sumo-drosiek (approved):

deploy/helm/sumologic/conf/setup/setup.sh
swiatekm-sumo: What is being repeated? We call `jq` exactly once, and then loop over the results.

sumo-drosiek: I think we used to have more the same code in `cleanup.sh`

PR #3777, opened 2024-06-20
deps: upgrade opentelemetry-operator Chart to 0.56.1

swiatekm-sumo:
This is the last version before changes to CRD handling that will require changes on our end as well.

### Checklist

### DIFF
--- a/deploy/helm/sumologic/values.yaml
+++ b/deploy/helm/sumologic/values.yaml
@@ -2464,8 +2464,7 @@ opentelemetry-operator:
 
   ## Specific for OpenTelemetry Operator chart values
   admissionWebhooks:
-    failurePolicy: Fail
-    enabled: true
+    create: true
 
     ## skip admission webhook on our own OpenTelemetryCollector object to avoid having to wait for operator to start
     objectSelector:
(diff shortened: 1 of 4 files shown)


sumo-drosiek (approved):

deploy/helm/sumologic/values.yaml
```diff
@@ -2464,8 +2464,7 @@ opentelemetry-operator:
 
   ## Specific for OpenTelemetry Operator chart values
   admissionWebhooks:
-    failurePolicy: Fail
-    enabled: true
+    create: true
```
rnishtala-sumo, line 2467: will this change affect helm chart upgrades in anyway?

swiatekm-sumo: I don't think so. If someone disabled webhooks, they'll get an error and will need to set `create: false` instead of `create: true`. The other values I removed because we were using defaults anyway.

rnishtala-sumo: > changes to CRD handling that will require changes on our end as well.
ok, what about the upcoming changes mentioned here. Is there a breaking changes in the operator that we should be tracking to upgrade to the next version?

rnishtala-sumo (approved):

Issue #2699, opened 2022-12-14
Add tests for global configuration attributes


swiatekm:
The Chart allows various attributes to be set for all eligible resources. A simple example are labels, where we always set some common labels on all resources.

These properties should be covered by tests. A test for common labels is [already present](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/helm/common_test.go), but there are other properties which should be tested this way:

- [x]  `sumologic.pullSecrets`
- [x] `sumologic.podLabels`
- [x]  `sumologic.podAnnotations`
- [x]  `sumologic.serviceAccount.annotations`

and possibly more. 

This issue tracks adding these tests.

PR #3795, opened 2024-07-02
test(integration): Added custom pod label tests for global configuration attributes

chan-tim-sumo:
### Overview
- github issue: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/2699
- added pod label tests for global configs

3 more to go:
```
 sumologic.pullSecrets
 sumologic.podAnnotations
 sumologic.serviceAccount.annotations
```

### Checklist
- [ ] Documentation updated


sumo-drosiek:
Is that actually works? I'm trying to understand the flow and for me it looks like you are trying to check if rendered pod definition (which for example looks like this: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/helm/testdata/goldenfile/events_otc_statefulset/basic.output.yaml) contains `sumologic.podLabels` key.

The code should look like more like https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e52a34cf5af69d1746877f5575c26cb7b0e20a3b/tests/helm/common_test.go#L129-L180

Because aim of the task is to check if all pods/serviceaccounts/etc have the labels set correctly

The jira is not accessible outside of the organization so it would be much more Open Source to link correlated github issue instead 

I would also split this PR per individual change (`podLabels`, `podAnnotations`, etc)

chan-tim-sumo:
> Is that actually works? I'm trying to understand the flow and for me it looks like you are trying to check if rendered pod definition (which for example looks like this: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/helm/testdata/goldenfile/events_otc_statefulset/basic.output.yaml) contains `sumologic.podLabels` key.
> 
> The code should look like more like
> 
> https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e52a34cf5af69d1746877f5575c26cb7b0e20a3b/tests/helm/common_test.go#L129-L180
> 
> Because aim of the task is to check if all pods/serviceaccounts/etc have the labels set correctly
> 
> The jira is not accessible outside of the organization so it would be much more Open Source to link correlated github issue instead
> 
> I would also split this PR per individual change (`podLabels`, `podAnnotations`, etc)

i see, i thought we needed to test to make sure labels are in the rendered objects, if it is to check if all pods/everything have the labels set correctly, i guess just need to check if the labels arent empty ? 

ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ 

if so, i'm confused, isn't it already done for labels on here? https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e52a34cf5af69d1746877f5575c26cb7b0e20a3b/tests/helm/common_test.go#L48-L83

sumo-drosiek:
> if so, i'm confused, isn't there a test for labels on here? 

It tests for Helm built-in labels: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e52a34cf5af69d1746877f5575c26cb7b0e20a3b/tests/helm/common_test.go#L332-L341

> if it is to check if all pods/everything have the labels set correctly, i guess just need to check if the labels arent empty?

Yes, but we want to check for exact value. We need to ensure that we are able to set additional pod labels, pod annotations and so on for every pod. Maybe we need to use multiple configuration keys to achieve it, but this is want we want to test and ensure. We want to test the following feature: https://help.sumologic.com/docs/send-data/kubernetes/best-practices/#pod-labels

chan-tim-sumo:
> > if so, i'm confused, isn't there a test for labels on here?
> 
> It tests for Helm built-in labels:
> 
> https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/e52a34cf5af69d1746877f5575c26cb7b0e20a3b/tests/helm/common_test.go#L332-L341
> 
> > if it is to check if all pods/everything have the labels set correctly, i guess just need to check if the labels arent empty?
> 
> Yes, but we want to check for exact value. We need to ensure that we are able to set additional pod labels, pod annotations and so on for every pod. Maybe we need to use multiple configuration keys to achieve it, but this is want we want to test and ensure. We want to test the following feature: https://help.sumologic.com/docs/send-data/kubernetes/best-practices/#pod-labels

 if we want to check for EXACT values for ALL/additional pod labels, do you mean testing a custom pod label? since currently there is a test for BUILT-IN label only? 🤔

sumo-drosiek:
> just to make sure i'm understanding this, if we want to check for EXACT values for ALL/additional pod labels, do you mean testing a custom pod label? since currently there is a test for BUILT-IN label only?

Yes, we want to check exact values for additional pod labels

tests/helm/testdata/custom-labels.yaml
sumo-drosiek, line 1: We should enable optional components as well. see https://github.com/SumoLogic/sumologic-openshift-images/blob/main/scripts/values.yaml

tests/helm/common_test.go
```diff
@@ -413,6 +413,48 @@ func GetPodSpec(object unstructured.Unstructured) (*corev1.PodSpec, error) {
...
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(object.Object, cronJob)
+		if err != nil {
+			return nil, err
+		}
+		return &cronJob.Spec.JobTemplate.Spec.Template, nil
+	default:
+		return nil, nil
+	}
+}
+
```
sumo-drosiek, line 457: It can be simplified to something like this:

```suggestion
func GetPodTemplateSpec(object unstructured.Unstructured) (*corev1.PodTemplateSpec, error) {
	spec, err := GetPodSpec(object)
	if err != nil {
		return nil, err
	}
	return spec.Template, nil
}

```

chan-tim-sumo:
since most of the comments are here: 

another question i have is, I see pullSecrets and imagePullSecrets, are they the same?

pod annotation and service account annotation is pretty intuitive.

sumo-drosiek:
> I see pullSecrets and imagePullSecrets, are they the same?

It seems like that. We cannot change the subcharts configuration keys :(

tests/helm/common_test.go
```diff
@@ -369,55 +371,65 @@ func isSubchartObject(object metav1.Object) bool {
...
-// This only works on Deployments, StatefulSets and DaemonSets
-func GetPodSpec(object unstructured.Unstructured) (*corev1.PodSpec, error) {
+func GetPodSpec(podTemplateSpec *corev1.PodTemplateSpec) *corev1.PodSpec {
+	if podTemplateSpec == nil {
+		return nil
+	}
+	return &podTemplateSpec.Spec
+}
+
+func GetPodTemplateSpec(object unstructured.Unstructured) (*corev1.PodTemplateSpec, error) {
```
sumo-drosiek, line 381: I would leave previous siganture
```suggestion
func GetPodTemplateSpec(object unstructured.Unstructured) (*corev1.PodTemplateSpec, error) {
	podTemplateSpec, err := GetPodTemplateSpec(object)
	if err != nil {
		return nil, err
	}
	return &podTemplateSpec.Spec, nil
}

func GetPodTemplateSpec(object unstructured.Unstructured) (*corev1.PodTemplateSpec, error) {
```

sumo-drosiek: It require less refactor, and we also keep checking for the `podTemplateSpec` inside the function

tests/helm/common_test.go
```diff
@@ -547,3 +559,65 @@ func TestServiceAccountPullSecrets(t *testing.T) {
...
+			continue
+		}
+
+		if podTemplateSpec == nil {
+			t.Logf("PodTemplateSpec is nil for object %s", renderedObject.GetName())
+			continue
+		}
+
+		require.NoError(t, err)
+
```
sumo-drosiek, line 600:
```suggestion

```

It will be always true

sumo-drosiek (approved):
LGTM. We should add missing configuration options (if any) to https://help.sumologic.com/docs/send-data/kubernetes/best-practices/#pod-labels

PR #3818, opened 2024-08-01
feat: do not send any histogram or summary metric

sumo-drosiek:
### Checklist

Histogram is a lot of data points which are not used on Sumo Logic side by default. In case we will need to send some of them, we can figure out additional flag which will be an exception from that rule, but for now it sounds reasonable to stop ingesting buckets at all

The same is about summary and quantiles

- [ ] Documentation updated


kkujawa-sumo (approved):

kkujawa-sumo:
Please remember to add changelog entry.

sumo-drosiek:
Not quite sure about that, but we may probably also want to do the same for summary metrics

tests/integration/helm_prometheus_metrics_test.go
```diff
@@ -14,7 +14,7 @@ func Test_Helm_Prometheus_Metrics(t *testing.T) {
 	// defaults without otel metrics collector metrics, but with Prometheus metrics
 	expectedMetricsGroups := make([][]string, len(internal.DefaultExpectedMetricsGroups))
 	copy(expectedMetricsGroups, internal.DefaultExpectedMetricsGroups)
-	expectedMetricsGroups = append(expectedMetricsGroups, internal.PrometheusMetrics, internal.DefaultOtelcolMetrics)
+	expectedMetricsGroups = append(expectedMetricsGroups, internal.PrometheusMetrics, internal.DefaultOtelcolMetrics, internal.BucketMetrics)
```
sumo-drosiek, line 17: I'm not convinced to this change. as we filter out all buckets in metadata layer, it shouldn't be needed. Maybe telegraf receiver do not create histogram metrics 🤔

rnishtala-sumo (approved):

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
```diff
@@ -16,7 +16,7 @@ filter/drop_unnecessary_metrics:
       - resource.attributes["job"] != "pod-annotations" and IsMatch(name, "scrape_.*")
 {{- if .Values.sumologic.metrics.dropHistogramBuckets }}
       # drop histograms we've extracted sums and counts from, but don't want the full thing
-      - IsMatch(name, "^(apiserver_request_duration_seconds|coredns_dns_request_duration_seconds|kubelet_runtime_operations_duration_seconds)$")
+      - type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY or IsMatch(name, ".*_bucket")
```
rnishtala-sumo, line 19: how would we send one or two histogram metrics?

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
```diff
@@ -16,7 +16,7 @@ filter/drop_unnecessary_metrics:
       - resource.attributes["job"] != "pod-annotations" and IsMatch(name, "scrape_.*")
```
rnishtala-sumo, line 16: should we also drop metrics coming in from pod annotations by default?

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
sumo-drosiek: by exposing new configuration option, and adding `not isMatch` at the end of the query

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
sumo-drosiek: It's breaking change

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
sumo-drosiek: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/3819

sumo-drosiek: I wouldn't play with `merge` if we can do it easily by the configuration option

PR #3821, opened 2024-08-06
feat(metrics): Define allowlist for histogram metrics

rnishtala-sumo:
feat(metrics): Define allowlist for histogram metrics. This change introduces an allowlist for Histograms, using a regex, since we now disable all histogram metrics by default.

- [x] Documentation updated


deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -57,8 +57,8 @@ processors:
     metric_statements:
       - context: metric
         statements:
-          - extract_sum_metric(true) where type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY
-          - extract_count_metric(true) where type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY
+          - extract_sum_metric(true) where not IsMatch(name, .Values.sumologic.metrics.allowHistogramRegex) and type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY
```
rnishtala-sumo, line 60: @sumo-drosiek seems like an allowlist (using regex) here is invalid, since ottl doesn't support template variables with regex pattern matching. The second argument to `IsMatch` needs to be a string.

sumo-drosiek: This is helm template, so we should use the curly braces to interpolate the variable, like here:

https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/8e3a9e765021467daa639e51ef65ea1efb058163/deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml#L9

We can use goldenfiles to ensure that the template is rendered correctly

sumo-drosiek: I would also use parenthesis to group remaining conditions to ensure that everything is resolved correctly

rnishtala-sumo: I understand that, however ottl doesn't seem to support using helm template variables or variables in general to specify regex patterns.

sumo-drosiek: I don't understand. After template will be generated, there won't be any variables

```suggestion
          - extract_sum_metric(true) where not IsMatch(name, {{ .Values.sumologic.metrics.allowHistogramRegex | quote }}) and type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY
```

rnishtala-sumo: ok let me test out the change.

rnishtala-sumo: I see what you mean, the value of the template variable is injected before evaluating the ottl expression. thanks!

rnishtala-sumo: for some reason the `_bucket` metrics are passing through with this change.

sumo-drosiek: I think everything matches empty string. We should make it more sophisticated, so add `not IsMatch(name, {{ .Values.sumologic.metrics.allowHistogramRegex | quote }}) and ` only if `.Values.sumologic.metrics.allowHistogramRegex` is not empty

echlebek (approved):
LGTM

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
```diff
@@ -16,7 +16,7 @@ filter/drop_unnecessary_metrics:
       - resource.attributes["job"] != "pod-annotations" and IsMatch(name, "scrape_.*")
 {{- if .Values.sumologic.metrics.dropHistogramBuckets }}
       # drop histograms we've extracted sums and counts from, but don't want the full thing
-      - type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY or IsMatch(name, ".*_bucket")
+      - (not IsMatch(name, {{ .Values.sumologic.metrics.allowHistogramRegex | quote }})) and (type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY or IsMatch(name, ".*_bucket"))
```
sumo-drosiek, line 19:
```suggestion
      - not IsMatch(name, {{ .Values.sumologic.metrics.allowHistogramRegex | quote }}) and (type == METRIC_DATA_TYPE_HISTOGRAM or type == METRIC_DATA_TYPE_EXPONENTIAL_HISTOGRAM or type == METRIC_DATA_TYPE_SUMMARY or IsMatch(name, ".*_bucket"))
```

sumo-drosiek (approved):
LGTM, we need to document the change for customers on `help.sumologic.com`

PR #3824, opened 2024-08-14
docs(histogram-metrics): Documentation updates for allowing histogram metrics (dropped by default)

rnishtala-sumo:
Updates to the doc around allowing histogram metrics

- [x] Documentation updated

### DIFF
--- a/docs/scraped-metrics.md
+++ b/docs/scraped-metrics.md
@@ -32,163 +32,119 @@
 
 ## Metrics
 
-The following table contains information about metrics scraped by Sumo Logic's Prometheus:
+The following table contains information about metrics scraped by Sumo Logic's metrics collector:
 
 - name - name of the metric
 - source - originated source of the metric
-- forwarded - `yes` if metric is being forwarded to Sumo Logic
 
-| name                                                     | source             | forwarded |
-| -------------------------------------------------------- | ------------------ | --------- |
-| `apiserver_request_count`                                | apiserver          | yes       |
-| `apiserver_request_total`                                | apiserver          | yes       |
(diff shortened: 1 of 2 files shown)


docs/scraped-metrics.md
```diff
@@ -38,157 +38,114 @@ The following table contains information about metrics scraped by Sumo Logic's P
...
-| `node_memory_MemTotal_bytes`                             | node-exporter      | yes       |
-| `node_network_receive_drop_total`                        | node-exporter      | yes       |
-| `node_network_transmit_drop_total`                       | node-exporter      | yes       |
-| `node_network_receive_bytes_total`                       | node-exporter      | yes       |
-| `node_network_transmit_bytes_total`                      | node-exporter      | yes       |
-| `node_filesystem_avail_bytes`                            | node-exporter      | yes       |
-| `node_filesystem_size_bytes`                             | node-exporter      | yes       |
-| `node_filesystem_files_free`                             | node-exporter      | yes       |
-| `node_filesystem_files`                                  | node-exporter      | yes       |
+| name                                              | source             | forwarded |
```
echlebek, line 41: This table now only includes metrics that are forwarded to Sumo Logic. So now the "forwarded" column is not useful. (Before there was only one metric mentioned that was not forwarded.) Maybe we should remove it.

echlebek (approved):

docs/scraped-metrics.md
rnishtala-sumo: yes I agree, if its scraped, we send it by default.

sumo-drosiek (approved):

PR #3884, opened 2025-02-04
feat: Make the metrics level in service telemetry configurable

rnishtala-sumo:
Make the metrics level in service telemetry configurable

- [] Changelog updated or skip changelog label added
- [ ] Documentation updated

### DIFF
--- a/tests/helm/testdata/goldenfile/events_otc/options.input.yaml
+++ b/tests/helm/testdata/goldenfile/events_otc/options.input.yaml
@@ -1,5 +1,6 @@
 otelevents:
   logLevel: debug
+  metricsLevel: detailed
 
 sumologic:
   clusterName: testCluster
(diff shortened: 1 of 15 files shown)


tests/helm/testdata/goldenfile/events_otc/options.input.yaml
```diff
@@ -1,5 +1,6 @@
 otelevents:
   logLevel: debug
+  metricsLevel: detailed
```
chan-tim-sumo, line 3: is there a difference between ```metricsLevel``` and 

```
  metrics:
    level: normal
```

rnishtala-sumo: yes the value for `metricsLevel` in the input `values` translates to `metric.level` in the output.

chan-tim-sumo (approved):

PR #3898, opened 2025-03-05
feat: Migrate container logs enrichment to use the k8sattributes processor

rnishtala-sumo:
Migrate container logs enrichment to use the k8sattributes processor. Keep this new processor behind a feature flag.

Docs for the upstream k8s processor - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md

- [x] Documentation updated


deploy/helm/sumologic/conf/logs/collector/common/filelog_receiver.yaml
```diff
@@ -196,6 +196,10 @@ filelog/containers:
     - from: attributes.pod_name
       to: attributes["k8s.pod.name"]
       type: move
+    
+    - from: attributes.uid
```
chan-tim-sumo, line 200: why are uid, pod label, and namespace labels added? just curious since I didn't see much in the description 🤔

rnishtala-sumo: pid UID is used to associate a pod with its metadata. When a new log flows through the processor, it has a pod UID as an attribute (with this change), which the processor uses to fetch the metadata.

Pod and namespace labels are part of the metadata enrichment.

samiura (approved):

deploy/helm/sumologic/conf/logs/collector/common/filelog_receiver.yaml
rnishtala-sumo: @chan-tim-sumo @echlebek other questions about this PR? :)

chan-tim-sumo: was ```k8s.pod.uid``` the only new data/field that was added? if yes, is the feature flag even needed?

rnishtala-sumo: `k8s.pod.uid` is only to associate metadata with the pod.

The feature flag is to toggle between the upstream k8s processor and the sumo k8s processor.

The proposal is that we continue with the sumo k8s processor as the default for the next release and provide the customer an option to switch to the upstream one. Does that make sense?

chan-tim-sumo: ya that makes sense, what testing was done for this? i'm assuming just deploying locally and making sure the container logs are still coming in?

chan-tim-sumo: looks good to me!

chan-tim-sumo (approved):

PR #3934, opened 2025-05-21
chore: fix link linter

echlebek:
Fix the link linter by not following Stack Overflow links. Stack Overflow is issuing 403 to crawlers and seems to consider Github Actions as one of them.

### DIFF
--- a/.markdown_link_check.json
+++ b/.markdown_link_check.json
@@ -14,6 +14,9 @@
     },
     {
       "pattern": "^http://localhost"
+    },
+    {
+      "pattern": "^https://stackoverflow.com/"
     }
   ]
 }


ccressent (approved):
There seem to be a precedent with `https://mikefarah.gitbook.io/yq` above, so I suppose this is fine. Conceptually though, if we just end up ignoring most URLs, we might as well remove that link check? 😓

echlebek:
There’s good reason to discuss removing it but this is one link out of dozens that isn’t possible to check. It unbreaks CI so I think we should merge this now and circle back to the linter.

chan-tim-sumo (approved):

PR #3935, opened 2025-05-22
fix: add affinity from events statefulset

### LINKED ISSUE #3830: Affinity in sumologic-sumologic-otelcol-events StatefulSet is Discarded
I want to be able configure affinity for sumologic-otelcol-events StatefulSet with using helm values file.

I checked how affinity is added. It seems like the template checks if it is empty, and if not, it adds other components such as podAntiAffinity. However, it does not add affinity itself. See: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.9.0/deploy/helm/sumologic/templates/events/otelcol/statefulset.yaml#L47-L57

If we look at https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.9.0/deploy/helm/sumologic/templates/logs/otelcol/statefulset.yaml#L56-L58 we can see that the affinity is included in the sumologic-sumologic-otelcol-logs StatefulSet.

```
sumologic:
  otelevents:
    statefulset:
      affinity:
        nodeAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 10
            preference:
              matchExpressions:
                - key: my-key
                  operator: In
                  values:
                    - "my-value"

```
You can try using above config as helm values file.
We should see the affinity block added to sumologic-otel-col-events StatefulSet.

echlebek:
It appears that affinity from the instrumentation pod was included incorrectly.

Closes #3830

### DIFF
--- a/deploy/helm/sumologic/templates/events/otelcol/statefulset.yaml
+++ b/deploy/helm/sumologic/templates/events/otelcol/statefulset.yaml
@@ -46,9 +46,6 @@ spec:
 {{- end }}
 {{- if or (not (empty (include "events.statefulset.affinity" .))) .Values.otelevents.statefulset.podAntiAffinity }}
       affinity:
-{{- if not (empty (include "otelcolInstrumentation.statefulset.affinity" .)) }}
-{{ include "otelcolInstrumentation.statefulset.affinity" . | indent 8 }}
-{{- end }}
 {{- if not (empty (include "events.statefulset.affinity" .)) }}
 {{ include "events.statefulset.affinity" . | indent 8 }}
 {{- end }}


deploy/helm/sumologic/templates/events/otelcol/statefulset.yaml
```diff
@@ -46,8 +46,8 @@ spec:
 {{- end }}
 {{- if or (not (empty (include "events.statefulset.affinity" .))) .Values.otelevents.statefulset.podAntiAffinity }}
       affinity:
-{{- if not (empty (include "otelcolInstrumentation.statefulset.affinity" .)) }}
-{{ include "otelcolInstrumentation.statefulset.affinity" . | indent 8 }}
+{{- if not (empty (include "events.statefulset.affinity" .)) }}
+{{ include "events.statefulset.affinity" . | indent 8 }}
```
rnishtala-sumo, line 50: might be worth adding a template test for this - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/helm/README.md#helm-template-tests

rnishtala-sumo: example - https://github.com/SumoLogic/sumologic-kubernetes-collection/commit/114d97cd2a50d7cfb6db4c06456043b9ab643f0a#diff-87c4773b744518d0cc99baee3dfadffcaa566f467cbdcce2db55e26785ca3d76

echlebek: Agreed, thanks for pointing me to that!

PR #3940, opened 2025-06-03
feat: allow for hostNetwork in otellogs.daemonset

poblahblahblah:
This allows the consumer to set hostNetwork for the otellogs daemonset. This defaults to `false` to preserve the current behavior.

For some background on this we run a handful of large clusters where IP space is at a premium. In our larger clusters this would save ~2000 IPs, which would be great to claw back!

### Checklist

- [X] Documentation updated
- [X] Integration tests added or modified for major features

### DIFF
--- a/.changelog/3940.feat.txt
+++ b/.changelog/3940.feat.txt
@@ -1,0 +1,1 @@
+feat: allow for hostNetwork in otellogs.daemonset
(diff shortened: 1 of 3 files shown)


echlebek:
Hi @poblahblahblah, just wondering if you have successfully tried this on your (EKS?) cluster. It seems to me that you would also need `dnsPolicy: ClusterFirstWithHostNet`?

We require new integration tests that exercise this feature before we can consider it for merge. Can you please add them?

Regarding the CI failure for the changelog, I think it's because you used "feat" instead of "added", which we have specified in the towncrier configuration.

Regarding the CI failure for the operator instrumentation, there could be a flaky test here. We will look into it.

echlebek:
Closing in favour of #3993 which also includes this functionality.

PR #3949, opened 2025-06-20
feat(config): ipv6 compatibiilty for otel collector pods

jagan2221:
1. Collector pods were crashing in ipv6 env due to invalid ip address format used for exposing endpoints such as health check, liveliness and otlp/http/grpc server endpoints etc.
       We are constructing these endpoints using podIP env variable , need to enclose it with square braces. Added a generic flag to configure this for all endpoints.

2. Regex used for extracting pod ip and port in pod-annotations scrape config does not support ipv6 parsing. 
       Introduced a method to construct scrape address pod ip and prometheus port

Both above changes are behind a feature flag - sumologic.ipv6mode
 
Detailed summary of the changes: https://docs.google.com/document/d/1aBfne-cN6k9p_Lw3zZ2ZqY8Ho0AHYTzQB-nIpYx--GA
These are the issue faced by a customer when setting up helm chart in ipv6 cluster. These fixes are deployed in customer setup using manual overrides using config merge.
Jira: https://sumologic.atlassian.net/browse/OSC-1043 

### Checklist

- [x] Documentation updated


deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -125,7 +129,7 @@ receivers:
               regex: (.+)
             - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
               action: replace
-              regex: ([^:]+)(?::\d+)?;(\d+)
+              regex: (\[.*?\]|[^:;\[]+)(?::\d+)?;(\d+)
```
rnishtala-sumo: Might make sense to have an integration test just for metrics collection because of this change.

rnishtala-sumo:
We need a changelog for this. The approach makes sense to me. Lets ensure that the integration tests pass. It might also make sense to write an integration test for metrics collection.

jagan2221:
https://github.com/SumoLogic/sumologic-kubernetes-collection/actions/runs/15817837079/job/44580109359?pr=3949
Something wrong with Helm_Routing_OT test - it was successful in above run for the same PR, but failing in other runs.

I've checked in other PR's too, this test is failing/passing randomly.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: We already have a test for checking metrics scraped by pod-annotations job [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/c1cf42c633c852696d9e01904ba36af88a5ddf88/tests/integration/features.go#L248)

This is included in Helm_Default_OT test where default metrics are ensured by setting up an nginx pod with prometheus.io/scrape: true and pod-annotations job scrape this pod. We can say existing functionality of the regex scraping pods with ipv4 address in already validated in existing IT.

For ipv6 mode, we need to setup the cluster itself in ipv6 single-stack mode which will be a major change to IT's. Once this is done, all tests can be run in ipv6 mode too without introducing new tests specific for ipv6.
@rnishtala-sumo

rnishtala-sumo: `Kind` supports IPv6 clusters via its configuration file

Here's where the current kind config is defined - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/yamls/cluster.yaml

> For ipv6 mode, we need to setup the cluster itself in ipv6 single-stack mode which will be a major change to IT's
 
We could define a cluster config for ipv6 and use it conditionally for one test.

jagan2221: > Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ Error: Failed to query available provider packages
> Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ 
> Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ Could not retrieve the list of available versions for provider
> Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ sumologic/sumologic: could not connect to registry.terraform.io: failed to
> Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ request discovery document: Get
> Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ "https://registry.terraform.io/.well-known/terraform.json": dial tcp:
> Test_Helm_Default_OT_NamespaceOverride 2025-06-23T17:38:23Z logger.go:67: │ lookup registry.terraform.io on [fd00:10:96::a]:53: server misbehaving

Tried creating the ipv6 cluster. It's created fine, but we need to setup the test running machine to be able to access internet using ipv6 config as [mentioned here](https://docs.google.com/document/d/1aBfne-cN6k9p_Lw3zZ2ZqY8Ho0AHYTzQB-nIpYx--GA/edit?tab=t.0#heading=h.j6zu73hg6bon)

It's not very simple to setup a proper ipv6 cluster with networking in existing IT's. I think it should be separate proper task itself to setup ipv6 based IT's. 

+ If we want to be safe, we can have new regex only when sumologic.ipv6mode enabled, else we can use the existing old regex.

@rnishtala-sumo

rnishtala-sumo: IMO we need either ITs or E2Es before rolling this feature out. We're not using AWS for running the ITs.

Might be worth looking at this - https://github.com/kubernetes-sigs/nat64

rnishtala-sumo (changes requested):
Requesting ITs for this change because of its large footprint.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: 2025/06/23 20:41:05 Applying nat64 install.yaml...
2025/06/23 20:41:06 nat64 install.yaml applied: clusterrole.rbac.authorization.k8s.io/nat64 created
clusterrolebinding.rbac.authorization.k8s.io/nat64 created
serviceaccount/nat64 created
daemonset.apps/nat64 created
=== RUN   Test_Helm_Default_OT_NamespaceOverride
Test_Helm_Default_OT_NamespaceOverride 2025-06-23T20:46:59Z logger.go:67: │ lookup registry.terraform.io on [fd00:10:96::a]:53: server misbehaving
Test_Helm_Default_OT_NamespaceOverride 2025-06-23T20:46:59Z logger.go:67: 

Tried setting up nat64. Doesn't looks to help!

rnishtala-sumo: this vagrant setup looks interesting - https://github.com/leblancd/kube-v6?tab=readme-ov-file#overview

Some registries like dockerhub/terraform don't support IPv6 (hence the above error) which is where NAT64 plays a role. If we're unable to setup this up in the ITs, we should probably look at updating out vagrant [doc](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/vagrant/README.md) to set this up locally to test this.

Either way, there needs to be a test strategy IMO before merging this to main.

jagan2221: @rnishtala-sumo Thanks for your inputs.

Yes, i'm exploring to setup nat64 to see if it helps. 
Trying multiple things..

1. Nat64 [setup](https://github.com/kubernetes-sigs/nat64) 
       I0624 14:42:02.208252       1 main.go:189] Watching for node, awaiting podCIDR allocation
       I0624 14:42:02.301363       1 main.go:197] podCIDR: fd00:10:244:1::/64 
       F0624 14:42:02.301393       1 main.go:206] current network parameters are not valid: mask for pod CIDR     
fd00:10:244:1::/64 must be of size 112 or narrower to fit into the NAT64 defined range: 169.254.0.0/16
     Seems nat64 needs specific subnet range in the pod and node level. Trying to fix this.

2. Since it's just tests, we won't need to create TF resources and we can skip the same using sumologic.setupEnabled=false so that we can avoid failures when reaching terraform registry. This is the method i have tested ipv6 in local setup.  Will update this in vagrant readme if Nat64 setup doesn't help.

deploy/helm/sumologic/conf/events/otelcol/config.yaml
```diff
@@ -30,7 +30,11 @@ extensions:
     timeout: 10s
 {{- end }}
   health_check:
+{{- if eq .Values.sumologic.ipv6mode true }}
```
chan-tim-sumo: instead of a bunch of if else in all the actual files, shouldn't this be inside a helper template instead? so then the actual files wouldn't be all jumbled up with if else statements

[example](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/7f5eb427be31e72e76cab013f146f5eefcd1236e/deploy/helm/sumologic/templates/_helpers/_common.tpl#L4), something like:
```
{{- define "sumologic.healthEndpoint" -}}
{{- if .Values.sumologic.ipv6mode -}}
[${env:MY_POD_IP}]:13133
{{- else -}}
${env:MY_POD_IP}:13133
{{- end -}}
{{- end }}
```

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: sumologic.setupEnable=false solves the issue by not creating sources, so not contacting terraform registry(but we need to create a dummy sumologic secret- because pods are referring this secret for all endpoints)
But we have another issue- Sumologic mock is not starting up since it's hard-coded to bind to ipv4 address.
https://github.com/SumoLogic/sumologic-kubernetes-tools/blob/main/src/rust/sumologic-mock/src/main.rs#L269

We need to first fix sumologic-mock to be ipv6 compatible to proceed further.
This creating of IPv6 test setup itself is taking lot of changes and IMO, it should be separate task by itself.

rnishtala-sumo: > This creating of IPv6 test setup itself is taking lot of changes and IMO, it should be separate task by itself

Then I recommend adding vagrant documentation with test instructions to setup an ipv6 cluster for local testing.

.markdownlint.jsonc
```diff
@@ -8,5 +8,6 @@
     "MD024": false, // no-duplicate-heading/no-duplicate-header
     "MD033": false, // no-inline-html
     "MD034": false, // no-bare-urls
-    "MD040": false // fenced-code-language
+    "MD040": false, // fenced-code-language
+    "MD059": false // disable checks for descriptive link text
```
jagan2221, line 12: This exclusion added since we have changes in readme, seems a newer validation added in recent lint checks to have only descriptive link texts, we have the name as "link" in many existing places. So, skipping this rule

deploy/helm/sumologic/README.md
```diff
@@ -28,6 +28,7 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
 | `namespaceOverride`                                                                   | Used to override the chart's default target namepace.                                                                                                                                                                                                                                                      | `Nil`                                                                                                                                                                                |
+| `sumologic.ipv6mode`                                                                  | If enabled, configures the chart to be compatible with ipv6 single stack clusters                                                                                                                                                                                                                          | `false`                                                                                                                                                                              |
```
jagan2221, line 31: All changes done here are behind this feature flag only

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -123,11 +123,19 @@ receivers:
               action: replace
               target_label: __metrics_path__
               regex: (.+)
+{{- if eq .Values.sumologic.ipv6mode true }}
+            - source_labels: [__meta_kubernetes_pod_ip, __meta_kubernetes_pod_annotation_prometheus_io_port]
+              action: replace
+              regex: (.*);(.*)
+              replacement: "[$1]:$2"
+              target_label: __address__
+{{- else }}
```
jagan2221, line 132: Constructing __address__ from podip and prometheus port, since old flow which we use below for ipv4 gives different __address__ values for ipv6 only clusters.
So, it's a better approach to construct this from podIP.
Ex. formats observed in __address__ value
1.  ipv6:port
2. ipv6

tests/integration/features.go
```diff
@@ -116,13 +117,15 @@ func GetMetricsK8sattributes(expectedMetrics []string, metricsCollector MetricsC
 		Feature()
 }
 
-func GetMetricsFeature(expectedMetrics []string, metricsCollector MetricsCollector) features.Feature {
+// addKubeVersionSpecificMetrics - adds metrics that are specific to the Kubernetes version to expectedMetrics
```
jagan2221: Existing WaitUntilExpectedMetricsPresent() method is adding some core-dns metrics to expectedMetrics internally based on k8's version. Using this flag to check only the values passed in expectedMetrics list and skip adding any additional metrics for validation.

tests/integration/helm_ot_default_ipv6_test.go
```diff
@@ -0,0 +1,66 @@
...
+	featLogs := GetAllLogsFeature(stepfuncs.WaitUntilExpectedExactLogsPresent, true)
+
+	featMultilineLogs := GetMultipleMultilineLogsFeature()
+
+	featEvents := GetEventsFeature()
+
+	featTraces := GetTracesFeature()
+
+	testenv.Test(t, featInstall, featMetrics, featTelegrafMetrics, featLogs, featMultilineLogs, featEvents, featTraces)
+}
```
jagan2221: Test will all default functionalities tested for IPv6

tests/integration/main_test.go
```diff
@@ -86,12 +101,18 @@ func ConfigureTestEnv(testenv env.Environment) {
...
+	beforeFuncs := []features.Func{
+		stepfuncs.KubectlCreateNamespaceTestOpt(),
 		stepfuncs.KubectlCreateOperatorNamespacesOpt(),
 		stepfuncs.KubectlCreateOverrideNamespaceOpt(),
 		stepfuncs.HelmVersionOpt(),
 		stepfuncs.HelmDependencyUpdateOpt(internal.HelmSumoLogicChartAbsPath),
-		// HelmInstallTestOpt picks a values file from `values` directory
+	}
+	if isIPv6Test {
+		beforeFuncs = append(beforeFuncs, stepfuncs.KubectlCreateSumologicSecretOpt())
```
jagan2221, line 112: Secret created manually in ipv6 clusters since we needed to disabled sumologic.setupEnabled flag to prevent installation failure due to terraform registry not reachable. Just "dummy" values added as secrets, refer sumologic-secrets.yaml

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: Added IT for ipv6. @rnishtala-sumo

tests/integration/features.go
rnishtala-sumo: let's not do this. Instead, there is a way to add `coredns_cache_hits_total` to an optional list. The list of expected metrics is maintained in this file - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.13.0/tests/integration/internal/constants.go

tests/integration/helm_ot_default_ipv6_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal"
+	"github.com/SumoLogic/sumologic-kubernetes-collection/tests/integration/internal/stepfuncs"
+)
+
+func FilterOutIPv6ExcludedMetrics(metrics []string) []string {
```
rnishtala-sumo: Remove this filter in favor of maintaining the list of expected metrics here - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.13.0/tests/integration/internal/constants.go

tests/integration/helm_ot_default_namespaceoverride_test.go
```diff
@@ -43,7 +43,7 @@ func Test_Helm_Default_OT_NamespaceOverride(t *testing.T) {
 
 	featInstall := GetInstallFeature(installChecks)
 
-	featMetrics := GetMetricsFeature(expectedMetrics, Prometheus)
+	featMetrics := GetMetricsFeature(expectedMetrics, Prometheus, true)
```
rnishtala-sumo: same here, please undo

tests/integration/internal/stepfuncs/assess_funcs.go
```diff
@@ -157,14 +157,17 @@ func WaitUntilExpectedMetricsPresent(
...
+	addKubeVersionSpecificMetrics bool,
 ) features.Func {
 	return func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
 		sumoMockServiceName := GetSumoMockServiceName(ctx, t)
 		k8s_internal.WaitUntilSumologicMockAvailable(ctx, envConf, t, sumoMockServiceName, waitDuration, tickDuration)
 
-		// We can't do it earlier, because we run the tests for different k8s versions
-		// and we can't fetch current version earlier
-		expectedMetrics = append(expectedMetrics, internal.GetVersionDependentMetrics(t)...)
+		if addKubeVersionSpecificMetrics {
```
rnishtala-sumo: shouldn't need this anymore based on the above comment

tests/integration/internal/stepfuncs/kubectl.go
```diff
@@ -172,3 +172,16 @@ func GetHelmValuesForT(t *testing.T) []byte {
 	require.NoError(t, err, "values file %s must exist", valuesFilePath)
 	return valuesFileBytes
 }
+
+// KubectlCreateSumologicSecretOpt returns a features.Func that will create a secret called "sumologic"
+// in the namespace where the chart is installed using the YAML file.
+func KubectlCreateSumologicSecretOpt() features.Func {
```
rnishtala-sumo, line 178: I don't think we should be doing this outside the helm chart installation to test ipv6 support.

tests/integration/main_test.go
rnishtala-sumo: The customer is going to deploy helm charts with the setup job enabled. Have you tried using https://github.com/kubernetes-sigs/nat64? If this isn't working I suggest we add manual setup/test instructions in the vagrant readme to test this locally.

tests/integration/yamls/sumologic-secret.yaml
```diff
@@ -0,0 +1,24 @@
+# This secret is used for ipv6 clusters where we disabled setup job which is responsible for creating the secret.
+# Reason is the limitation of Github runners are ipv4 only. So setting up NAT64 would be very tricky and tedious and we also don't need backend url's for IT's
```
rnishtala-sumo, line 2: Disabling the setup job to test ipv6 is not preferred. The setup job does more than managing secrets. Instead of implementing this workaround, we should consider adding manual test instructions to our vagrant docs if using dns64 or nat64 is complex.

rnishtala-sumo (changes requested):
Disabling the setup job to test ipv6 is not preferred. The setup job does more than managing secrets. Instead of implementing this workaround, we should consider adding manual test instructions to our vagrant docs if using dns64 or nat64 is complex.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
rnishtala-sumo: I've reviewed the changes and appreciate the effort. I think we should think about the limitation we're running into more carefully before rolling this out.

Here's what we could do - 
- Remove the IT from this PR and changes around that.
- Add test instructions to our vagrant doc to setup an ipv6 cluster and deploy the helm chart, so someone from the team could use those instructions to test.

jagan2221: @rnishtala-sumo 
IT's with setup disabled is helping us test out all the features with ipv6. 
Even when we add test steps in vagrant, it would be disabling the setup job and install the helm chart since configuring NAT64 will require substantial effort and analysis to implement.

This disableSetupJob is just in IT's and better than not having IT's at all right? I don't understand the issue with disabling the setupjob for ipv6 tests alone as long as it gives us feasibility to test functionalities.

tests/integration/main_test.go
jagan2221: Tried nat64, but didn't work

tests/integration/yamls/sumologic-secret.yaml
jagan2221: Agreed that setupjob could do multiple things. But there are multiple tests to test setup job features, but this test focuses on specifically IPv6 communication and helped me even catch the issue with previously proposed regex for scrape job(pod-annotations) and to forced me to fix it.  Similarly i think this test will help us capture much issues coming in ipv6 setups.

In future , if we succeed settingup nat64, we can just remove secretcreation and enable setup job. Till then, IMO, this internal ipv6 test will help us to a good extent.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
rnishtala-sumo: All tests install/deploy the helm chart. I'd rather be sure that there aren't issues with the installation including setup on an ipv6 cluster than turn it off. This is a pre-requisite for collecting any signal.

My recommendation is to ensure that there are no installation issues (including setup), even if this is manual (documented) to start with.

jagan2221:
Even vagrant test steps will be based on disabling setupJob and test helm chart further. Hope this would be fine for local test steps @rnishtala-sumo

jagan2221:
@rnishtala-sumo  @echlebek We have deployed these changes in second customer setup(ChargeCloud) this week in their ipv6 only cluster and the deployment is fine. 

From the two customers we have seen, the issue customer asked to solve us is the Sumo pod level issues and not nat64/dns64 Network level issues. 

QE started testing as well. We would need this to be merged for QE to test E2E's.

rnishtala-sumo:
Recommend considering documentation for atleast one cluster type before merging this. A walkthrough of the manual steps before the helm chart is deployed is ideal. Could be reviewed by QE as well. An example can be seen in the doc for [EKS Fargate](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/docs/fargate.md). This could be in a different PR.

jagan2221:
https://docs.google.com/document/d/1ifCHtPsrz9ntTYyigGV0RkOb_9OzVp8j_DmmfqHJ2yM
@rnishtala-sumo Creating a draft doc for the same. Can you review this? Will get it reviewed from QE as well.

rnishtala-sumo:
@jagan2221 what I meant was a public facing doc on github like this one - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/docs/fargate.md (could call it ipv6.md) that runs the customer through the following steps as pre-requisites before installing the helm chart. It could say something like the following and QE could use the same method to test this for consistency.

```
In AWS, the way to provide IPv6→IPv4 egress is by setting up an NAT64 + DNS64 environment. The recommended approach is to use AWS NAT64 (via NAT gateway + VPC Route 64) with DNS64 resolver. This allows your IPv6-only pods to reach IPv4 destinations transparently.

To set up a NAT Gateway:
aws ec2 create-nat-gateway \
  --subnet-id <public-subnet-id> \
  --allocation-id <eip-allocation-id> \
  --connectivity-type public

Note NAT gateway only translates IPv4 → IPv4, so far.

Enable DNS64 on your VPC’s AmazonProvidedDNS run:
aws ec2 modify-vpc-attribute \
  --vpc-id <vpc-id> \
  --enable-dns64

For subnets where your IPv6-only pods run, add a default route for the well-known IPv6 NAT64 prefix (64:ff9b::/96) to the NAT Gateway:
aws ec2 create-route \
  --route-table-id <rtb-id> \
  --destination-ipv6-cidr-block 64:ff9b::/96 \
  --nat-gateway-id <nat-gateway-id>

Then run a test from a pod:

spec:
  containers:
    - name: busybox
      image: busybox
      command: ["sh", "-c", "ping ipv4-only-endpoint.com"]
```

We could then add a release note saying IPv6 is supported on EKS clusters.

jagan2221:
@rnishtala-sumo The draft doc I shared is just the same which would be added as something like ipv6.md .

Also,
We have enough EKS docs on how to configure IPv6 clusters and explaining the default egress ipv6->ipv4 in EKS clusters.

https://docs.aws.amazon.com/eks/latest/userguide/cni-ipv6.html#_ip_address_assignments
https://docs.aws.amazon.com/eks/latest/userguide/deploy-ipv6-cluster.html
Screenshot 2025-08-20 at 11 07 16 PM


Should we point to existing comprehensive AWS docs to customer / should we re-capture things from AWS docs from our side? I think recapturing AWS side things from our side could be a mutating thing and we can't adopt to the changes.
My idea is just to specify the workflow and direct customer to existing detailed vendor specific instead of maintaining those from our end. WDYT?

rnishtala-sumo:
In situations where we're asking a customer to go through manual steps before installing the helm chart, it helps to be specific. We're asking them to do the following

- Create a standard AWS NAT Gateway and enable DNS64 translation at the VPC level
- Add routing for 64:ff9b::/96 to the NAT Gateway.
- Let EKS pods use the default VPC DNS resolver

Asking them to run specific aws cli commands and a test using a pod ensures that the prerequisites are satisfied before the helm chart is deployed. QE can test using the same steps in the E2E.

jagan2221:
> In situations where we're asking a customer to go through manual steps before installing the helm chart, it helps to be specific. We're asking them to do the following
> 
> * Create a standard AWS NAT Gateway and enable DNS64 translation at the VPC level
> * Add routing for 64:ff9b::/96 to the NAT Gateway.
> * Let EKS pods use the default VPC DNS resolver
> 
> Asking them to run specific aws cli commands and a test using a pod ensures that the prerequisites are satisfied before the helm chart is deployed. QE can test using the same steps in the E2E.

@rnishtala-sumo 
Makes sense.
Created an EKS ipv6 cluster for testing . We don't need a NAT gateway , VPC-CNI plugin does the job. There was no additional things required.
1. Makes sure VPC and Subnets have both ipv4 and ipv6 CIDR's and subnets needs to  "auto-assign IPv6 address for nodes" setting enabled.
2. Install VPN-CNI plugin which has the ipv6-> ipv4 natting capability inbuilt.

Helm chart in installing properly and running fine with above changes. Will capture the steps to do above steps and share the doc.

jagan2221:
@rnishtala-sumo 
https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3977
Please review the doc PR.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: Done end to end testing and Raised the doc PR.

jagan2221:
@rnishtala-sumo Doc PR is throwing markdown lint errors . Checking that, will merge once resolved. Can you please provide a final approval for this PR?

tests/integration/features.go
rnishtala-sumo: @jagan2221 I think we still have this action item remaining?

jagan2221: @rnishtala-sumo This is not just about coredns_cache_hits_total. This metric is a required valid metric for other tests, only not required for ipv6 tests. 

coredns_forward_requests_total, coredns_proxy_request_duration_seconds_count, coredns_proxy_request_duration_seconds_sum are [added internally](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.13.0/tests/integration/internal/constants.go#L363C2-L363C25) inside WaitUntilExpectedMetricsPresent method which aren't required only for ipv6 tests.



In short,
1. coredns_cache_hits_total is not supported in ipv6 cluster, should we remove it only for ipv6 test as it is / make it optional for all tests?
2. coredns_forward_requests_total, coredns_proxy_request_duration_seconds_sum and coredns_proxy_request_duration_seconds_count are added implicitly by WaitUntilExpectedMetricsPresent() depending on k8's version.  So exposed a flag(addKubeVersionSpecificMetrics) to make this addition configurable. The idea behind this is, We pass expectedMetricsList to WaitUntilExpectedMetricsPresent, but it adds some metrics internally with no control/visibility to the caller.
In future if any other tests don't require these internally added metrics and only want to validate metrics passed in expectedMetrics list, this flag can be used.

rnishtala-sumo: If this is not required for ipv6 tests, we could follow existing the pattern and maintain a list of expected metrics for ipv6 tests here -  https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.13.0/tests/integration/internal/constants.go

This avoids changing the function signature for all tests. Also, maybe we don't need these tests to run for v1.29.0 anymore?

jagan2221: WaitUntilExpectedMetricsPresent add metrics implicitly without any checks. It is referring a list from constants only, but adding it for all tests which we don't expect.
And we can't make some metrics optional just because one test doesn't need those.
An offline sync would help closing this.

tests/integration/helm_ot_default_ipv6_test.go
```diff
@@ -0,0 +1,66 @@
...
+    ipv6ExcludedMetrics := []string{ //Add metrics that are not relevant for IPv6
+        "coredns_cache_hits_total",
```
rnishtala-sumo: is this the only metric that needs to be excluded for ipv6?

tests/integration/features.go
rnishtala-sumo: What are the metrics that shouldn't be included in ipv6 tests?

jagan2221: coredns_cache_hits_total, coredns_proxy_request_duration_seconds_sum, coredns_proxy_request_duration_seconds_count

And there is an old [comment](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/internal/stepfuncs/assess_funcs.go#L165) explaining why those metrics are added implicitly, which is not a good practice, but it's been present for a long time due to a limitation we have. I have just given a configurable flag to give control for a test to opt-in to these implicit metrics.

rnishtala-sumo: OK, in the interest of simplifying things, recommend removing these version dependent metrics from expected metrics here - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/internal/stepfuncs/assess_funcs.go#L167

And adding it to the optional/flaky list here - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/internal/constants.go#L309

This would remove the need for additional code changes to accomodate these 3 metrics and give us the opportunity to revisit what the expected metrics should be for specific k8s versions (including ipv6).

tests/integration/helm_ot_default_ipv6_test.go
jagan2221: Yes @rnishtala-sumo

tests/integration/helm_ot_default_ipv6_test.go
jagan2221: Done

tests/integration/helm_ot_default_ipv6_test.go
jagan2221: Removed this function

tests/integration/features.go
jagan2221: @rnishtala-sumo  Done, added in the flaky metrics list and removed implicit addition of metrics from WaitUntilExpectedMetricsPresent

tests/integration/helm_ot_default_namespaceoverride_test.go
jagan2221: Done

rnishtala-sumo (approved):
LGTM! lets ensure the new E2E tests run for this feature.

PR #3954, opened 2025-06-26
feat: Add service name and namespace to the k8sattributes processor

rnishtala-sumo:
Add service name and namespace to the k8sattributes processor

### Checklist

### DIFF
--- a/.changelog/3954.added.txt
+++ b/.changelog/3954.added.txt
@@ -1,0 +1,1 @@
+feat: Add service name and namespace to the k8sattributes processor
--- a/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
@@ -347,6 +347,8 @@ processors:
         - k8s.statefulset.name
         - k8s.namespace.name
         - k8s.node.name
+        - service.name
+        - service.namespace
     ## Has to be false to enrich metadata
     passthrough: false
     pod_association:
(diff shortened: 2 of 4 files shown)


tests/integration/values/values_common.yaml
```diff
@@ -123,6 +123,28 @@ otellogs:
...
+        startupProbe:
+          periodSeconds: 3
+          failureThreshold: 60
+    resources:
+      limits:
+        memory: 1Gi
+        cpu: 1000m
+      requests:
+        memory: 64Mi
+        cpu: 500m
```
rnishtala-sumo, line 147: resource changes to reduce test flakyness

tests/integration/values/values_common.yaml
```diff
@@ -123,6 +123,28 @@ otellogs:
...
+          timeoutSeconds: 10
+          failureThreshold: 3
+        readinessProbe:
+          initialDelaySeconds: 5
+          periodSeconds: 10
+          timeoutSeconds: 3
+          failureThreshold: 3
+        startupProbe:
+          periodSeconds: 3
+          failureThreshold: 60
```
ccressent, line 140: I feel like this should be in a separate PR, with the explanation as to why we need it.

chan-tim-sumo (approved):

tests/integration/values/values_common.yaml
rnishtala-sumo: Just ensuring that the otel logs collector doesn't fail readiness probes. This and the other resource changes are targeted to reduce test flakiness.

echlebek: Since it's an integration test values.yaml I agree with Raj here

PR #3976, opened 2025-08-20
feat(metrics): Added node labels to the upstream k8satrributeprocessor's config and updated integration tests 

samiura:
### Checklist

- [x] Documentation updated

Screenshot 2025-08-20 at 15 58 57
Screenshot 2025-08-20 at 15 59 08


deploy/helm/sumologic/README.md
```diff
@@ -497,7 +497,7 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
 | `metadata.logs.metricsLevel`                                                          | Metrics level for the OpenTelemetry Collector. Can be `basic`, `normal`, `detailed`.                                                                                                                                                                                                                       | `normal`                                                                                                                                                                             |
-| `metadata.logs.useSumoK8sProcessor`                                                   | Use the SumoLogic K8s processor                                                                                                                                                                                                                                                                            | `true`                                                                                                                                                                               |
```
rnishtala-sumo: I think we should change the default after marking the sumologic k8s processor as deprecated. We could do that in this release, gather some feedback from some of the early adopters and then proceed with changing the default processor.

tests/helm/testdata/goldenfile/metadata_logs_otc/debug.output.yaml
```diff
@@ -155,34 +155,42 @@ data:
         - _sourceName
         - _sourceHost
         - _collector
-      k8s_tagger:
+      k8sattributes:
```
rnishtala-sumo: All of these should remain unchanged if we don't change the default. Recommend adding a new test for testing node labels. See the k8s processor test for references below:

- https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/values/values_helm_k8s_processor.yaml

- https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/integration/helm_k8s_processor_test.go

deploy/helm/sumologic/README.md
samiura: So, In other words do not change the ReadME now? After these change, gather feedback and then when satisfied mark that as deprecated, correct?

tests/helm/testdata/goldenfile/metadata_logs_otc/debug.output.yaml
samiura: if we do not change the default, then it will still use the Sumo's attribute processor, right? 

I am aware of these test. It already check for all the node labels AFAIK.

rnishtala-sumo: In the test one can pick a processor to run the test with, using

```
metadata:
  metrics:
    useSumoK8sProcessor: false
```

We currently don't check for node labels, only the node name (this is also available as a pod label) - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/05f7a59b20637a8811ff1add108efb22406b08c2/tests/integration/features.go#L96

In the new test we probably want to check for labels only available at the node level, like node type, instance family etc.

deploy/helm/sumologic/README.md
rnishtala-sumo: yes we could mark the sumo processor as deprecated in the upcoming release ( we won't be maintaining it anymore). And change the default processor in a later release. So this would mean no change to the readme or values.yaml for the k8s processor yet.

tests/helm/testdata/goldenfile/metadata_logs_otc/debug.output.yaml
samiura: oh I see now what you mean! I have seen labels already in the logs. It should not be hard to add tests for
them, thanks for explaining. I will undo the changes ( wholesale test yaml changes as I thought we are deprecating sumok8s in one go )

samiura: @rnishtala-sumo The jobs are passing and nodel labels are seen as well.

https://github.com/SumoLogic/sumologic-kubernetes-collection/actions/runs/17687407604/job/50275524190

tests/integration/internal/constants.go
```diff
@@ -433,7 +439,7 @@ func InitializeConstants() error {
 	}
 
 	DefaultExpectedMetrics = []string{}
-	metricsGroupsWithOtelcol := append(DefaultExpectedMetricsGroups, DefaultOtelcolMetrics, LogsOtelcolMetrics, MetricsCollectorOtelcolMetrics)
+	metricsGroupsWithOtelcol := append(DefaultExpectedMetricsGroups, DefaultOtelcolMetrics, LogsOtelcolMetrics, MetricsCollectorOtelcolMetrics, NodeLabelsMetrics)
```
rnishtala-sumo: Adding these node metrics to all tests might break some tests which aren't using the upstream k8s processor.

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
```diff
@@ -63,6 +63,9 @@ k8sattributes:
       - key_regex: "(.*)"
         tag_name: "pod_labels_$$1"
         from: pod
+      - key_regex: "(.*)" 
+        tag_name: "node_labels_$$1"
```
chan-tim-sumo, line 67: is node labels only for k8sattributes? or should this also be in k8s tagger?

tests/integration/internal/constants.go
samiura: Good catch!

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
samiura: K8sattribute processor only provides node labels AFAIK and not sure about k8s tagger. K8s tagger is going to be phased out eventually.

tests/integration/features.go
```diff
@@ -94,22 +94,28 @@ func GetMetricsK8sattributes(expectedMetrics []string, metricsCollector MetricsC
...
+					"namespace":                           ctxopts.Namespace(ctx),
+					"node":                                internal.NodeNameRegex,
+					"pod_labels_app":                      deployment,
+					"pod_labels_pod-template-hash":        ".+",
+					"pod":                                 podList.Items[0].Name,
+					"replicaset":                          fmt.Sprintf("%s-.*", deployment),
+					"service":                             deployment,
+					"service.namespace":                   ctxopts.Namespace(ctx),
+					"k8s.node.uid":                        internal.UIDRegex,
+					"node_labels_beta.kubernetes.io/os":   internal.NodeLabelRegex,
```
rnishtala-sumo: Could these regexes be more specific? Using a predictable substring in the test will give is more visibility into what the expected values could look like.

deploy/helm/sumologic/conf/metrics/otelcol/processors.yaml
```diff
@@ -63,6 +63,9 @@ k8sattributes:
...
+        from: node
```
rnishtala-sumo, line 68: Could we add node labels to the logs configuration as well? The [k8sattributes](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) processor can be used to enrich logs, metrics and traces. At present, we use this processor for both logs and metrics.

samiura: Sure!

tests/integration/features.go
samiura: The idea was to re-purpose but sure it can be. F
or example os can be just linux|windows|osx etc

rnishtala-sumo:
@samiura LGTM! One remaining item is to add a changelog and call this a feature.

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -329,13 +329,19 @@ processors:
...
+          from: node
       annotations:
         - key_regex: "(.*)"
           tag_name: "pod_annotations_$$1"
           from: pod
         - key_regex: "(.*)"
           tag_name: "namespace_annotations_$$1"
           from: namespace
+        - key_regex: "(.*)" 
+          tag_name: "node_labels_$$1"
```
arpitasumologic: Why its not `node_annotations_$$1`? @samiura

samiura: @arpitasumologic Good catch! Thank you! It should be annotations.

samiura:
> @samiura LGTM! One remaining item is to add a changelog and call this a feature.

Added changelog!

PR #3977, opened 2025-08-22
docs(ipv6 support): add documentation for ipv6 cluster support

jagan2221:
Doc for supporting ipv6 clusters 

### Checklist

- [ ] Documentation updated

### DIFF
--- a/docs/ipv6.md
+++ b/docs/ipv6.md
@@ -1,29 +1,12 @@
 # IPV6 Support
 
-Supports EKS IPv6 clusters and any other k8’s cluster which has IPv6(Cluster)->IPv4(Internet) Egress communication
+Supports EKS IPv6 clusters and any other k8’s cluster type which has IPv6(Cluster)->IPv4(Internet) Egress communication
 
 
-
-**Pre-requisites for EKS Cluster**
-
-### 1. Configure IPv4 and IPv6 addresses for VPC and Subnets: 
-The VPC and subnets that your cluster is using must have an IPv6 CIDR block. They must also have an IPv4 CIDR block assigned to them. This is because, even if you only want to use IPv6, a VPC still requires an IPv4 CIDR block for performing IPv6->IPv4 Egress communication.
-
-```bash
-aws ec2 associate-vpc-cidr-block \
-    --vpc-id <your-vpc-id> \
-    --amazon-provided-ipv6-cidr-block \
-    --ipv6-cidr-block-network-border-group <your-network-border-group>
-```
-
-For detailed documentation on adding ipv6 and ipv4 CIDR to VPC. Please Refer, https://docs.aws.amazon.com/vpc/latest/userguide/add-ipv4-cidr.html
+## Pre-requisites for EKS Cluster
 
 
-
-
-
-
-### 2. Ensure Amazon VPC CNI version 1.10.1 or later is deployed in cluster.
+### 1. Ensure Amazon VPC CNI plugin v1.10.1 or later is deployed in cluster.
 IP prefix delegation and ENABLE_IPv6 settings must be enabled. If you already deployed VPC-CNI Plugin while creating the cluster, these will be enabled automatically. If you adding the plugin to an existing cluster, please add the plugin and ensure these settings are enabled.
 
 Create VPC-CNI addon:
@@ -31,9 +14,10 @@ Create VPC-CNI addon:
 aws eks create-addon --cluster-name my-cluster --addon-name vpc-cni --addon-version v1.20.0-eksbuild.1 \
 --service-account-role-arn arn:aws:iam::111122223333:role/AmazonEKSVPCCNIRole
 ```
-For more information, please refer: https://docs.aws.amazon.com/eks/latest/userguide/vpc-add-on-create.html
+For more information and different methods to create the plugin, please refer: https://docs.aws.amazon.com/eks/latest/userguide/vpc-add-on-create.html
 
 Adjust VPC-CNI configuration:
+
 ```bash
 aws eks update-addon \
   --cluster-name <your-cluster> \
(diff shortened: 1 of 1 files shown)


docs/ipv6.md
```diff
@@ -0,0 +1,84 @@
...
+
+### 2. Route table with ipv4 external route
+
+Make sure that VPC’s Route table has a route from IPv4(local) to Internet gateway. Ex. 0.0.0.0/0→igw-XXX (Internet Gateway)
+
+```bash
+aws ec2 create-route \
+    --route-table-id rtb-xxxxxxxx \
+    --destination-cidr-block 0.0.0.0/0 \
+    --gateway-id igw-xxxxxxxx
```
rnishtala-sumo, line 46: AFAIK, using only the IGW isn't the recommended way for ipv6 pods to reach out to ipv4 domains. A NAT is required to do the translation. Also, we need a solution that is generic enough that it potentially extends to other cloud environments in future. For example - [Azure](https://learn.microsoft.com/en-us/azure/nat-gateway/nat-overview)

I think we could get QE signoff on this doc. They can run this test on an ipv6 cluster with the test pod described.

At a high level I recommend a script that does the following:

```
- Allocates an Elastic IP → required for public NAT Gateway.
- Creates a NAT Gateway in a specified public subnet.
- Enables DNS64 on your VPC, so the AmazonProvidedDNS resolver will synthesize AAAA records for IPv4-only destinations.
- Adds 64:ff9b::/96 routes in the private subnets’ route tables, directing IPv6→IPv4 traffic to the NATGW
```
which is consistent with AWS documentation - https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-nat64-dns64.html (which also has the cli commands)

And then, to start with, claim support for ipv6 clusters on AWS.

jagan2221: @rnishtala-sumo 

Screenshot 2025-08-25 at 9 38 25 PM

We need DNS64 and NAT64 for standalone EC2 resources, but not for an EKS cluster as it handles it automatically using the CNI plugin which is also mentioned in EKS documentation and we have confirmed the same in our local EKS testing.
EKS doc--> https://docs.aws.amazon.com/eks/latest/userguide/cni-ipv6.html#_ip_address_assignments
This built-in NAT capability eliminates the need for [DNS64 and NAT64](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-nat-gateway.html#nat-gateway-nat64-dns64).

>  A NAT is required to do the translation. Also, we need a solution that is generic enough that it potentially extends to other cloud environments in future
-
    If the vendor supports it natively with a simple plugin, IMO, I don't think we should should complicate setup to the customer . If we find that Azure don't support such ease of use plugins(Could be supported in Azure Managed K8's too, need to check), then we can suggest how to do it in Azure accordingly when we claim Azure support.

rahul-babarwal: @jagan2221 , is this true for Azure and GCP as well?

jagan2221: We haven't validated how AKS and GCP provides ipv6-> ipv4 egress communication since we are considering/claiming support for EKS in this initial attempt. Although we mention that whichever cluster which has ipv6-> ipv4 egress communication supported is good to go.

fguimond: @rahul-babarwal - Would you mind checking with QE to see what their recommendation is on this topic?

rahul-babarwal: Sure @fguimond, we can check that. Thanks @jagan2221 for the details. We can create separate items for AKS and GCP

docs/ipv6.md
```diff
@@ -0,0 +1,123 @@
+# IPV6 Support
+
+Supports EKS IPv6 clusters and any other k8’s cluster type which has IPv6(Cluster)-to-IPv4(Internet) Egress communication enabled.
```
rnishtala-sumo:
```suggestion
Supports EKS IPv6 clusters and any other k8s cluster type which has IPv6(Cluster)-to-IPv4(Internet) Egress communication enabled.
```

docs/ipv6.md
```diff
@@ -0,0 +1,123 @@
...
+
+```bash
+aws eks create-addon --cluster-name my-cluster --addon-name vpc-cni --addon-version v1.20.0-eksbuild.1 \
+--service-account-role-arn arn:aws:iam::111122223333:role/AmazonEKSVPCCNIRole
+```
+
+For more information and different methods to create the plugin, please refer:
+https://docs.aws.amazon.com/eks/latest/userguide/vpc-add-on-create.html
+
+Adjust VPC-CNI configuration:
```
rnishtala-sumo, line 17: If the IPv6 cluster comes with the CNI with IPv6 enabled, then why do we need to adjust this configuration

docs/ipv6.md
```diff
@@ -0,0 +1,123 @@
...
+Both VPC-CNI plugin and NAT gateways provide IPv6-to-IPv4 communication. Please choose accordinly based on your cluster setup/preference.
+
+### 1. If you are using EKS's inbuilt VPC-CNI as the CNI plugin, then please ensure below
+
+#### 1.1. Ensure Amazon VPC CNI plugin v1.10.1 or later is deployed in cluster
+
+IP prefix delegation and ENABLE_IPv6 settings must be enabled. If you already deployed VPC-CNI Plugin while creating the cluster, these will
+be enabled automatically. If you adding the plugin to an existing cluster, please add the plugin and ensure these settings are enabled.
+
+Create VPC-CNI addon:
```
rnishtala-sumo: Do we even need to ask the customer to do this manually. It isn't possible to create an EKS cluster without a CNI.

docs/ipv6.md
```diff
@@ -0,0 +1,123 @@
...
+      "ENABLE_PREFIX_DELEGATION": "true",
+      "WARM_PREFIX_TARGET": "1"
+    }
+  }'
+```
+
+For more information and different methods to update plugin settings, please refer
+https://docs.aws.amazon.com/eks/latest/userguide/updating-an-add-on.html
+
+#### 1.2. Route table with ipv4 external route
```
rnishtala-sumo: If this supported out of the box, Do we need to add this route manually?

docs/ipv6.md
```diff
@@ -0,0 +1,123 @@
...
+
+Adjust VPC-CNI configuration:
+
+```bash
+aws eks update-addon \
+  --cluster-name <your-cluster> \
+  --addon-name vpc-cni \
+  --configuration-values '{
+    "env": {
+      "ENABLE_IPv6": "true",
```
rnishtala-sumo, line 25: Are these enabled by default in an IPv6 cluster?

docs/ipv6.md
jagan2221: This is just basic routing which every VPC should have to connect to internet. ie) VPC to Internet.
There is a chance that Customer might only be having ipv6 route to internet gateway, **This is to ensure that ipv4 traffic from your cluster can reach internet** since even if natting is default, ipv4 route to internet is required.

docs/ipv6.md
jagan2221: If customer added VPC plugin at the time of ipv6 cluster creation, all settings related to ipv6 are enabled by default in the plugin. But if customer have added the plugin after cluster creation, these settings needs to be adjusted. We have mentioned the same [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3977/files#diff-ba9b3951a5956673563806d35a3fccac724ff26e3d2e69a7fd72364e91b97683R14)

docs/ipv6.md
jagan2221: Agreed, we can remove VPC plugin add-on creation step and could ask customer to only ensure the required settings  "IP prefix delegation and ENABLE_IPv6 settings" for the plugin are configured properly as these settings will not be enabled automatically if customer adds the plugin after cluster creation. Though as mentioned above, if customer added the plugin at the time of cluster creation, no need to adjust the settings.

docs/ipv6.md
jagan2221: This is only for those who have added the plugin after ipv6 cluster creation. Mentioned [here](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3977/files#diff-ba9b3951a5956673563806d35a3fccac724ff26e3d2e69a7fd72364e91b97683R14).
So, those who have added the plugin while creating cluster does not need to adjust any settings.

docs/ipv6.md
```diff
@@ -0,0 +1,112 @@
...
+aws ec2 create-route \
+    --route-table-id rtb-xxxxxxxx \
+    --destination-cidr-block 0.0.0.0/0 \
+    --gateway-id igw-xxxxxxxx
+```
+
+For more information and different methods to add route, please refer
+https://docs.aws.amazon.com/vpc/latest/userguide/create-vpc-route-table.html#AddRoutes
+
+#### 2.2. If your worker nodes are in private subnet or you are not using AWS VPC-CNIusing custom CNI like celium, then please ensure you have provisioned a NAT gateway.
```
rnishtala-sumo:
```suggestion
#### 2.2. If your EKS worker nodes are in private subnet (recommended by AWS) or you are using a custom CNI like Cilium, then please ensure you have provisioned a NAT gateway.
```

rnishtala-sumo (approved):
LGTM, thanks!

PR #3993, opened 2025-09-19
feat(otellogs): add hostNetwork and dnsPolicy configuration

azamorano:
Allows for configurable `hostNetwork` and `dnsPolicy`. Related with https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/3940 

### Checklist

### DIFF
--- a/tests/helm/logs_test.go
+++ b/tests/helm/logs_test.go
@@ -587,3 +587,65 @@ otellogs:
 
 	require.Equal(t, "50%", logsCollectorDaemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable)
 }
+
+func TestCollectorDaemonsetHostNetwork(t *testing.T) {
+	t.Parallel()
+
+	valuesYaml := `
+otellogs:
+  daemonset:
+    hostNetwork: true
+    dnsPolicy: ClusterFirstWithHostNet
+`
+	templatePath := "templates/logs/collector/otelcol/daemonset.yaml"
+
+	renderedTemplate, err := RenderTemplateFromValuesStringE(t, valuesYaml, templatePath)
+	require.NoError(t, err)
+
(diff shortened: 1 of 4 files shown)


rnishtala-sumo:
@azamorano thank you for your contribution! Recommend that we also add an Integration test for this change to validate that the pods in the logs daemonset startup without issues. This should also test whether a pod would need associated clusterRole changes to run on this mode and that the logs are being received by sumologic.

echlebek (changes requested):
It appears that this change has lead to one of the integration tests breaking

azamorano:
@echlebek sorry for the late response. I did not have much time to take a look to this.
I'm failing to understand what I might have broken with my change in the current integration tests.
I'm trying to run it locally and it works for me. Please find attached the snippet with the output of my test.
[test.log](https://github.com/user-attachments/files/23521637/test.log)
Could it be something transient?

azamorano:
Hi @echlebek did you have a chance to take a look to my comment above?

tests/helm/logs_test.go
```diff
@@ -587,3 +587,65 @@ otellogs:
 
 	require.Equal(t, "50%", logsCollectorDaemonset.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable)
 }
+
+func TestCollectorDaemonsetHostNetwork(t *testing.T) {
```
rnishtala-sumo, line 591: Could we plug into the existing template tests? I think we'll need to define the input and output files. See - https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/tests/helm/README.md

rnishtala-sumo (commented):
Added a comment about the template tests. I'd also recommend creating an issue or updating the description for this PR, to help understand the motivation to enable hostNetwork.

PR #4015, opened 2025-10-15
chore: set GODEBUG=netdns=go in most environments

echlebek:
This commit sets GODEBUG=netdns=go in deployments, jobs and statefulsets, selecting the netgo DNS resolver instead of the cgo resolver that is made available by the operating system. Previously, we relied on the collector to be built with only the netgo resolver. However, future releases of the collector will include both DNS resolvers, and a known issue in k8s collection requires the use of the netgo resolver.

### DIFF
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -4,6 +4,8 @@ version: 4.17.0
 appVersion: 4.17.0
 description: A Helm chart for collecting Kubernetes logs, metrics, traces and events into Sumo Logic.
 type: application
+env:
+  GODEBUG: netdns=go  # use the netgo DNS resolver for k8s
 keywords:
   - monitoring
   - logging
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -4,8 +4,6 @@ version: 4.17.0
 appVersion: 4.17.0
 description: A Helm chart for collecting Kubernetes logs, metrics, traces and events into Sumo Logic.
 type: application
-env:
-  GODEBUG: netdns=go # use the netgo DNS resolver for k8s
 keywords:
   - monitoring
   - logging
--- a/deploy/helm/sumologic/Chart.yaml
+++ b/deploy/helm/sumologic/Chart.yaml
@@ -5,7 +5,7 @@ appVersion: 4.17.0
 description: A Helm chart for collecting Kubernetes logs, metrics, traces and events into Sumo Logic.
 type: application
 env:
-  GODEBUG: netdns=go  # use the netgo DNS resolver for k8s
+  GODEBUG: netdns=go # use the netgo DNS resolver for k8s
 keywords:
   - monitoring
   - logging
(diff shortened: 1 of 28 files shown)


deploy/helm/sumologic/Chart.yaml
```diff
@@ -4,6 +4,8 @@ version: 4.17.0
 appVersion: 4.17.0
 description: A Helm chart for collecting Kubernetes logs, metrics, traces and events into Sumo Logic.
 type: application
+env:
```
ccressent: I don't see an `env` key for [Helm Charts](https://helm.sh/docs/topics/charts/#the-chartyaml-file). Are you sure this works?

echlebek: No, and I think I probably got this wrong. It looks like it needs to be specified as a top-level key in all values.yaml files.

echlebek: ...and then threaded into the deployment templates

ccressent: I don't see the need to involve Helm. If it's indeed "a known issue in k8s collection requires the use of the netgo resolver", making it non-negotiable, then you could just "hardcode" the environment variable in the relevant deployments [e.g. thereabouts for the collection pods](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml#L120)

echlebek:
I need to fix up the goldenfiles for all these objects, but the unexpected inclusion from the test's POV does prove that the changes are effectful.

ccressent:
Are you not missing at least the daemonset pods? i.e. I was expecting [this section](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/netgo-resolver/deploy/helm/sumologic/templates/logs/collector/otelcol/daemonset.yaml#L120-L129) to be touched too. There might be others.

Since the equivalent effect used to be part of the Sumologic collector build process, why not add this environment variable to the [Sumologic collector Docker images](https://github.com/SumoLogic/sumologic-otel-collector-containers/tree/main/dockerfiles)? That way you know you won't be missing a spot in any of the k8s manifests and you will also fix it for the images themselves, not just their use within a k8s environment.

Also, keep in mind that the images used in those Daemonsets, Statefulsets, ... are configurable in the chart's values. They default to using the. Sumologic collector images we release, but in principle one could set them to something else that might not want your enforced variable. To me, this points to setting the variable in the container image itself being a better approach instead of "blindly" in the k8s manifest. A single `ENV` instruction [in the relevant Dockerfiles](https://github.com/SumoLogic/sumologic-otel-collector-containers/tree/main/dockerfiles) should do the trick.

echlebek:
@ccressent Outside of K8s there's no reason not to use the cgo resolver, but I don't mind the approach. It would prevent any misses like you say. I'll look into doing that instead.

ccressent:
Can we close this?

jagan2221:
I think this will be required when linux otel binary is compiled as both cgo and netgo compatible. but now it's only netgo.
If future if we make it run on both, we can close this pr and add an [env var in docker](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/4015#issuecomment-3462933447) to point it to netgo mode as cyril says. For now,  closing this PR.

PR #4136, opened 2026-04-09
feat(config): add retry config of sumologic exporter for improving reliability, reduce data drop

jagan2221:
As per findings and improvements put togther in this [doc](https://docs.google.com/document/d/1IAhkPmozOz_eb6CUTRlsprDWsKLgMFl_0-8DmKoT2Ng/edit?tab=t.0#heading=h.7vz84qepgsyc)

making the changes for retry settings, so that sumologic exporter doesn't drop data in case of transient backend outages.

- [ ] Documentation updated


Gourav2906:
can we directly change this as default to upstream ?

jagan2221:
> can we directly change this as default to upstream ?

Good suggestion, let me close this and add this as default in upstream.

PR #4129, opened 2026-04-07
feat(OSC-1274): Add export level batching

ravishankar-sumo:
### Checklist

- [x] Changelog updated or skip changelog label added
- [x] Documentation updated
- [ ] Template tests added for new features
- [ ] Integration tests added or modified for major features

We have introduced two internal flags
1. Before rolling this out externally we wanted to test this in sumologic clusters and that can done only after release hence we wanted to have the additional check `useExporterBatching` to switch the behaviour.
2. The queue_size is a critical change for the customer who has configured their own exporters we wanted to use `customBatchingConfigured` flag as a acknowledgement from the user which can be removed in the next major release of the helm.

In an ideal default scenario the current PR is expected to retain the existing behaviors of all observability signals. When the flags are enabled the new implementation should take effect. 

Related Docs PR: https://github.com/SumoLogic/sumologic-documentation/pull/6683

PR #4218, opened 2026-06-01
feat: add single-layer metrics pipeline using forward connector

dhruv-shah-sumo:
Merge the collector and metadata layers into a single pod behind a feature flag (singleLayerPipeline.enabled). Uses a forward connector to bridge two logical pipelines, preserving config.merge backward compatibility.

https://sumologic.atlassian.net/wiki/spaces/PM1/pages/3124297811/K8s+Metric+pipeline+compaction+2+layered+to+1+layered

### Checklist

- [ ] Documentation updated


deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -106,8 +107,246 @@ service:
...
+    - pprof
+{{- if .Values.metadata.persistence.enabled }}
+    - file_storage
+{{- end }}
+  pipelines:
+    ## Pipeline 1: Collector stage — scraping + light processing → forward connector
+    metrics/collector:
+      receivers: [prometheus]
+      processors:
+        - filter/drop_stale_datapoints
```
himanshu130700, line 200: should we have the memory limiter here also, so that it can check the memory burst before entering in the pipeline?

deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml
```diff
@@ -1,4 +1,7 @@
 {{- if and (eq (include "metrics.otelcol.enabled" .) "true") .Values.sumologic.metrics.collector.otelcol.enabled }}
+{{- if and .Values.sumologic.metrics.collector.otelcol.singleLayerPipeline.enabled .Values.metadata.metrics.config.override }}
+{{- fail "metadata.metrics.config.override cannot be used with singleLayerPipeline.enabled. Use metadata.metrics.config.merge instead, or disable single-layer mode." }}
```
jagan2221, line 3: Can we move this check to deploy/helm/sumologic/templates/checks.txt so that it'll be a central place to maintain helm value pre-checks?

jagan2221:
Some initial thoughts: @dhruv-shah-sumo 
1. Helm template check for resource bump like CPU and memory . If not overridden, we can increase the limit . if customer has overridden , need to guide them to increase it accordingly.
2. Queue size doesn't have impact since we have 2 logical pipelines retained in single collector, but i think underlying PVC would need to be resized to accomadate data from both pipelines. Pls verify.

Will do another pass tmrw.

himanshu130700:
I think few single-layer test are missing:

- No golden test for dropHistogramBuckets+ allowHistogramRegex (covered only in metadata_metrics_otc/allow_histograms for 2-layer).
- No coverage for sourceType: http routing/sub-pipelines in single-layer (2-layer equivalent: metadata_metrics_otc/custom_routing_connector).
- No test for enableDefaultFilters: true in single-layer (2-layer equivalent: metadata_metrics_otc/filtered_app_metrics).

dhruv-shah-sumo:
> 2. Queue size doesn't have impact since we have 2 logical pipelines retained in single collector, but i think underlying PVC would need to be resized to accomadate data from both pipelines. Pls verify.
> 
> Will do another pass tmrw.

Why would underlying PVC need to be resized? I agree that memory/cpu has to be increased since the data egress would become costly and need to keep it in memory until it is send to Sumo. PVC is mainly used when data cannot be sent to Sumo after several retries. Until then the data remains in the in-memory queue, right?

deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml
dhruv-shah-sumo: ACK.

dhruv-shah-sumo:
> 1. Helm template check for resource bump like CPU and memory . If not overridden, we can increase the limit . if customer has overridden , need to guide them to increase it accordingly.

Do you mean, we should increase it in our default values in case single layer pipeline is selected?

jagan2221:
> > 2. Queue size doesn't have impact since we have 2 logical pipelines retained in single collector, but i think underlying PVC would need to be resized to accomadate data from both pipelines. Pls verify.
> > 
> > Will do another pass tmrw.
> 
> Why would underlying PVC need to be resized? I agree that memory/cpu has to be increased since the data egress would become costly and need to keep it in memory until it is send to Sumo. PVC is mainly used when data cannot be sent to Sumo after several retries. Until then the data remains in the in-memory queue, right?

Metadata layer now has 10Gb default PVC(metadata.persistence.size) , but collector layer doesn't have a PVC and it seems it stores data(queue) on pod filesystem itself. I thought that if collector has a PVC too(Like 1GB size), then we should keep merged pipeline's PVC size as 11GB. But since metadata layer is the only one which has PVC, I see you already[ migrated PVC](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/dhruv/single-layer-metrics/deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml#L191) to collector layer. This is fine!

jagan2221:
> > 1. Helm template check for resource bump like CPU and memory . If not overridden, we can increase the limit . if customer has overridden , need to guide them to increase it accordingly.
> 
> Do you mean, we should increase it in our default values in case single layer pipeline is selected?

Yes, let's keep it 1.5x as discussed in review discussion if customer hasn't overridden.
 let's say customer has a large cluster and he has increased memory for metadata metrics pods to 4GB. Now collector pods which are going to handle this work needs their RAM to be increased proportionally right? We can add a helm template upgrade test to capture if resources are overridden ,if yes, let customer to increase resources of collector layer  accordingly.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: These 2 pipelines are connected using forward connector. we now have memory limiter in 2nd pipeline already, i think we shouldn't add another one here which creates 2 memory limiters in a single logically connected pipeline. May be we can think of moving the memory limiter from metadata pipeline to collector pipeline ? @dhruv-shah-sumo

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -106,8 +107,246 @@ service:
...
+        read_timeout = "30s"
+        write_timeout = "30s"
+        service_address = ":9888"
+        data_format = "prometheusremotewrite"
+        paths = [
+{{ include "metric.endpoints" . | indent 10 }}
+        ]
+{{- else }}
+  ## Configuration for upstream Prometheus Remote Write Receiver (v2 only)
+  prometheusremotewrite:
```
jagan2221, line 344: Prometheus remote write endpoint which users pointed to will change now.
As earlier, they would have pointed to metadata metrics service, now they should point to metrics collector service.

Old flow: http://$(METADATA_METRICS_SVC).$(NAMESPACE).svc.cluster.local.:9888

New single pipeline flow: http://$(METRICS_COLLECTOR_SVC).$(NAMESPACE).svc.cluster.local.:9888

To get metrics collector service endpoint, 
kubectl get svc -n {namespace}
Look for *-metrics-collector service with 9888 TCP port listed under Ports column.
Modify your prometheus rw to write to this new endpoint.

We need to add this in a KB doc. As we cannot catch this via any feature flags customer has enabled, the remote write url will always change if customer using either telegraf/prometheusremotewrite receiver. We can put this in a KB doc along with all other changes and ask customer to ACK similarly we did for helm v5 upgrade > important changes.

deploy/helm/sumologic/templates/metrics/common/service.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if eq (include "metrics.otelcol.enabled" .) "true" }}
+{{- if and (eq (include "metrics.otelcol.enabled" .) "true") (ne (include "metrics.collector.singleLayerPipeline.enabled" .) "true") }}
```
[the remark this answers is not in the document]
jagan2221: Important: Seems to be a valid concern. 
Remote write prxy forwards to [otlp 4318](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/main/deploy/helm/sumologic/templates/metrics/remote-write-proxy/service.yaml#L15) port here, but it is no-longer present.

Also remote write proxy made sense when collector and metadata layers were separate. Now once they're unified, i think we can disable remote write proxy and relevant resources if singleLayer pipeline is enabled. 

@dhruv-shah-sumo

deploy/helm/sumologic/templates/metrics/common/hpa.yaml
jagan2221, line 15: Default value of minReplica is 3 here. But default for metrics collector (sumologic.metrics.collector.otelcol.autoscaling.minReplicas) is 1. Need to bump min replica for metrics collector layer . Similarly other resources, overridden values needs to be proportionally increased.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
dhruv-shah-sumo: Noted here in comments: https://sumologic.atlassian.net/wiki/spaces/PM1/pages/3124297811/K8s+Metric+pipeline+compaction+2+layered+to+1+layered?focusedCommentId=3154804876

Will be working on doc changes once the PoC is over.

deploy/helm/sumologic/templates/metrics/common/hpa.yaml
dhruv-shah-sumo: Done. minReplicas: 3 when single layer pipeline is enabled. 

For other cpu/memory resources, it is a bit tricky. For example, if the user has overwritten default configs by mentioning the cpu/memory, the change can again 1.5X that value. My suggestion is to mention this up-scaling cpu/memory (user has to add overwriting config for cpu/mem in case enabling single-layer metric pipeline) in a release doc and when we switch the singlePipelineLayer flag true by default, we can increase the resource in values.yaml as default increase.

dhruv-shah-sumo:
> > > 1. Helm template check for resource bump like CPU and memory . If not overridden, we can increase the limit . if customer has overridden , need to guide them to increase it accordingly.
> > 
> > 
> > Do you mean, we should increase it in our default values in case single layer pipeline is selected?
> 
> Yes, let's keep it 1.5x as discussed in review discussion if customer hasn't overridden. let's say customer has a large cluster and he has increased memory for metadata metrics pods to 4GB. Now collector pods which are going to handle this work needs their RAM to be increased proportionally right? We can add a helm template upgrade test to capture if resources are overridden ,if yes, let customer to increase resources of collector layer accordingly.

I tried implementing this but it is not possible to identify if resource values are overwritten or not. Helm only gets the final set value. Let me know if you have any suggestion for this.

deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml
jagan2221: FYI @dhruv-shah-sumo

jagan2221:
> > > > 1. Helm template check for resource bump like CPU and memory . If not overridden, we can increase the limit . if customer has overridden , need to guide them to increase it accordingly.
> > > 
> > > 
> > > Do you mean, we should increase it in our default values in case single layer pipeline is selected?
> > 
> > 
> > Yes, let's keep it 1.5x as discussed in review discussion if customer hasn't overridden. let's say customer has a large cluster and he has increased memory for metadata metrics pods to 4GB. Now collector pods which are going to handle this work needs their RAM to be increased proportionally right? We can add a helm template upgrade test to capture if resources are overridden ,if yes, let customer to increase resources of collector layer accordingly.
> 
> I tried implementing this but it is not possible to identify if resource values are overwritten or not. Helm only gets the final set value. Let me know if you have any suggestion for this.

Agree, But if anyone enables single layer pipeline , things like prometheus remote write receiver url change and this resource adjustments silently go ignored.

Let's add a check in deploy/helm/sumologic/templates/checks.txt which capture if singlelayerPipeline flag is enabled, We can stop upgrade and redirect Users to help doc page similar to our k8s upgrade docs. There we can document all things customer need to change manually and then let him add an ACK flag to let him upgrade. This forces anyone moving to single layer pipeline to read all the concerning changes and act before proceeding to upgrade. WDYT? @dhruv-shah-sumo .
This is not a blocker for this PR,we can add it in a separate PR. ACK flag similar to customBatchingConfigured added in [this](https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/4129)  PR.

jagan2221 (approved):

PR #4233, opened 2026-06-15
chore(SUMO-286652): Remove useSumoK8sProcessor and k8s_tagger

ravishankar-sumo:
Remove the legacy k8s tagger https://sumologic.atlassian.net/browse/SUMO-286652

### Checklist

- [ ] Documentation updated


jagan2221:
For someone who is still using sumo k8s tagger, this is a silent migration to upstream k8sprocessor which starts adding nodelabels additionally . In this [thread](https://sumologic.slack.com/archives/C09MVFED44S/p1781607192235559?thread_ts=1781593668.008699&cid=C09MVFED44S) we saw similar problem due to node labels started coming in.

I think along with this change, it's better to also introduce extractNodeLabels(which can true by default) flag to logs pipeline's k8sattributes processor too , similarly how we already have for metrics  to give user the control to disable it if they face any such issues.
https://sumologic.atlassian.net/browse/SUMO-286974

@ravishankar-sumo

dhruv-shah-sumo:
A lot of our customers are on 4.x release where k8s_tagger was the default processor. Some of them would have configured k8s_tagger in their config.merge pipeline and would only about this change when the changes are applied and otel-collectors fail. 

Can you look for a way to fail in such cases during helm upgrade/installation? That would save the customers from doing such blind upgrades. Erroring out with a doc link would be ideal.

wdyt @jagan2221 ?

jagan2221:
> A lot of our customers are on 4.x release where k8s_tagger was the default processor. Some of them would have configured k8s_tagger in their config.merge pipeline and would only about this change when the changes are applied and otel-collectors fail.
> 
> Can you look for a way to fail in such cases during helm upgrade/installation? That would save the customers from doing such blind upgrades. Erroring out with a doc link would be ideal.
> 
> wdyt @jagan2221 ?

Yes, I think it's possible to have a template check to see if k8s_tagger is added into the pipeline using config.merge / config.override , and throw an error blocking upgrade. @ravishankar-sumo

deploy/helm/sumologic/templates/checks.txt
```diff
@@ -114,6 +114,14 @@
 {{- end -}}
 {{- end -}}
 
+{{/* Helper: fail if the k8s_tagger is configured in the given processors map.
+     Call with: (dict "processors" <map> "path" <string>) */}}
+{{- define "check.processors.k8s_tagger" -}}
```
dhruv-shah-sumo, line 119: Thanks for adding the checks. I am assuming that these checks are tested locally.

dhruv-shah-sumo (approved):

deploy/helm/sumologic/templates/checks.txt
ravishankar-sumo: Yeah !!

Screenshot 2026-06-19 at 4 48 26 PM

PR #4260, opened 2026-07-06
feat: Use sumologic extension for authentication and move out of Hosted collector and sources.

jagan2221:
**Changes**:
1. This PR adds changes to support using sumologic extension for authentication , so pods can register and send data to open-receiver directly without using http/otlp source url.
2. Added Terraform changes to create installation token needed for sumo extension and add it in k8s secret for pods to refer later. 
2. Sourceless mode will be disabled by default, because on upgrade it should be enabled automatically. Also existing customers re-installing should not get it. So let's keep it disabled now, In next major helm release, we can try to enforce it to true. 

4. We don't delete any existing sources/hosted collector for Upgrades. For upgrades, if customers opts for sourceless mode, data flow goes via new mode and existing entities will be left as it is unless they explicitly enable cleanup flag. 
sumologic.cleanupHostedCollector - Enabling this flag this delete the hosted collector from Sumo. 

This way, there won't be any accidental deletion of actively used sources when user switches to sourceless mode.

5. All pods which send data to sumologic will now appear in OT collector page once FF is enabled.

More details: https://sumologic.atlassian.net/wiki/spaces/PM1/pages/3183181929/UI+and+usability+flow+changes+-+k8s+sourceless+model

### Testing

1. New install with existing installation command. -> Sourceless mode disabled.
2. New install with sourceless mode FF enabled -> Sourceless mode enabled -> No Hosted collector created and collector pods registered and start appearing in OT collectors page.
3. Upgrade with sourceless mode disabled -> no change
4. Upgrade with sourceless mode enabled -> Hosted collector and sources remain in UI -> But collection pods register and move to new sourceless flow.
5. Upgrade with sourceless mode disabled and sources cleanup enabled -> Hosted collector deleted from UI -> Collection pods moved to sourceless flow.
6. New install with metrics single layer pipeline enabled and now metrics collector pods start registering and sending data using sourceles mode.

ITs: Sumologic mock does not have register api capability now, so we cannot add IT's, added golden test alone. E2E's would be fine to capture testing changes since it's just the data upload endpoint that's getting changed. 

You can add a changelog entry by running `make add-changelog-entry`
See [/docs/dev.md] for more details
-->

- [ ] Documentation updated


deploy/helm/sumologic/conf/setup/sources.tf.json
```diff
@@ -1,3 +1,7 @@
+{{- if and .Values.sumologic.sourcelessMode .Values.sumologic.cleanupHostedCollector }}
+{{- /* sourcelessMode + cleanup: remove HTTP sources from Terraform management so they are destroyed on apply. */ -}}
+{{- toPrettyJson (dict "locals" (dict "source_configs" dict "sources" dict)) | nindent 2 }}
+{{- else }}
```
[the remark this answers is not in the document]
jagan2221: ACK,

deploy/helm/sumologic/conf/setup/resources.tf
```diff
@@ -1,16 +1,27 @@
 resource "sumologic_collector" "collector" {
+  count       = (var.use_extension && var.cleanup_hosted_collector) ? 0 : 1
   name        = var.collector_name
   description = format("Sumo Logic Kubernetes Collection\nversion: %s", var.chart_version)
   fields      = var.collector_fields
```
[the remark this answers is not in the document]
jagan2221: ACK, changes made

deploy/helm/sumologic/README.md
```diff
@@ -29,6 +29,9 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
+| `sumologic.cleanupHostedCollector`                                                    | When `sourcelessMode` is enabled, setting this to `true` will delete the hosted collector and all HTTP/OTLP sources managed by the setup job. Defaults to `false` so existing sources are preserved and can be cleaned up later by flipping this flag.                                                                                                                                               | `false`                                                                                                                                                                              |
```
dhruv-shah-sumo: if set to true, this will delete the hosted collector and the underlying sources, right? If yes, it would be worth adding a note that manually created source under the collector will be deleted as well. 

If the user was using those custom source URL, is there any other option for them?

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -160,12 +161,66 @@ else
     echo "Please refer to https://www.sumologic.com/help/docs/manage/fields/ to manually create the fields after you have removed unused fields to free up capacity."
 fi
 
+# Delete hosted collector (and all its sources) via API when cleanup is requested.
+# Terraform cannot destroy resources it cannot import, so we use the API directly.
+function delete_hosted_collector() {
```
dhruv-shah-sumo, line 169: if setupEnabled: false then this functionality is not available, right? How does users with argo use this functionality?

deploy/helm/sumologic/README.md
```diff
@@ -29,6 +29,9 @@ The following table lists the configurable parameters of the Sumo Logic chart an
...
+| `sumologic.sourcelessMode`                                                            | If enabled, uses the SumoLogic extension for auth and endpoint discovery instead of HTTP/OTLP source URLs. Terraform will auto-create an installation token. OTLP/HTTP sources are not created when this is enabled.                                                                                                                                                                                 | `false`                                                                                                                                                                              |
```
dhruv-shah-sumo, line 32: Since we are removing the hosted collector if user opts in, how does the pre-ingestion processing (some users are using it to configure data filtering). Do we plan to deprecate that feature?

deploy/helm/sumologic/README.md
jagan2221: They can opt not to use this cleanup flag. (or) Create a separate hosted collector manually and use that instead and then opt to cleanup. I think if some customer is using custom sources under the k8s hosted collector, they will prefer to continue use it rather than opting for cleanup.

jagan2221: Changed the text, to mention that "all http/otlp sources under the Hosted collector will be deleted"

deploy/helm/sumologic/conf/setup/setup.sh
jagan2221: yes, Those who have setup disabled, they usually create the Hosted collector and source manually and added the source url as k8s secret for the pods to refer. So, for deletion of that and moving to sumo extension, they need to do manually. For switching to Sourceless mode, they need to re-enable setup job. 

Check for setup job when sourceless mode is enabled -> https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/4260/changes#diff-9f23b91eb1489b92812c1f7545a98868f3b856262dfa7eca388250e8076d2edaR3

deploy/helm/sumologic/README.md
jagan2221: Yes, Pre-ingestion processing in K8s Hosted collector was not an intended feature but an add-on which came along since we are using a Hosted collector. Now we are losing that feature, but the right place to do it is via Otel processors in the pipeline. 

We will  add a separate section in helm v6 upgrade doc when we enable this feature to let customers know all the detailed impacts of this.

deploy/helm/sumologic/conf/metrics/otelcol/config.yaml
```diff
@@ -19,6 +19,17 @@ extensions:
   health_check:
     endpoint: {{ template "sumologic.healthEndpoint" . }}
   pprof: {}
+{{- if .Values.sumologic.sourcelessMode }}
+  sumologic:
+    installation_token: ${SUMOLOGIC_INSTALLATION_TOKEN}
```
dhruv-shah-sumo, line 24: For how long is this token valid? Can it be rotated externally or does it get renewed whenever the setup is re-deployed?

deploy/helm/sumologic/conf/metrics/otelcol/exporters.yaml
```diff
@@ -72,7 +72,9 @@ sumologic/sumologic-mock-http:
 sumologic/default:
   client: {{ include "sumologic.sumo_client" . }}
   metric_format: {{ include "metrics.otelcol.exporter.format" . }}
+{{- if not .Values.sumologic.sourcelessMode }}
   endpoint: {{ include "metrics.otelcol.exporter.endpoint" . }}
```
dhruv-shah-sumo, line 76: How do we plan to support config.merged exporters? Mention the guidelines in a doc for users to follow?

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -160,12 +161,66 @@ else
...
+    # Only import sources when collector exists.
+    if terraform import 'sumologic_collector.collector[0]' "${SUMOLOGIC_COLLECTOR_NAME}"; then
+        jq -r '.resource[] | to_entries[] | "\(.key) \(.value.name)"' sources.tf.json | while read -r resource_name source_name; do
+            terraform import "sumologic_http_source.${resource_name}" "${SUMOLOGIC_COLLECTOR_NAME}/${source_name}"
+        done || true
+    fi
+fi
+
+# Import existing installation token if extension mode is enabled (prevents recreation on upgrades)
+function import_installation_token() {
```
dhruv-shah-sumo, line 214: Is it possible to not run setup job and manually provide the installation token from the UI? Users who have disabled the setupJob might be hesitant to change the config.

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -160,12 +161,66 @@ else
...
+    local TOKEN_NAME="kubernetes-collection-${SUMOLOGIC_COLLECTOR_NAME}"
+    local RESPONSE TOKEN_ID
+    RESPONSE="$(curl -XGET -s \
```
dhruv-shah-sumo: Please verify that we are not priting this entire curl in the setup job's log. We might end up leaking secrets in the logs.

deploy/helm/sumologic/conf/setup/setup.sh
```diff
@@ -160,12 +161,66 @@ else
...
+        -u "${SUMOLOGIC_ACCESSID}:${SUMOLOGIC_ACCESSKEY}" \
+        "${SUMOLOGIC_BASE_URL}v1/tokens")"
+    local JQ_OUTPUT
+    JQ_OUTPUT=$(jq -r ".data[] | select(.name == \"${TOKEN_NAME}\") | .id" <<< "${RESPONSE}")
```
dhruv-shah-sumo: What happens when response is 400/500? Possible to add a few retries since this is critical to collection deployment?

deploy/helm/sumologic/templates/checks.txt
```diff
@@ -1,3 +1,13 @@
+{{/* Check that sourcelessMode is not enabled when setupEnabled is false */}}
+{{- if and .Values.sumologic.sourcelessMode (not .Values.sumologic.setupEnabled) -}}
+{{- fail "\nsumologic.sourcelessMode requires sumologic.setupEnabled=true.\nThe setup job creates the installation token needed for sourceless authentication." -}}
```
dhruv-shah-sumo: As mentioned above, is it possible to generate accessToken from the Sumo UI manually and provide it during installation?

deploy/helm/sumologic/values.yaml
```diff
@@ -15,6 +15,19 @@ sumologic:
 
   ### Setup
 
+  ## If enabled, uses SumoLogic extension for auth/endpoint discovery instead of HTTP source URLs.
+  ## Terraform will auto-create an installation token and store it in the sumologic secret.
+  ## OTLP/HTTP sources are not created when this is enabled.
+  sourcelessMode: false
+
+  ## Endpoint used by collector pods to register and send data. Only used when sumologic.sourcelessMode is true.
+  collectorEndpoint: ""
```
dhruv-shah-sumo, line 24: Isn't it same as apiEndpoint that we used to provide during source creation and other Sumo APIs?  Is it different per customer or a generic per environment URL?

deploy/helm/sumologic/conf/setup/setup.sh
dhruv-shah-sumo: From what I have seen, argo only uses `helm template ...` to generate the config and does not use `helm upgrade/install` to deploy the resources. If only `helm template` is used with setupJob enabled, the setup job still does not get triggered for fetching the installation, right? I might be wrong but please double check it with the team maintaining argo or test it yourself.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
```diff
@@ -137,6 +137,17 @@ extensions:
       on_rebound: true
       directory: /tmp
 {{- end }}
+{{- if .Values.sumologic.sourcelessMode }}
+  sumologic:
```
ravishankar-sumo, line 141: We could move this to templates the logic would always remain identical except for the path

somewhere to `deploy/helm/sumologic/templates/_helpers/_common.tpl`

deploy/helm/sumologic/conf/setup/setup.sh
jagan2221: Usually Argo runs the setup job as a PreSync hook during sync. This is the same mechanism that creates the hosted collector and sources for all existing ArgoCD users today, if Argo fails to run setup job, no installation will succeed today since hosted collector and sources creation will be skipped. With this change, setup job just creates install token instead of Hosted collector and sources. so i believe it's safe here @dhruv-shah-sumo .

deploy/helm/sumologic/conf/metrics/otelcol/exporters.yaml
jagan2221: This will come in the help doc once we enable this feature. For config.merged exporters , users are free to use both source url mode and sourceless mode, we will capture this as recommended bast practices.
https://sumologic.atlassian.net/browse/SUMO-288600 - Tracking jira.

deploy/helm/sumologic/conf/metrics/otelcol/config.yaml
jagan2221: The token has same lifecycle of k8s helm chart. k8s helm chart once deleted(with sumologic.cleanupEnabled true), the token is deleted. It's not rotated , it stays the same after creation.

It can be managed manually also using sumologic.installationToken

deploy/helm/sumologic/conf/setup/setup.sh
jagan2221: Makes sense, added the support for providing install token manually via sumologic.installationToken .
Now token can either be manually managed via helm value (or) automatically-managed by our setup job.

deploy/helm/sumologic/conf/setup/setup.sh
jagan2221: Verified, setup pod only prints explicit echo command o/p and not the curl commands it runs.

deploy/helm/sumologic/conf/setup/setup.sh
jagan2221: Added 3 retries with 5 sec delay.

deploy/helm/sumologic/templates/checks.txt
jagan2221: Yes, implemented that flow now.

deploy/helm/sumologic/values.yaml
jagan2221: https://open-collectors.sumologic.com
 
 Sumologic extension always talks to open receiver which has above api base url.
 
Usually sumo extension always auto detects base url using token, so users need not to provide it. This was added just for our stag and long testing/epd setups where auto-detection won't work as an override aid.

Mentioned in readme the same.

deploy/helm/sumologic/conf/metrics/collector/otelcol/config.yaml
jagan2221: It's better to have the extension stay in the same config since they have different credential path.
Collector_fields value might get different in future when we add pipeline specific tags like logs-collector, metrics-collector etc to group pods with tags.

deploy/helm/sumologic/conf/cleanup/cleanup.sh
```diff
@@ -16,18 +16,58 @@ export TF_VAR_collector_name="${SUMOLOGIC_COLLECTOR_NAME}"
 export TF_VAR_secret_name="${SUMOLOGIC_SECRET_NAME}"
 export TF_VAR_chart_version="${CHART_VERSION:?}"
 export TF_VAR_namespace_name="${NAMESPACE:?}"
+export TF_VAR_use_extension="${SUMOLOGIC_USE_EXTENSION:-false}"
+export TF_VAR_extension_secret_name="${SUMOLOGIC_EXTENSION_SECRET_NAME:-sumologic-extension}"
+export TF_VAR_provided_installation_token="${SUMOLOGIC_INSTALLATION_TOKEN_PROVIDED:-false}"
```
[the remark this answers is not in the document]
jagan2221: ACK.

deploy/helm/sumologic/templates/cleanup/job.yaml
```diff
@@ -68,6 +68,12 @@ spec:
             value: "{{ .Chart.Version }}"
           - name: SUMOLOGIC_SECRET_NAME
             value: "{{ template "terraform.secret.name" }}"
+          - name: SUMOLOGIC_USE_EXTENSION
+            value: "{{ .Values.sumologic.sourcelessMode }}"
+          - name: SUMOLOGIC_EXTENSION_SECRET_NAME
+            value: "sumologic-extension"
+          - name: SUMOLOGIC_INSTALLATION_TOKEN_PROVIDED
+            value: {{ if .Values.sumologic.installationToken }}"true"{{ else }}"false"{{ end }}
```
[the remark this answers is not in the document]
jagan2221: ACK

deploy/helm/sumologic/templates/setup/job.yaml
```diff
@@ -97,6 +97,16 @@ spec:
...
         - name: CHART_VERSION
           value: "{{ .Chart.Version }}"
+        - name: SUMOLOGIC_USE_EXTENSION
+          value: "{{ .Values.sumologic.sourcelessMode }}"
+        - name: SUMOLOGIC_EXTENSION_SECRET_NAME
+          value: "sumologic-extension"
+{{- if .Values.sumologic.installationToken }}
+        - name: SUMOLOGIC_INSTALLATION_TOKEN
+          value: {{ .Values.sumologic.installationToken | quote }}
+{{- end }}
```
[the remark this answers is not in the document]
jagan2221: ACK

dhruv-shah-sumo (approved):

PR #4280, opened 2026-07-24
Fix: k8s sumologic extension - collector name format change to podName+clusterName

jagan2221:
Use collector_name field in sumologic extension to override name since default hostname has 2 cons
1. collector name more than 128 characters - This is already extended to 256 chars and fixed in backend. This is an enhancement in k8s side to make collector name more simpler. Instead of default hostname detected by sumo extension, override it with collector_name param by setting pod name from env variables.

2. collector name collision in multi cluster environment. Consider a user who has 30 different clusters where they use same release name and namespace for sumologic k8s. Now pod name generated by all such clusters will be of format 
`{pod-name}.{service-name}.{namespace}.svc.cluster.local`
This might collide, so better approach would be adding `podname+clusterName` which will always be unique.

Old collector name: `my-sumologic-sumologic-otelcol-events-0.my-sumologic-sumologic-otelcol-events-headless.sumologic3.svc.cluster.local`

new name: `my-sumologic-sumologic-otelcol-events-0-Kubernetes_cluster_sumo_extension17`

tested in local and ensured collector name is set properly:

Screenshot 2026-07-24 at 6 40 05 PM

### Checklist

- [ ] Documentation updated

### DIFF
--- a/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
@@ -160,7 +160,7 @@ extensions:
 {{- if .Values.sumologic.sourcelessMode }}
   sumologic:
     installation_token: ${SUMOLOGIC_INSTALLATION_TOKEN}
-    collector_name: ${MY_POD_NAME}
+    collector_name: ${MY_POD_NAME}-{{ .Values.sumologic.clusterName }}
     collector_fields:
       cluster: {{ .Values.sumologic.clusterName | quote }}
 {{- if .Values.metadata.persistence.enabled }}
--- a/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
+++ b/deploy/helm/sumologic/conf/logs/otelcol/config.yaml
@@ -160,6 +160,7 @@ extensions:
 {{- if .Values.sumologic.sourcelessMode }}
   sumologic:
     installation_token: ${SUMOLOGIC_INSTALLATION_TOKEN}
+    collector_name: ${MY_POD_NAME}
     collector_fields:
       cluster: {{ .Values.sumologic.clusterName | quote }}
 {{- if .Values.metadata.persistence.enabled }}
--- a/deploy/helm/sumologic/templates/_helpers/_common.tpl
+++ b/deploy/helm/sumologic/templates/_helpers/_common.tpl
@@ -525,6 +525,11 @@ Example Usage:
     fieldRef:
       apiVersion: v1
       fieldPath: status.podIP
+- name: MY_POD_NAME
+  valueFrom:
+    fieldRef:
+      apiVersion: v1
+      fieldPath: metadata.name
 {{- end -}}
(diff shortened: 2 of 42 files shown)


Gourav2906:
I think we should start with just pod name ?

jagan2221:
> I think we should start with just pod name ?

large customers use multiple clusters, if they install with same generic namespace and release name like sumologic to install chart in all clusters, it will result in pods entities getting mixed up between clusters. @Gourav2906

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
```diff
@@ -160,6 +160,7 @@ extensions:
 {{- if .Values.sumologic.sourcelessMode }}
   sumologic:
     installation_token: ${SUMOLOGIC_INSTALLATION_TOKEN}
+    collector_name: ${MY_POD_NAME}-{{ .Values.sumologic.clusterName }}
```
dhruv-shah-sumo, line 163: This change is limited to sourceless mode only, correct? I see the condition above but just wanted to confirm here.

In current architecture where sources exist, nothing changes, correct?

dhruv-shah-sumo (commented):
The changes look good to me. Just check if the new collector character limit can accommodate pod_name+cluster_name as the pods names can be too long.

deploy/helm/sumologic/conf/logs/otelcol/config.yaml
jagan2221: Correct, sumologic extension is applied only when sourceless mode is enabled.

jagan2221: Old collector name: my-sumologic-sumologic-otelcol-events-0.my-sumologic-sumologic-otelcol-events-headless.sumologic3.svc.cluster.local

new name: my-sumologic-sumologic-otelcol-events-0-Kubernetes_cluster_sumo_extension17

If you see new name, we already stripped a lot of unwanted things from pod name, so 256 chars now is very good limit to accomodate pod name+ clustername

deploy/helm/sumologic/templates/_helpers/_common.tpl
```diff
@@ -525,6 +525,11 @@ Example Usage:
     fieldRef:
       apiVersion: v1
```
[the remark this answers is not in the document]
jagan2221: Not worth introducing a new definition,

deploy/helm/sumologic/templates/metrics/collector/otelcol/opentelemetrycollector.yaml
```diff
@@ -148,6 +148,11 @@ spec:
       fieldRef:
         apiVersion: v1
         fieldPath: status.podIP
```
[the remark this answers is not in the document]
jagan2221: NA

dhruv-shah-sumo (approved):

PR #4334, opened 2026-09-11
fix(pvc-cleaner): do not fail when the HPA is temporarily missing

oribe-tala:
### Problem

`pvc-cleaner.sh` looks up the collector's HPA to decide how many replicas to keep, and treats a missing HPA as a hard failure:

```bash
if ! delete_all && ! check_hpa_exists; then
  echo "Provided HPA ${HPA_NAME} not found."
  exit 1
fi
```

The HPA is only *transiently* absent — `helm upgrade` deletes and recreates it. A CronJob tick landing inside that window fails, and then `backoffLimit: 6` retries it 6 more times against a condition that cannot succeed until the upgrade completes, leaving 7 `Error` pods behind per incident:

```
Error from server (NotFound): horizontalpodautoscalers.autoscaling "sumologic-collector-otelcol-metrics" not found
Provided HPA sumologic-collector-otelcol-metrics not found.
Caught error with code 1 on line 1.
The unmounted PVCs deletion has failed.
```

We hit this across 12 clusters in two rollout waves. Every single failure landed 9–24 minutes before a completed `helm upgrade` of the release — no other failures, and every run outside an upgrade window succeeds in 3–4s.

### Fix

A missing HPA isn't an error. It means there's no safe replica count to compare against, so there is nothing to clean this run — exit 0 and let the next tick (15 min later by default) pick it up. This matches how the script already handles an empty PVC selector a few lines below ("Nothing to clean.", exit 0).

No PVC is left behind: the cleanup is idempotent and runs every 15 minutes.

### Also included

A separate, independent commit adds `pvcCleaner.job.startingDeadlineSeconds`, plumbed into both the logs and metrics CronJobs, defaulting to `null` so rendered output is unchanged unless set.

To be explicit, since it would be a reasonable thing to assume: **this does not fix the bug above.** `startingDeadlineSeconds` only governs runs the CronJob controller failed to *create* at their scheduled time; the failing runs here were created on schedule and then exited 1. It's offered purely as configurability that the chart doesn't currently expose. Happy to drop it into its own PR if you'd rather keep this one focused.

### Testing

- `make regenerate-goldenfiles` — the two `full_config` pvc-cleaner goldens gain one line each; `single_layer_pipeline_metrics` (which does not set the new value) regenerates byte-identical, covering the unset path
- `go test ./...` in `tests/helm` — pass
- `make check-configuration-keys` — pass
- `prettier --check` on yaml and md — clean
- `helm lint` — pass

### Checklist

- [x] Documentation updated


dhruv-shah-sumo (approved):

dhruv-shah-sumo:
Thanks for your contribution @oribe-tala. Merging the PR.

